PR #16, opened 2026-07-23
Remove stale docs tree

alessiodevoto:
Cherry-picks the stale docs cleanup from PR #14 onto main, since #14 landed on nooa-skill-renames after PR #13 was merged.


alessiodevoto:
Merging since it was merged already but in the wrong branch.

PR #26, opened 2026-07-27
security: require auth token and restrict CORS on viewer API

rdasilveiracabral:
The viewer exposed trace, evaluation, memory, annotation, explorer, OTLP and journal APIs with no authorization while listening on all interfaces, and its CORS policy allowed every origin, method and header.

Add a `_require_viewer_authorization` dependency on every API router and every `/v1` ingest/sync endpoint. When `NOOA_VIEWER_AUTH_TOKEN` is set, requests must send `Authorization: Bearer <token>`, compared with `hmac.compare_digest()`. When it is unset, loopback clients keep working so existing local development and ingest clients are unaffected, while non-loopback clients get a 403.

Restrict CORS to `http://localhost:5001` and `http://127.0.0.1:5001` by default, overridable via a comma-separated `NOOA_VIEWER_CORS_ORIGINS`, and narrow allowed methods and headers to the ones the viewer APIs use.

## Checklist

- [ ] Code follows the project style (`uv run ruff check .` and `uv run ruff format --check .` pass)
- [ ] Tests added/updated and passing (`uv run pytest`)

### DIFF
--- a/src/nooa/viewer/main.py
+++ b/src/nooa/viewer/main.py
@@ -8,6 +8,8 @@ for client-side routing.
 """
 
 import asyncio
+import hmac
+import ipaddress
 import logging
 import os
 import time
@@ -15,7 +17,7 @@ from concurrent.futures import ThreadPoolExecutor
 from contextlib import asynccontextmanager
 
 from dotenv import load_dotenv
-from fastapi import FastAPI, Request
+from fastapi import Depends, FastAPI, HTTPException, Request
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import FileResponse, JSONResponse
 from starlette.requests import ClientDisconnect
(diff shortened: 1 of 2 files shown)


rdasilveiracabral:
## Security issue

The viewer exposed trace, evaluation, memory, annotation, playground, OTLP, and journal APIs without an authorization check while listening on all interfaces. Its CORS policy allowed every origin, method, and header. A network client or malicious browser origin could read stored agent information or invoke state-changing routes.

rdasilveiracabral:
## Patch

Set `NOOA_VIEWER_AUTH_TOKEN` to enable authorization on every API router and every `/v1` ingest/synchronization endpoint. Protected requests must send:

```text
Authorization: Bearer <NOOA_VIEWER_AUTH_TOKEN>
```

The token is read at request time and compared with `hmac.compare_digest()`. Frontend assets and the SPA shell remain accessible without the API token. Leaving the variable unset preserves existing loopback development and ingest clients, but protected endpoints reject non-loopback clients. Network-exposed deployments must configure the token.

CORS now allows only `http://localhost:5001` and `http://127.0.0.1:5001` by default, with explicit override through a comma-separated `NOOA_VIEWER_CORS_ORIGINS` value. Allowed methods and headers are limited to the ones the viewer APIs use.

PR #27, opened 2026-07-27
Constrain playground custom-model endpoint/api_key_env to server-declared pairs

sklinglernv:
## Summary

Fixes a secret-exfiltration + SSRF vulnerability in the viewer playground.

Custom models registered via the playground API (`POST /api/playground/models`) are **browser-controlled state**, but their `endpoint` and `api_key_env` were forwarded unvalidated into `litellm` in `run_inference`:

```python
kwargs["api_base"] = model_config["endpoint"]                       # attacker URL
api_key = resolve_api_key_from_config(request.model, model_config)  # os.getenv(any name)
kwargs["api_key"] = api_key
await litellm.acompletion(**kwargs)                                 # key POSTed to that URL
```

An attacker who can reach the viewer could register a model where:
- `endpoint` → their own host and `api_key_env` → any process env var (e.g. a cloud/DB secret) → the resolved secret is sent to the attacker (**secret exfiltration**), or
- `endpoint` → `http://169.254.169.254/...` or an internal service → **SSRF** into the metadata service / internal network.

## Fix

Custom models may only *reuse* an `(endpoint, api_key_env)` pair the operator already declared server-side in `models.yaml` (or provider defaults) — never invent one.

- **`unifiedllm/registry.py`** — `resolve_api_key_from_config` validates `api_key_env` against a shell-style env-var regex and accepts an optional `allowed_env_vars` allowlist (defense-in-depth for callers handling untrusted config).
- **`viewer/trace_routes.py`** — trusted pairs are built from the server-side models config; enforced on write (HTTP 400) and stripped on read (`get_playground_models`, `get_model_config`). `run_inference` resolves keys only from the known api-key allowlist. `load_custom_models` also gains a type-confusion guard.

Extracted from `0001-playground-model-boundary` of the security-hardening patch set (the other patches in that set cover unrelated issues and are not included here).

## Testing

- Full suite: **6437 passed, 4 skipped** (`uv run pytest`).
- All pre-commit hooks pass (ruff, ruff-format, pyright, hygiene checks). One pyright type-inference fix was applied on top of the patch (`pairs` set annotation).

### DIFF
--- a/src/nooa/unifiedllm/registry.py
+++ b/src/nooa/unifiedllm/registry.py
@@ -46,8 +46,9 @@ from __future__ import annotations
 
 import logging
 import os
+import re
 import threading
-from collections.abc import Mapping
+from collections.abc import Collection, Mapping
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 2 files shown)


alessiodevoto:
LGTM!

PR #32, opened 2026-07-27
test: fill signal, sandbox, and TaskWrapper edge-case stubs

skundu42:
## Summary
- Implemented edge-case coverage for concurrent signals during generation (`test_signal_edge_cases.py`)
- Implemented sandboxed CodeAct REPL edge cases: state mutation, tool brokering, and `execute_python` NameError recovery (`test_sandbox_edge_cases.py`)
- Added TaskWrapper contract tests for `task.abort` / `task.replan` / `task.current`, skipped until that builtin is implemented (`test_task_wrapper_edge_cases.py`)

## Test plan
- [x] `uv run pytest tests/edge_cases/test_signal_edge_cases.py tests/edge_cases/test_sandbox_edge_cases.py tests/edge_cases/test_task_wrapper_edge_cases.py -v`


tests/edge_cases/test_task_wrapper_edge_cases.py
alessiodevoto: I think we can safely remove this file

tests/edge_cases/test_signal_edge_cases.py
```diff
@@ -6,6 +6,198 @@
...
+        gate=gate,
+    )
+    agent = SignalAgent(llm=fake)
+
+    gen_task = asyncio.create_task(agent.long_gen())
+    await fake.entered.wait()
+
+    sig_task = asyncio.create_task(agent.signal_calls_gen())
+    # Give the signal time to enter and block on the generation lock.
+    await asyncio.sleep(0.05)
```
alessiodevoto: Can we make `GatedFakeLLM` track entries into `acall()` before it waits on `gate`, then assert that only the outer generation has entered the LLM before `gate.set()`? Otherwise this could still pass if `nested_gen()` incorrectly gets past `_generation_lock` and blocks inside the fake LLM on the same gate.

alessiodevoto:
Hi @skundu42! Thanks for opening this PR 🙂 

I ran locally and got `7 passed, 3 skipped`, I noticed the TaskWrapper tests but we actually don't have that implemented yet, so I’m not comfortable merging the TaskWrapper part as `task.abort`, `task.replan`, and `task.current` are not implemented today. Please remove `tests/edge_cases/test_task_wrapper_edge_cases.py`.

For the signal tests, I think those make sense but could you please make the lock assertion deterministic, for example replace the `asyncio.sleep(0.05)` checks with an explicit event/counter in the new `GatedFakeLLM`, and assert that the nested generation has not entered the LLM call before the outer generation releases the lock ? I left a comment in the code too!

skundu42:
@alessiodevoto Thanks for the review. I have made the requested changes. Let me know if this works. 

Also, please feel free to assign me other issues directly. I am planning to be an active contributor in this.

alessiodevoto:
Hi @skundu42 thanks for updating ! LGTM, merging 🙂 !

PR #42, opened 2026-07-28
fix: preserve sandbox state after recoverable exits 🤖🤖🤖

jmolz:
## What does this PR do?

Sandbox cells capture locals in the generated wrapper finally block, but the worker previously merged them into its persistent namespace only after the wrapper returned successfully. An ordinary exception or an ExecutionSignal from return_result() skipped that merge, so assignments made earlier in the cell disappeared and the next cell failed with NameError.

This change:

- merges captured locals from an outer finally block so state survives errors and control-flow signals;
- keeps sandbox behavior aligned with the in-process ActorRuntime REPL semantics;
- adds normal-CI cell-engine regressions plus real-worker integration coverage;
- resolves the two existing Pyright errors in the touched block so the required pre-commit gate passes.

## Related issues

No open issue currently tracks this defect.

## Validation

- focused regression tests before the fix: 2 failed with NameError
- focused regression tests after the fix: 2 passed
- cell-engine regression suite: 8 passed
- non-live suite: 6,475 passed, 6 skipped
- all pre-commit hooks passed, including Ruff, formatting, Pyright, and SPDX
- DCO sign-off and cryptographic commit signature verified locally

## Checklist

- [x] Change is focused on sandbox REPL state persistence
- [x] Unit and integration tests added
- [x] Existing behavior on successful cells preserved
- [x] Source-file SPDX headers retained

🤖🤖🤖

### DIFF
--- a/src/nooa/runtime/sandbox/cell_core.py
+++ b/src/nooa/runtime/sandbox/cell_core.py
@@ -22,7 +22,7 @@ import io
 import linecache
 import types
 from collections.abc import Callable
-from typing import Any
+from typing import Any, cast
 
 from nooa.events import _NO_RETURN, ExecutionResult, ExecutionSignal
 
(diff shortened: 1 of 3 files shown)


alessiodevoto (commented):
@dcoapp recheck

alessiodevoto (approved):
Thanks LGTM!

PR #45, opened 2026-07-28
fix(unifiedllm): message-level cache_control on non-Anthropic providers

maemayv:
## What does this PR do?
Cache-control injection converted the last tool-role message's content to the Anthropic parts-array form for every provider. OpenAI-compatible gateways (e.g. NVIDIA vLLM/NIM) reject a content list on non-user roles, so every request ending in a tool result failed there, and structured-output correction retries could never recover. The litellm patch strips the marker for non-Anthropic models but leaves the list shape behind, so stripping alone didn't fix it.

Gate the parts-array conversion on _is_anthropic_model(): non-Anthropic providers get cache_control at the message level instead. Adds a regression test and points the position-injection fixture at an Anthropic model (that suite asserts the content-block form, now Anthropic-specific).

## Related issues

No related issue

## Checklist

- [x ] Code follows the project style (`uv run ruff check .` and `uv run ruff format --check .` pass)
- [ x] Tests added/updated and passing (`uv run pytest`)

### DIFF
--- a/src/nooa/unifiedllm/unifiedllm.py
+++ b/src/nooa/unifiedllm/unifiedllm.py
@@ -1280,12 +1280,18 @@ class UnifiedLLM(ABC):
                         msg["cache_control"] = {"type": "ephemeral"}
                         break
 
+        # Anthropic needs cache_control on a content block (parts form); other providers
+        # reject a content list on non-user roles, so mark at the message level instead.
+        anthropic = _is_anthropic_model(self.model)
         for role in roles_to_cache_last:
             # Search for matching messages by role OR by equivalent native type
             native_type = _ROLE_TO_TYPE.get(role)
             for msg in reversed(messages):
                 if msg.get("role") == role or (native_type and msg.get("type") == native_type):
-                    self._inject_cache_control_on_content(msg)
+                    if anthropic:
+                        self._inject_cache_control_on_content(msg)
+                    else:
+                        msg["cache_control"] = {"type": "ephemeral"}
                     break
 
         return messages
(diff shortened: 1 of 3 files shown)


sklinglernv:
### Capability suite — no regressions

Ran `tests/capability/config.yaml` (40 tests, 116 samples/model) via `eval_pipeline` against the NVIDIA gateway, comparing `main` to this change cherry-picked onto `main` tip:

| | claude-opus-4-8 | gpt-5.5 |
|---|---|---|
| `main` | 115/116 | 116/116 |
| this PR | **116/116** | **116/116** |

The one `main` failure (`realfmt_slice_keys_list_predict_007`) is a model reasoning miss unrelated to this diff; it passed on the PR run. Unit suite: 6483 passed, 4 skipped.

Consistent with the diff: Claude routes as `openai/azure/anthropic/...`, so `_is_anthropic_model()` matches and the Anthropic path is byte-identical. For non-Anthropic, litellm's `filter_value_from_dict` strips the message-level marker recursively, so it never reaches the wire — the fix is the preserved string `content` shape. `ResponsesClient` already gates injection on `_is_anthropic_model`, so the new `else` branch is unreachable there.

Two notes:
- I couldn't reproduce the original failure. A minimal request ending in a tool result with `{"role": "tool", "position": "last"}` succeeded on `main` for `claude-opus-4-8`, `gpt-5.5`, `nemotron-3-ultra`, and `nemotron3-nano-30b`. The trigger is likely a specific structured-output path — worth recording the failing request shape.
- Branch is behind `main` (missing `f8cd906`). Cherry-picks cleanly; all numbers above are from the rebased version.

sklinglernv:
Thank you for the fix. LGTM!

PR #46, opened 2026-07-28
fix: honor tool_call_timeout so slow MCP tool calls stop hanging

callingmedic911:
`MCPStreamableHTTPClient` builds its httpx client without a timeout, so httpx's
5 second default governs the MCP read stream. The transport takes a pre-built
client and has no timeout arguments of its own, so nothing else applies. When a
tool takes longer, the server finishes the work and writes the reply to a stream
httpx already dropped. Nothing raises, so the caller hangs.


sklinglernv:
LGTM, thanks a lot for the fix!

PR #52, opened 2026-07-29
feat: support deterministic generator agent methods (fix #38)

### LINKED ISSUE #38: Generator-bodied agent methods: the span ends at generator creation, so body spans attach to the consumer
**Describe the bug**

When an agent method is a generator (`async def` with `yield`, or a plain `def` with `yield`), its AGENT span opens and closes around the *creation* of the generator object, before any line of the body has run. The body executes later, at consumption time, outside that span. Calls the body makes — including LLM calls — are then recorded as children of whichever method drained the generator.

The result is not a missing span. It is a span tree that asserts a call that never happened: below, `run_batch()` is recorded as the parent of two `classify()` LLM calls it never issued, while the helper that did issue them shows a span covering no work.

Cause: `AgentMeta.__new__` dispatches on `inspect.iscoroutinefunction` (`src/nooa/metaclass.py:78`) and then `elif inspect.isfunction` (`:96`). An async generator function is `False` for the first and `True` for the second, so both async and sync generator methods are routed to `create_sync_agent_method_wrapper` — whose own docstring says it "Skips the generation path entirely (sync methods can't await an LLM)" (`src/nooa/runtime/method_wrapper.py:397`). An async generator method can await an LLM.

Nothing misbehaves functionally: both agents below return `['approved', 'approved']` and make exactly 2 LLM calls. Only the trace differs.

**To reproduce**

No API key or provider needed — `FakeLLMClient` + `InMemorySpanExporter`.

```bash
git clone https://github.com/NVIDIA-NeMo/labs-OO-Agents.git
cd labs-OO-Agents && git checkout ec4bd58c026aeb40e0b7e6468c57dc5a02b5d31e
uv sync --all-extras
uv run python repro.py    # the script below
```

The two agents are identical except that `GeneratorHelper.review` uses `yield` and `PlainHelper.review` uses `return`.

```python
import asyncio
import inspect
import json
from typing import Any

from opentelemetry import trace
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import SimpleSpanProcessor
from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
from pydantic import BaseModel

from nooa import Agent
from nooa.config import PredictConfig
from nooa.decorators import strategy
from nooa.strategies.predict import PredictStrategy
from nooa.tracing import NemoOOAgentsInstrumentor
from nooa.tracing._session import set_session
from nooa.tracing._session_processor import SessionSpanProcessor
from nooa.unifiedllm.fake import FakeLLMClient
from nooa.unifiedllm.unifiedllm import LLMResponse, Tool

exporter = InMemorySpanExporter()
provider = trace.get_tracer_provider()
if not hasattr(provider, "add_span_processor"):
    provider = TracerProvider()
    trace.set_tracer_provider(provider)
provider.add_span_processor(SessionSpanProcessor())
provider.add_span_processor(SimpleSpanProcessor(exporter))
NemoOOAgentsInstrumentor().instrument(tracer_provider=provider)
set_session("repro-session")


class Verdict(BaseModel):
    label: str


class ScriptedLLM(FakeLLMClient):
    """Always returns a valid Verdict; counts calls."""

    async def acall(
        self,
        messages: list[dict[str, Any]],
        tools: list[Tool] | None = None,
        output_model: type[BaseModel] | None = None,
        **kwargs: Any,
    ) -> LLMResponse:
        self.call_count += 1
        content = json.dumps({"label": "approved"})
        return LLMResponse(
            raw_response=None,
            content=content,
            tool_calls=[],
            finish_reason="stop",
            assistant_message={"role": "assistant", "content": content},
        )


llm = ScriptedLLM()
PREDICT = PredictStrategy(config=PredictConfig(max_retries=3))


class GeneratorHelper(Agent, llm=llm):
    """The helper that issues the LLM calls is an async generator."""

    @strategy(PREDICT)
    async def classify(self, item: str) -> Verdict:
        """Classify {item}."""
        ...

    async def review(self, items: list[str]):
        """Async generator helper -- issues one LLM call per item."""
        for it in items:
            yield (await self.classify(it)).label

    async def run_batch(self, items: list[str]) -> list[str]:
        """Orchestrator. Only iterates; issues no LLM call of its own."""
        return [label async for label in self.review(items)]


class PlainHelper(Agent, llm=llm):
    """Control: identical work, helper returns a list instead of yielding."""

    @strategy(PREDICT)
    async def classify(self, item: str) -> Verdict:
        """Classify {item}."""
        ...

    async def review(self, items: list[str]) -> list[str]:
        """Ordinary async helper -- issues one LLM call per item."""
        return [(await self.classify(it)).label for it in items]

    async def run_batch(self, items: list[str]) -> list[str]:
        """Orchestrator. Only delegates; issues no LLM call of its own."""
        return await self.review(items)


def parent_name(span, spans):
    if span.parent is None:
        return None
    for s in spans:
        if s.context.span_id == span.parent.span_id:
            return s.name
    return "<unknown>"


async def show(agent_cls, tag):
    exporter.clear()
    out = await agent_cls().run_batch(["a", "b"])
    await asyncio.sleep(0)
    spans = sorted(exporter.get_finished_spans(), key=lambda s: s.start_time)

    orig = getattr(agent_cls.review, "_original", agent_cls.review)
    shape = "async generator" if inspect.isasyncgenfunction(orig) else "coroutine"
    print(f"\n{'=' * 76}\n{tag}   (review() is a {shape})   result={out}\n{'=' * 76}")
    for s in spans:
        dur = (s.end_time - s.start_time) / 1000
        print(f"  {s.name:26s} parent={str(parent_name(s, spans)):26s} {dur:9.1f}us")

    cls_spans = [s for s in spans if s.name == "method.classify"]
    parents = {parent_name(s, spans) for s in cls_spans}
    review = [s for s in spans if s.name == "method.review"]
    print(f"\n  ground truth : review() issued all {len(cls_spans)} classify calls")
    print(f"  recorded     : classify parented to {parents}")
    if review:
        r = review[0]
        print(f"  method.review span: {(r.end_time - r.start_time) / 1000:.1f}us")
    return parents == {"method.review"}


async def main():
    bad_ok = await show(GeneratorHelper, "CASE -- generator helper")
    good_ok = await show(PlainHelper, "NEGATIVE CONTROL -- non-generator helper")
    print(f"\n{'=' * 76}\nSUMMARY\n{'=' * 76}")
    print(f"  generator helper     : parent linkage correct? {bad_ok}")
    print(f"  non-generator helper : parent linkage correct? {good_ok}")


asyncio.run(main())
```

**Expected behavior**

README.md:157 states: "Every LLM call, code execution, and method invocation is traced by default — orchestrators, generation methods, and helpers, with parent-child spans preserved."

`review()` issues both `classify()` calls in both agents; `run_batch()` issues none. I expected `method.classify` to be recorded as a child of `method.review` in both cases, and the `method.review` span to cover the work its body does.

Actual, at commit `ec4bd58c`:

| helper shape | `method.review` span | recorded parent of `method.classify` | matches ground truth |
| --- | --- | --- | --- |
| async generator (`yield`) | 172µs | `method.run_batch` | no |
| coroutine (`return`) | 12288µs | `method.review` | yes |

```
CASE -- generator helper   (review() is a async generator)   result=['approved', 'approved']
  method.run_batch           parent=None                         16607.0us
  method.review              parent=method.run_batch               172.0us
  method.classify            parent=method.run_batch              9194.0us
  generation                 parent=method.classify               8246.0us
  method.classify            parent=method.run_batch              6564.0us
  generation                 parent=method.classify               5993.0us

NEGATIVE CONTROL -- non-generator helper   (review() is a coroutine)   result=['approved', 'approved']
  method.run_batch           parent=None                         12603.0us
  method.review              parent=method.run_batch             12288.0us
  method.classify            parent=method.review                 5725.0us
  generation                 parent=method.classify               5293.0us
  method.classify            parent=method.review                 6070.0us
  generation                 parent=method.classify               5590.0us
```

**Environment**

- OS: macOS 15.7.3 (Darwin 24.6.0), Apple silicon
- Python version: 3.12.13
- NOOA version (`nooa --version` or `python -c "import nooa; print(nooa.__version__)"`): `0.0.7.dev35`, clone commit `ec4bd58c026aeb40e0b7e6468c57dc5a02b5d31e` on `main`
- Model / provider: none — `nooa.unifiedllm.fake.FakeLLMClient`, no credentials required

Baseline before investigating: `uv run pytest -m "not integration and not stress"` → 6442 passed, 6 skipped, 282 deselected; `ruff check src/` and `ruff format --check src/` clean.

**Additional context**

*Mechanism.* `create_sync_agent_method_wrapper` (`method_wrapper.py:383`) calls the target once and treats the return value as the result (`:468`). For a generator function that returns the generator object immediately; the surrounding `try/finally` pops the agent call-id stack and closes the span (`:492-501`) while the body is still suspended at its first `yield`. When the caller drains it, `_push_agent_call_id`/`_pop_agent_call_id` have already unwound, so nested calls resolve their parent from whatever is on the stack at consumption time.

The dispatch docstring at `metaclass.py:38-39` enumerates what the `isfunction` branch is meant to exclude — properties, classmethods, staticmethods — and generator functions are not in that list, which suggests unintended routing rather than deliberate exclusion. `inspect.isgeneratorfunction` and `inspect.isasyncgenfunction` appear nowhere in `src/`, and no test in `tests/` declares a generator method on an `Agent` subclass. `git log` shows both files carry only `f6f28f5`, so this is not a regression.

*Other observed shapes.* Sync generators (`def` + `yield`) behave the same — a generator method setting `self.touched = True` leaves `touched` still `False` after its span has ended, flipping only on drain. If a generator is created inside a traced method and drained after that method returns, the body's spans have no ancestor on the stack and are emitted as new roots (`parent=None`), splitting one logical run across two traces.

*Stability.* The parent edge is stable across 10 runs (two independent sets of 5): `method.classify` under `method.run_batch` every time in the generator case, under `method.review` every time in the control. I expect that to be structural — no concurrency, the body resumes inside the consumer's frame, and the call-id stack is an ordinary context-local stack — but that is corroboration, not proof for every scheduling. The microsecond durations above are from a single run and vary; nothing rests on them, the parent edge is the claim.

*Scope.* I have not observed this in a running system. No test declares a generator agent method, and the one generator I found in `examples/` (`examples/arc_agi_3/solver_agent.py:444`) is a module-level function, not an agent method, so it never reaches this dispatch. This is a latent defect on a documented interface, not a demonstrated production failure. I also did not check whether generated CodeAct code calling a generator helper hits the same path — the defect is in the metaclass dispatch, which runs at class creation regardless of caller, so I would expect it to, but I have not demonstrated it and am not asserting it.

*Existing issues.* I checked the tracker first. #23 asks for a runtime-owned effect-record primitive and does not report missing or misattributed spans. #20 is per-method LLM selection. #35/#31 concern `eval()` in CodeAct's constructor-string coercion; #30 reframes the code validators as guardrails. None overlap.

*Security routing.* I filed publicly on my own judgement that this is a correctness defect in observability output rather than a vulnerability: it grants no capability and crosses no trust boundary, and NOOA does not claim traces are a security control. That is a judgement call and yours to overrule — if you treat trace integrity as security-relevant, tell me and I will withdraw this and re-route through PSIRT per SECURITY.md.

*Two resolutions look legitimate.* **(a) Support generator methods** — add `isasyncgenfunction` / `isgeneratorfunction` arms at `metaclass.py:78`/`:96`, routing to a wrapper that opens the span on first `__anext__`/`__next__` and closes on exhaustion or `aclose()`. Keeps README.md:157 true as written; the cost is that the span's lifetime no longer coincides with the call that created the generator, and an abandoned generator needs a defined end condition. **(b) Reject them at class creation** — detect them in the same dispatch, raise at class-creation time, and narrow README.md:157 to match.

The current behaviour is neither: a generator method is silently mis-dispatched to a wrapper documented as not handling it, and the run emits a trace asserting a call that never happened, with nothing signalling that anything was skipped. (b) is a fine answer to "we don't support generators" — but said out loud, rather than expressed as a confidently wrong span tree. A third option, one span per `yield` step, is available if you want per-resumption granularity.

**I am happy to submit the patch**: the dispatch arms, the wrapper, and a regression test asserting that a generator-bodied agent method emits a span covering its body — that a call made from inside the body is recorded as a child of the generator's own span, with the non-generator equivalent as the control. The two agents above are already that pair. Say which of (a) or (b) you want and I will open the PR.

inaor:
## Problem

Fixes #38.

Generator methods were previously sent through regular method wrappers. Those wrappers opened and closed a span around creation of the generator object, before its body ran, so nested calls made while consuming the generator received the wrong trace parent.

## Design

This PR now uses the framework's semantic boundary rather than rejecting all generators:

- Deterministic sync and async generator methods are supported as ordinary Python.
- Generator methods containing the `...` generation marker are rejected at class creation with a clear error.
- Generation strategies retain their existing one-final-result contract. Generated streaming would require a separate contract for yielded item types, completion, retries, partial failure, and backpressure.
- The unary `agent_call` middleware contract does not wrap generator lifecycles. A future generated-stream feature should define stream middleware explicitly rather than overloading unary middleware.

## Implementation

- Adds generator-aware sync and async wrappers that preserve `next`/`anext`, `send`/`asend`, `throw`/`athrow`, close, cancellation, errors, and sync-generator return values.
- Starts the method span on first iteration and ends it on exhaustion, explicit close, cancellation, or failure.
- Activates agent and OpenTelemetry context only while generator code is running, so nested calls are parented to the generator while consumer work between yields remains separate.
- Carries tracing context safely across sequential resumes from different asyncio tasks and cleans up the originating task's span registry.
- Distinguishes actual generation markers from ordinary Python Ellipsis expressions such as `tuple[int, ...]` and `items[...]`, including source-less generated functions.
- Documents the single-result generation contract.

## Testing

The tests cover native generator protocols, generated-generator rejection and diagnostics, explicit close and cancellation, cross-task tracing, consumer isolation, span-registry cleanup, traced cleanup during `aclose()`, ordinary Ellipsis expressions, and source-less bytecode detection. OpenTelemetry assertions use an in-memory exporter with real spans rather than mocks.

```bash
uv run pytest tests/tracing tests/runtime tests/test_metaclass.py tests/test_pre_ellipsis_code.py tests/test_ellipsis_detection_exec.py tests/external/test_decorators.py -q -p no:cacheprovider
# 1369 passed, 6 skipped

PYTHON_DOTENV_DISABLED=1 NOOA_VIEWER_AUTH_TOKEN= uv run pytest -m 'not integration and not stress' --ignore=tests/test_mcp -q -p no:cacheprovider
# 6384 passed, 5 skipped

uv run ruff check ...
uv run ruff format --check ...
uv run pyright ...
# all clean
```

## Summary by CodeRabbit

* **New Features**
  * Added support for deterministic synchronous and asynchronous generator methods.
  * Preserved native generator controls, including send, throw, close, and cancellation behavior.
  * Improved tracing and span continuity while generators pause and resume across tasks.

* **Bug Fixes**
  * Added validation to reject unsupported generated streaming methods with clear errors.
  * Improved tracing context restoration and cleanup during generator execution and cancellation.

* **Documentation**
  * Documented generator support requirements and streaming generation limitations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ class SupportAgent(Agent):
 
 - **Agents are Python objects.** Fields are state, methods are capabilities, docstrings are prompts, type annotations are contracts.
 - **`...` bodies are LLM-driven.** A method with `...` becomes an agentic loop; a real body stays deterministic Python.
+- **Generation commits one result.** Deterministic `def` and `async def` generators are supported, but combining `yield` with the `...` generation marker is rejected; generated streaming needs a separate stream contract.
 - **Code as action.** The model acts by writing Python in a Jupyter-style REPL with access to `self`, imports, and helpers — Python methods and type annotations supply the callable interfaces, reducing the need to write separate tool-schema definitions.
 - **Pythonic and agent-ready.** Typed I/O with auto-retry, live-object arguments passed by reference, and model-callable context and event APIs — designed around agent-oriented Python workflows.
 
--- a/src/nooa/decorators.py
+++ b/src/nooa/decorators.py
@@ -11,7 +11,7 @@ from collections.abc import Callable
 from typing import TYPE_CHECKING, Any, ParamSpec, TypeVar
 
 from nooa.context_blocks import ScopedContext
-from nooa.ellipsis_detection import has_ellipsis_body
+from nooa.ellipsis_detection import has_ellipsis_body, has_ellipsis_marker
 
 if TYPE_CHECKING:
     from nooa.config.truncation_config import TruncationConfig
@@ -92,6 +92,22 @@ def strategy(
         setattr(func, "_strategy_context", final_context)  # noqa: B010
         setattr(func, "_strategy_events", final_events)  # noqa: B010
         setattr(func, "_strategy_truncation", truncation)  # noqa: B010
+
+        # Strategies currently have a single-result contract. Deterministic
+        # generators do not need @strategy and are handled by AgentMeta; generated
+        # streams need a separate strategy protocol rather than this decorator.
+        if inspect.isasyncgenfunction(func) or inspect.isgeneratorfunction(func):
+            if has_ellipsis_marker(func):
+                raise TypeError(
+                    f"@strategy method '{func.__name__}' is an LLM-generated generator. "
+                    f"Generation methods must produce one final result; streaming "
+                    f"generation is not supported."
+                )
+            raise TypeError(
+                f"@strategy cannot be applied to deterministic generator method "
+                f"'{func.__name__}'. Remove @strategy; deterministic generators are "
+                f"supported directly as agent methods."
+            )
 
         # Also create runtime wrapper for dynamically-defined methods
         if not inspect.iscoroutinefunction(func):
--- a/src/nooa/ellipsis_detection.py
+++ b/src/nooa/ellipsis_detection.py
@@ -7,10 +7,12 @@ that use `...` as a marker for LLM code generation.
 
 Functions:
     has_ellipsis_body: Check if function body ENDS with `...`
+    has_ellipsis_marker: Check if a function body contains a generation ellipsis
     get_pre_ellipsis_code: Extract setup code before `...`
 """
 
 import ast
+import dis
 import inspect
 import textwrap
 import tokenize
@@ -114,6 +116,75 @@ def has_ellipsis_body(func: Callable[..., Any]) -> bool:
 
     # If we can't determine, assume it's not ellipsis (safer default)
     return False
+
+
+def has_ellipsis_marker(func: Callable[..., Any]) -> bool:
+    """Return whether *func* contains an ellipsis generation marker.
+
+    Ordinary generation methods are identified by :func:`has_ellipsis_body`,
+    because their body ends in a bare ``...`` statement. Generator functions
+    need a slightly broader check: ``yield ...`` is also a natural spelling of
+    an attempted generated stream. Generated streams are not currently a
+    supported execution model, so both spellings must be diagnosed rather than
+    accidentally executed as deterministic Python.
+
+    Ellipses inside nested functions or classes do not mark the outer function.
+    """
+    func_def = _get_function_ast(func)
+    if func_def is not None:
+
+        class _MarkerVisitor(ast.NodeVisitor):
+            found = False
+
+            def visit_Expr(self, node: ast.Expr) -> None:
+                if _is_ellipsis_stmt(node):
+                    self.found = True
+                    return
+                self.generic_visit(node)
+
+            def visit_Yield(self, node: ast.Yield) -> None:
+                if isinstance(node.value, ast.Constant) and node.value.value is ...:
+                    self.found = True
+
+            def visit_YieldFrom(self, node: ast.YieldFrom) -> None:
+                if isinstance(node.value, ast.Constant) and node.value.value is ...:
+                    self.found = True
+
+            # Nested definitions have their own generation semantics.
+            def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
+                return
+
+            def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:
+                return
+
+            def visit_ClassDef(self, node: ast.ClassDef) -> None:
+                return
+
+            def visit_Lambda(self, node: ast.Lambda) -> None:
+                return
+
+        visitor = _MarkerVisitor()
+        for stmt in _get_body_without_docstring(func_def.body):
+            visitor.visit(stmt)
+        return visitor.found
+
+    # Dynamically created functions may not have source. Detect the direct
+    # ``LOAD_CONST Ellipsis; YIELD_VALUE`` bytecode shape without treating
+    # ordinary uses such as ``items[...]`` as generation markers. The existing
+    # body detector covers functions carrying ``_generated_source``.
+    try:
+        instructions = list(dis.get_instructions(func))
+        for index, current in enumerate(instructions):
+            if current.opname != "LOAD_CONST" or current.argval is not ...:
+                continue
+            for following in instructions[index + 1 : index + 3]:
+                if following.opname == "YIELD_VALUE":
+                    return True
+                if following.opname not in {"ASYNC_GEN_WRAP", "CALL_INTRINSIC_1"}:
+                    break
+    except Exception:
+        pass
+    return has_ellipsis_body(func)
 
 
 def get_pre_ellipsis_code(func: Callable[..., Any]) -> str | None:
--- a/src/nooa/metaclass.py
+++ b/src/nooa/metaclass.py
@@ -19,7 +19,7 @@ from abc import ABCMeta
 from collections.abc import Callable
 from typing import Any
 
-from nooa.ellipsis_detection import has_ellipsis_body
+from nooa.ellipsis_detection import has_ellipsis_body, has_ellipsis_marker
 
 
 class AgentMeta(ABCMeta):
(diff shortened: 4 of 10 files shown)


alessiodevoto (commented):
@dcoapp recheck

sklinglernv:
@CodeRabbit review full

sklinglernv:
Superseded by #120, which now contains the canonical generator implementation plus the OpenTelemetry cross-task tracing and generator-diagnostic hardening from this PR (ported in 5c5e244). GitHub build, lint, test, frontend-build, and secret-scan all pass on #120.

Closing this PR so the two implementations do not compete. Thank you @inaor for the original fix and investigation that #120 built on.

PR #57, opened 2026-07-30
docs: install from PyPI, and fix the broken image on pypi.org

sklinglernv:
Two fixes, both consequences of `README.md` doubling as the PyPI `long_description`.

## 1. Install from PyPI

All four packages went live with `v0.0.7`, so `git+https://…` is no longer the primary install path.

```bash
uv add nooa                 # or: pip install nooa
uv add "nooa[cli,memory]"   # siblings as extras
```

The sibling packages are documented both by name and as extras, with a table of what each adds. From-source instructions are kept in a collapsed `<details>` for anyone tracking `main` or pinning a tag. `eval_pipeline` stays a git install — it isn't published.

## 2. Broken image on pypi.org

The header used repo-relative paths:

```html
<img src="assets/nvidia-labs-object-oriented-agents-light.svg">
```

PyPI renders the `long_description` standalone, with no repo alongside it, so `assets/…` 404s. That's the broken image right below "Project description" on <https://pypi.org/project/nooa/>.

Five doc links had the same problem: `AGENTS.md`, `CONTRIBUTING.md`, `LICENSE`, `THIRD_PARTY_NOTICES.md`, `examples/README.md`.

All now absolute. GitHub resolves relative and absolute alike, so this works in both places, and the dark/light `<picture>` switching is unaffected — only the URLs changed, not the markup.

## ⚠️ This does not fix the 0.0.7 page

The `long_description` is baked into the uploaded artifact, and PyPI does not accept re-uploads of an existing version. **The corrected README first appears on the next release.** The 0.0.7 page will keep its broken image permanently.

## Verified

- All 7 new URLs return `200`
- Built wheel `METADATA` contains **0** `assets/` relative paths and 3 absolute ones
- `twine check` passes on wheel + sdist
- No relative links remain in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,18 +2,21 @@
 
 <br />
 
+<!-- Absolute URLs, not repo-relative paths: this README is also the PyPI
+     long_description, and PyPI renders it standalone with no assets/
+     directory alongside it, so relative paths 404 there. -->
 <picture>
   <source
     media="(prefers-color-scheme: dark)"
-    srcset="assets/nvidia-labs-object-oriented-agents-dark.svg"
+    srcset="https://raw.githubusercontent.com/NVIDIA-NeMo/labs-OO-Agents/main/assets/nvidia-labs-object-oriented-agents-dark.svg"
   >
   <source
     media="(prefers-color-scheme: light)"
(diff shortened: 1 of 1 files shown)


alessiodevoto (approved):
Thanks for fixing this!

PR #62, opened 2026-07-30
feat(cli): let external packages add nooa subcommands via entry points

sklinglernv:
Adds a `nooa_cli.commands` entry-point group so packages outside this repo can contribute `nooa` subcommands, instead of shipping standalone console scripts.

```toml
[project.entry-points."nooa_cli.commands"]
tui = "my_package.cli.tui:command"
```

`discover_commands()` is now the existing built-in scan plus an entry-point pass. Built-ins are yielded first and win name collisions; each name is yielded once. Plugin failures (broken metadata, failing import, non-`click.Command`) are logged and skipped — the built-in scan still raises, since that's a bug in this repo. Plugins are sorted by entry-point name so `--help` doesn't depend on install order.

Mirrors the existing `nooa.skills` and `nooa.bundled_configs` groups.

### Also included

The new tests wouldn't have run: `testpaths` pointed at `packages/nooa-cli/tests/{cli,integration}`, neither of which exists, so nothing under `packages/nooa-cli/tests` was collected in CI — `test_cli_smoke.py` included. Pointing it at the parent exposed a collision: the empty `packages/nooa-cli/tests/__init__.py` made that a package named `tests`, shadowing the root `tests` package. Deleted it; `packages/nooa-bench/tests` already ships without one.

### Verification

11 new tests. Full suite 6385 passed / 5 skipped; ruff clean; `nooa --help` unchanged.

### DIFF
--- a/packages/nooa-cli/src/nooa_cli/AGENTS.md
+++ b/packages/nooa-cli/src/nooa_cli/AGENTS.md
@@ -101,6 +101,35 @@ def command(args: tuple[str, ...]):
     subprocess.run([sys.executable, "-m", "some_other_tool", *args])
 ```
 
+## Commands From Another Package
+
+If your command lives in a *different* installed package (not in this repo),
+register it in the `nooa_cli.commands` entry-point group instead of dropping a
+file here. The entry-point name becomes the subcommand name:
+
+```toml
+# pyproject.toml of your package
+[project.entry-points."nooa_cli.commands"]
+tui  = "my_package.cli.tui:command"
+term = "my_package.cli.term:command"
+```
+
+`my_package.cli.tui:command` must be a `click.Command` or `click.Group` — the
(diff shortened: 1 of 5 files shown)


src/nooa/tracing/_litellm_patch.py
```diff
@@ -249,17 +255,144 @@ def _patched_get_attributes_from_message_param(
         yield (MessageAttributes.MESSAGE_TOOL_CALL_ID, tool_call_id)
 
 
+# ---------------------------------------------------------------------------
+# Patch 3: drop the redundant second copy of the prompt from LLM spans.
+# ---------------------------------------------------------------------------
```
furgalep: Is this supposed to be in this PR?

sklinglernv: no this should not be here. Thanks for catching it

PR #72, opened 2026-08-01
fix: surface prompt template eval errors, fail-safe async_safety fallback, and set _parent_agent_var in sync wrapper

arnabwithab:
## What does this PR do?

Three defensive fixes targeting silently swallowed errors and a sync-wrapper
parity gap:
1. **agent.py** — `Agent._resolve_docstring_template()` catches `eval()`
   failures with a bare `except Exception:` and silently falls back to raw
   `{placeholder}` text. If a user has a typo in `{self.attr}`, the broken
   expression ships in the prompt with zero feedback. Added
   `logger.debug(..., exc_info=True)` inside the except block so template
   errors are visible at debug level.

2. **async_safety.py** — `_is_event_loop_thread()` returns `True` when it
   cannot determine the thread (missing `_thread_id` attribute). This is
   fail-dangerous: it would block legitimate cross-thread `Future.result()`
   calls. Changed fallback to `False` so calls are allowed through when we
   can't verify.

3. **method_wrapper.py** — `create_sync_agent_method_wrapper` never calls
   `_parent_agent_var.set(self)`, unlike its async counterpart (line 183).
   Subagents instantiated inside a sync method cannot inherit the parent's
   LLM. Added `parent_token = _parent_agent_var.set(self)` after the
   call-id push and `_parent_agent_var.reset(parent_token)` in `finally`,
   mirroring the async wrapper.

## Related issues
N/A

## Checklist

### Checklist notes
- **ruff:** Clean on all three changed files.
- **pyright:** 0 errors, 0 warnings.
- **Tests:** No new tests required — each change is an internal quality fix
  (debug logging, boolean flip, contextvar parity with existing async path).
  Existing test suite covers all modified code paths.
- **Docs:** No public API changes. No new files created.

## Summary by CodeRabbit

- **Bug Fixes**
  - Synchronous agent calls now correctly isolate scoped context when crossing agent boundaries while preserving context for calls within the same agent.
  - Added diagnostic logging when system prompt placeholders cannot be resolved, while preserving the original placeholder text as a fallback.
- **Documentation**
  - Clarified event-loop safety behavior for preventing blocking operations in unsupported runtime conditions.
  - Improved clarity around how runtime safety checks determine whether operations can proceed safely.


alessiodevoto (commented):
@dcoapp recheck

arnabwithab:
Hi @alessiodevoto . What's up?

alessiodevoto:
@coderabbit full review

arnabwithab:
@alessiodevoto looking forward to your own thoughts on the small change!

src/nooa/runtime/method_wrapper.py
```diff
@@ -453,9 +434,12 @@ def wrapper(self: Any, *args: Any, **kwargs: Any) -> Any:
         # nearest traced ancestor — same semantics as the async wrapper.
         _push_agent_call_id(call_id if _tracing_enabled[0] else parent_call_id)
 
-        # Same agent-call event as the async wrapper. The sync wrapper doesn't
-        # set _parent_agent_var, so read it directly for is_top_level.
-        is_top_level = _parent_agent_var.get() is None
+        current_parent = _parent_agent_var.get()
```
alessiodevoto, line 482: This sync wrapper now sets `_parent_agent_var`, but unlike the async wrapper it does not clear `_scoped_blocks_var` / `_scoped_events_var` when entering a different agent. That means a parent `ScopedContext` leaks into child sync methods, even though the async wrapper explicitly prevents scoped context from crossing agent boundaries.

Could you mirror the async wrapper here by detecting `current_parent is not None and current_parent is not self`, setting both scoped context vars to `None` for that call, and resetting them in `finally`? A sync-specific regression test should cover this.

alessiodevoto:
@coderabbit review

alessiodevoto:
@coderabbit full review

alessiodevoto (approved):
LGTM, thanks for your contribution @arnabwithab !

PR #79, opened 2026-08-03
fix(mcp): keep environment placeholders literal

furgalep:
## Summary

- remove `${VAR}` expansion from `.mcp.json` and inline/TUI MCP server configuration
- keep all repository-provided URL, header, argument, and child-environment values literal
- deep-copy selected server configuration before transport creation to preserve mutation isolation
- require trusted caller code to resolve and pass any intended secret values explicitly
- cover HTTP and stdio leak sinks, attacker-supplied expansion flags, unset variables, and inline configuration
- add real local stdio and streamable-HTTP discovery/invocation tests
- document the security change and migration path

## Security validation

A reporter-aligned PoC using a fake `OPENAI_API_KEY` canary was rerun through the exact `from nooa.mcp import MCPManager` path. It confirmed that the canary does not reach:

- arguments passed to `create_mcp_client()`
- the URL path or `Authorization` header observed by a loopback HTTP listener
- the environment observed by a spawned stdio child

An attacker-supplied `"expand_env_vars": true` field is inert. The expansion helper and caller opt-in API have been removed entirely.

## Test plan

- `pytest tests/test_mcp -q` — 93 passed
- real stdio and loopback streamable-HTTP discovery/invocation — passed
- Ruff lint and format checks — passed
- `git diff --check` — passed

The real transport tests use the bundled wiki MCP server and require no external services, credentials, or network access beyond loopback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,3 +7,6 @@ to follow semantic versioning.
 ## [Unreleased]
 
 - Initial public release of NVIDIA Object-Oriented Agents (NOOA).
+- Security: MCP server configurations no longer expand host environment variables
+  from `${VAR}` placeholders. Trusted caller code must resolve secrets and pass
+  their values explicitly.
--- a/examples/README.md
+++ b/examples/README.md
@@ -503,6 +503,22 @@ confluence_tool = MCPManager.create_from_server(
 )
 ```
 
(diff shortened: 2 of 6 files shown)


sklinglernv (approved):
LGTM, thanks for the fix!

PR #81, opened 2026-08-03
feat(strategy): allow a callable for @strategy(llm=...) 

sklinglernv:
## Motivation

There was no way to choose a **per-method** LLM at **agent-construction time**. `MyAgent(llm=...)` is per-instance but hits every method; `@strategy(llm=client)` is per-method but is baked onto the function at import time, so every instance shares it.

## What changed

`@strategy(llm=...)` now also accepts a callable taking the agent and returning a client, resolved on each generation call:

```python
class Researcher(Agent, llm=fast):
    @strategy(llm=lambda self: self.big)              # varies per instance
    async def analyze(self, doc: str) -> str: ...

    @strategy(llm=lambda self: self.big if self.retries > 2 else fast)
    async def solve(self, problem: str) -> str: ...   # varies per call
```

A string alias (`llm="reasoner"`) was considered and rejected: it needs a role map with its own MRO cascade, and is a static special case of a callable — it can't express the second example. Cost is that typos fail at first call, so resolver failures are wrapped to name the method.

## Adjacent bug fix (worth reviewing separately)

`ArgumentValidator` stripped `_session_locals` from call kwargs but not `llm` or `_strategy`, though `_execute_with_generation` pops all three — so `await agent.method(x, llm=other)` was rejected as an unexpected kwarg before reaching the resolution code. The precedence `actor.py` documents was unreachable for direct calls.

The strip is scoped to the agent path only; the strategy-helper path routes to `execute_nested()`, which pops nothing.

## Reviewer notes

- **Duck-typed clients still work** — detection probes for `acall` rather than requiring a `UnifiedLLM` subclass, which several fakes rely on. Classes are excluded from the probe first, since `hasattr(SomeClass, "acall")` is True for the unbound function.
- **Known gap, not fixed:** a method declaring its own `llm` parameter loses that argument. Broken on `main` too; nothing in the repo declares one. Fixing it means implementing the "Reserved Parameters" check `AGENTS.md` describes but no code enforces — happy to do a follow-up.
- **Pre-commit `pyright`** fails on a pre-existing error at `actor.py:1509`, surfaced only because this PR touches that file. CI runs ruff + pytest, so the PR should be green.

## Testing

24 new tests in `tests/test_method_llm_callable.py`. Full suite: 6531 passed — the `test_nemo_relay_middleware.py` failures reproduce identically on `main`. `ruff` clean; `pyright` clean on all changed files. Subagent-reviewed; findings folded in.

Docs updated in `skills/nooa-agent-authoring/SKILL.md` and `examples/README.md`, which also omitted the call-level override.

### DIFF
--- a/src/nooa/decorators.py
+++ b/src/nooa/decorators.py
@@ -15,8 +15,8 @@ from nooa.ellipsis_detection import has_ellipsis_body
 
 if TYPE_CHECKING:
     from nooa.config.truncation_config import TruncationConfig
-    from nooa.method_llm import MethodLLM
     from nooa.strategies import GenerationStrategy as GenerationStrategyABC
+    from nooa.unifiedllm import UnifiedLLM
 
 P = ParamSpec("P")
 R = TypeVar("R")
@@ -26,7 +26,7 @@ def strategy(
     strategy_instance: "GenerationStrategyABC | None" = None,
     context: "ScopedContext | dict[str, Any] | None" = None,
     *,
-    llm: "MethodLLM | None" = None,
+    llm: "UnifiedLLM | Callable[[Any], UnifiedLLM] | None" = None,
     truncation: "TruncationConfig | None" = None,
 ) -> Callable[[Callable[P, R]], Callable[P, R]]:
     """Strategy decorator for agent methods.
--- a/src/nooa/decorators.py
+++ b/src/nooa/decorators.py
@@ -15,8 +15,8 @@ from nooa.ellipsis_detection import has_ellipsis_body
 
 if TYPE_CHECKING:
     from nooa.config.truncation_config import TruncationConfig
+    from nooa.method_llm import MethodLLM
     from nooa.strategies import GenerationStrategy as GenerationStrategyABC
-    from nooa.unifiedllm import UnifiedLLM
 
 P = ParamSpec("P")
 R = TypeVar("R")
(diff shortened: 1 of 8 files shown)


src/nooa/decorators.py
```diff
@@ -26,7 +26,7 @@ def strategy(
     strategy_instance: "GenerationStrategyABC | None" = None,
     context: "ScopedContext | dict[str, Any] | None" = None,
     *,
-    llm: "UnifiedLLM | None" = None,
+    llm: "MethodLLM | None" = None,
```
sklinglernv: why do we need this new type? I find this much less readable htan `UnifiedLLM | Callable[[Any], UnifiedLLM]` which is much more explicit

sklinglernv: Agreed — inlined in 4c723eb. The alias had exactly one use site, and being `TYPE_CHECKING`-only and unexported it was not nameable by downstream typed code anyway, so it bought indirection and nothing else. The signature now reads `UnifiedLLM | Callable[[Any], UnifiedLLM] | None`. 🤖🤖🤖

furgalep:
You have to use a lambda? No way you can do 
`@strategy(llm=self.big)`
?

If not, I get it, it's fine.

furgalep (approved):

sklinglernv:
> You have to use a lambda? No way you can do `@strategy(llm=self.big)` ?
> 
> If not, I get it, it's fine.

I don't think there is an easy way, `self` is not defined in the decorator context, we would have to use some trickery or go back to a string based "self.big" approach like we do for dynamic context blocks.

PR #83, opened 2026-08-03
docs(examples): add Ollama and vLLM to the model selection snippet

Yeriahz:
## What does this PR do?

Adds Ollama and vLLM to the `get_llm_client()` snippet in `examples/README.md`, matching the local-model examples already shown in the main README. Documentation only.

Originally this also proposed a `NEMO_OO_MODEL` / `NEMO_OO_API_BASE` override in `quickstart.py`. Dropped per review: it would have been the only env-driven model selection in the codebase, and `get_llm_client()` plus the `llm_config` registry is the framework's canonical pattern. See the review thread below.

## Related issues

None.

## Validation

Docs only, no code changed.

## Checklist

- [ ] Code follows the project style (`uv run ruff check .` and `uv run ruff format --check .` pass) — no code changed
- [ ] Tests added/updated and passing (`uv run pytest`) — no code changed
- [ ] New source files carry an SPDX license header — no new files

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -53,7 +53,18 @@ from nooa.unifiedllm import get_llm_client
 llm = get_llm_client("gpt-4o-mini")                 # OpenAI (needs OPENAI_API_KEY)
 llm = get_llm_client("claude-sonnet-4-5-20250514")  # Anthropic (needs ANTHROPIC_API_KEY)
 llm = get_llm_client("gemini/gemini-2.5-flash")     # Google (needs GEMINI_API_KEY)
+llm = get_llm_client("ollama_chat/qwen3:1.7b",      # Ollama (no key)
+                     api_base="http://localhost:11434")
 ```
+
+The checked-in examples take their model from the environment. Set `NEMO_OO_MODEL` to any litellm name, plus `NEMO_OO_API_BASE` for a local server:
+
+```bash
+NEMO_OO_MODEL="ollama_chat/qwen3:1.7b" NEMO_OO_API_BASE="http://localhost:11434" \
+  uv run python examples/quickstart/01_first_generation_method.py
+```
+
+Without it, the examples select whichever provider credential is present (`NVIDIA_API_KEY`, `OPENAI_API_KEY`, ...).
 
 Provider packages can register bundled model aliases automatically through the `nooa.bundled_configs` entry-point group. To customize the registry, run `nooa config eject`, drop an `llm_config.yaml` in your project's `.nooa/`, or point `NEMO_OO_LLM_CONFIG` at one or more YAML files. See [`src/nooa/unifiedllm/registry.py`](../src/nooa/unifiedllm/registry.py) for the YAML schema.
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -55,16 +55,9 @@ llm = get_llm_client("claude-sonnet-4-5-20250514")  # Anthropic (needs ANTHROPIC
 llm = get_llm_client("gemini/gemini-2.5-flash")     # Google (needs GEMINI_API_KEY)
 llm = get_llm_client("ollama_chat/qwen3:1.7b",      # Ollama (no key)
                      api_base="http://localhost:11434")
+llm = get_llm_client("hosted_vllm/Qwen/Qwen3-1.7B", # vLLM (no key)
+                     api_base="http://localhost:8000/v1")
 ```
-
-The checked-in examples take their model from the environment. Set `NEMO_OO_MODEL` to any litellm name, plus `NEMO_OO_API_BASE` for a local server:
-
-```bash
-NEMO_OO_MODEL="ollama_chat/qwen3:1.7b" NEMO_OO_API_BASE="http://localhost:11434" \
-  uv run python examples/quickstart/01_first_generation_method.py
-```
-
-Without it, the examples select whichever provider credential is present (`NVIDIA_API_KEY`, `OPENAI_API_KEY`, ...).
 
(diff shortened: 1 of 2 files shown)


alessiodevoto:
HI @Yeriahz ! Thanks for opening this PR 🙂 

I have two comments:
- I would keep the Ollama line you added to the` get_llm_client()` snippet in ` examples/README.md`. That's a useful addition to an already-documented pattern. We could also add the vLLM path (like we have in the main README).

- As for the new env var mechanism, I would drop the NEMO_OO_MODEL / NEMO_OO_API_BASE mechanism in `quickstart.py,` and the paragraph documenting it. `examples/README.md` already documents way to "run the examples against a different model" three lines above your new paragraph. That's the framework's canonical model-selection pattern. Adding an env-var override on top of that would make quickstart.py the only place in the codebase where model choice is env-driven — no *_MODEL / *_API_BASE env var is used anywhere else. I'd rather not introduce a parallel mechanism scoped just to the demos when the primary one already works and is documented.

Would you be up for trimming the PR to just the README addition? 

Also: the traceback-clarity issue you flagged at the bottom would be a great follow-up if you want to pick it up separately!

examples/README.md
```diff
@@ -53,8 +53,19 @@ from nooa.unifiedllm import get_llm_client
 llm = get_llm_client("gpt-4o-mini")                 # OpenAI (needs OPENAI_API_KEY)
 llm = get_llm_client("claude-sonnet-4-5-20250514")  # Anthropic (needs ANTHROPIC_API_KEY)
 llm = get_llm_client("gemini/gemini-2.5-flash")     # Google (needs GEMINI_API_KEY)
+llm = get_llm_client("ollama_chat/qwen3:1.7b",      # Ollama (no key)
```
alessiodevoto, line 56: Keep and ideally add vLLM

examples/README.md
```diff
@@ -53,8 +53,19 @@ from nooa.unifiedllm import get_llm_client
...
+                     api_base="http://localhost:11434")
 ```
 
+The checked-in examples take their model from the environment. Set `NEMO_OO_MODEL` to any litellm name, plus `NEMO_OO_API_BASE` for a local server:
```
alessiodevoto: I would remove the new ENV VARs as it is as they do not follow the framework convention for model selection

src/nooa/util/quickstart.py
```diff
@@ -26,10 +26,18 @@
 #   * OPENAI_API_KEY           -> OpenAI (public)
 #   * NVIDIA_INFERENCE_API_KEY -> NVIDIA internal inference gateway
 #                                 (inference-api.nvidia.com; NVIDIA employees)
-# To use a specific model, set MODEL to any litellm name and provide its key,
-# e.g. MODEL = "claude-haiku-4-5" with ANTHROPIC_API_KEY.
+# To pin a specific model, set NEMO_OO_MODEL to any litellm name and provide its
```
alessiodevoto: Same thing here, once we document ollama in the readme, no need for this

alessiodevoto (changes requested):

Yeriahz:
Thanks for the review. Trimmed to just the README addition, with vLLM added alongside Ollama. `quickstart.py` is back to `main`, and I've retitled the PR to match what it actually does now.

Your point about not introducing a second model-selection mechanism scoped to the demos is fair, and I'd rather not be the reason that exists.

One thing I'd still like to understand, since it's what sent me down this path: is there a supported way to point a checked-in example at a local model? `get_llm_client()` is what I'd reach for writing my own agent, but the examples import `llm` from `nooa.util.quickstart`, so as far as I could tell running `examples/quickstart/01_first_generation_method.py` against Ollama meant editing that file. If there's a registry or config path I missed, happy to document it as a follow-up.

And yes, I'll pick up the traceback clarity issue separately.

Yeriahz:
Followed up on the traceback issue. It turned out not to be a NOOA problem.

NOOA's retry classifier is doing the right thing: 500 is in the default `retryable_status_codes`, so a 500 gets retried. The misclassification is upstream. The OpenAI SDK raises `OpenAIError` at client construction with no `status_code`, litellm's handler defaults a missing status to 500 (`llms/openai/openai.py:983`), and `exception_mapping_utils.py` maps that to `InternalServerError`.

It reproduces without NOOA:

```python
import litellm  # no OPENAI_API_KEY set
litellm.completion(model='gpt-4o-mini', messages=[{'role':'user','content':'hi'}])
# InternalServerError: OpenAIException - Missing credentials...
```

Filed upstream: https://github.com/BerriAI/litellm/issues/35860

There may still be something worth doing on the NOOA side, since failing at client construction would beat failing on the first call, but that is a separate question from the misclassification and I did not want to assume it is wanted.

Yeriahz:
@alessiodevoto friendly ping on this one. The requested changes are
addressed as of d42453c: quickstart.py is back to main and the diff is
now four lines in examples/README.md, adding the Ollama and vLLM lines
that the top-level README already documents. No code changed.

Merging is still blocked by the changes-requested review, and there's a
workflow awaiting maintainer approval, so both need a hand from your
side when you get a chance.

Still curious about the question above if you have a moment: is there a
supported way to point a checked-in example at a local model? The
examples import llm from nooa.util.quickstart, so running
examples/quickstart/01_first_generation_method.py against Ollama meant
editing library source. Happy to document the right path as a follow-up
if I missed one.

alessiodevoto:
Hi @Yeriahz thanks for updating! That's actually what the paragraph directly above the lines you're adding:  f`rom nooa.util.quickstart import *` gives you the convenience llm, but if you want a specific model (local or otherwise) the supported path is to instantiate `get_llm_client(...)` yourself in the example.  After this merges the local-model case is covered by the pattern you are adding!

alessiodevoto (approved):

Issue #86, opened 2026-08-04
Design: value-preview extractor registry for pformat (numpy/pandas structural previews)


pierreolivierbonin:
## Motivation

Sec 3.2 of the paper names bounded-preview coverage as open work: *"Finding even better formats that are obvious to LLMs, and supporting more types, remains open work."* Today numpy arrays and pandas frames — the types a data-touching CodeAct agent handles most — hit `_pformat`'s truncated-repr fallback, which slices repr text mid-escape and hides shape, dtype, and column schema:

```
ndarray(repr_len=10250, [:100]='array([[   0. ,    1.5,    3. ,    4.5],\n       [   6. , ...')
```

## Proposed design (implementation ready)

A **value-preview extractor registry** mirroring the existing `doc()` adapter architecture:

- `spec.define_preview(SomeType)` registers `(instance, PreviewBudget) -> str | None`; MRO-aware lookup, thread-safe, cleared by `clear_registry()`.
- `_pformat` consults it just before the repr fallback (and at the `max_depth` shallow path, which today writes full untruncated reprs). `None` or an exception declines → existing behavior. Zero change for unregistered types.
- numpy/pandas adapters ship in `agentdoc/adapters/`, registered via `register_all()`; neither library becomes a dependency. Formats stay in the truncation 3.0 marker family:

```
ndarray(shape=(250, 4), dtype=float64, [:3]=[[0., 1.5, 3., 4.5], ...], [-3:]=[...])
DataFrame(shape=(500, 3), columns={'price': 'float64', 'qty': 'int64', 'label': 'str'}, [:3]=[{...}], [-2:]=[{...}])
```

- Completeness triggers use total element counts (`arr.size`, `df.size`) because numpy/pandas elide their own reprs with `...` — a short repr is not evidence of a complete value.
- CodeAct registers adapters at execution start; previously `register_all()` ran only on the opaque-return-type path, so schemable-return methods never activated adapters (found via a live agent run; regression-tested).

## Design questions for maintainers

1. Separate `define_preview` registry vs. overloading `define_doc` extractors with a preview capability?
2. Is the two-field `PreviewBudget` (max_length, max_string) the right contract, or should previews see the full truncation config?
3. Preview format opinions welcome — e.g. whether DataFrame row samples should be records dicts (current) or a columnar form.

PR with the full implementation, 40 tests, and live-run evidence: #87

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #87, opened 2026-08-04
feat(agentdoc): shape-aware pformat previews for numpy and pandas

pierreolivierbonin:
## Motivation

Section 3.2 of the paper calls bounded-preview coverage open work: *"Finding even better formats that are obvious to LLMs, and supporting more types, remains open work."* Today, numpy arrays and pandas frames — the two types a data-touching CodeAct agent handles most — fall through to the truncated-repr fallback, which slices the repr text mid-escape and hides the structural facts:

```
ndarray(repr_len=10250, [:100]='array([[   0. ,    1.5,    3. ,    4.5],\n       [   6. , ...')
DataFrame(repr_len=323, [:100]='        price  qty label\n0    0.000000    0     x\n1  ...')
```

The shape, dtype, and column schema — exactly what the model needs to write its next cell — are invisible or accidental.

## What this PR does

**1. A value-preview extension point mirroring the existing `doc()` adapter architecture.** `spec.define_preview(SomeType)` registers an extractor (MRO-aware, thread-safe, cleared by `clear_registry()`) that `pformat`/`pprint` consults before the repr fallback. Extractors receive the instance plus the active truncation budget and may decline by returning `None` (raising also declines), so behavior is unchanged for every unregistered type and for complete-within-budget values — the marker-presence-signals-truncation rule is preserved.

**2. numpy + pandas adapters** (`agentdoc/adapters/numpy.py`, additions to `adapters/pandas.py`), registered via `register_all()` so CodeAct rendering picks them up automatically. Neither library becomes a dependency. With the runtime's default budgets:

```
ndarray(shape=(250, 4), dtype=float64, [:3]=[[0., 1.5, 3., 4.5], ...], [-3:]=[..., [1494., 1495.5, 1497., 1498.5]])
DataFrame(shape=(500, 3), columns={'price': 'float64', 'qty': 'int64', 'label': 'str'}, [:3]=[{'price': 0.0, 'qty': 0, 'label': 'x'}, ...], [-2:]=[...])
Series(len=500, dtype=float64, name='price', [:3]=[0.0, 0.002, 0.004], [-2:]=[0.998, 1.0])
```

Formats stay inside the truncation 3.0 marker family (`[:H]=` / `[-T:]=` slice keys). Details worth reviewing:

- The count trigger uses **total element counts** (`arr.size`, `df.size`): numpy and pandas elide their own reprs with `...`, so a short repr is not evidence of a complete value (a 2×1M array's repr is short but far from complete).
- Row/column samples are bounded per axis (≥2-D arrays get one edge item per deep axis; frames cap row records and per-cell strings), so previews stay a few hundred characters no matter the value's size, and nothing over-materializes (the 1000×3000 test asserts this).
- Previews also apply at the `max_depth` shallow path, which previously wrote **full untruncated reprs** for values nested at the depth limit.

**3. Adapter activation at CodeAct execution start** (`codeact.py`). Live end-to-end testing (Ollama `qwen3:1.7b`, a real CodeAct run over a 500-row DataFrame + 250×4 ndarray) caught that `register_all()` previously ran only on the opaque-return-type path (`_render_return_type_doc`), so a method with a schemable return type (`-> str`) never activated the adapters and argument inspection fell back to truncated reprs. `_run_generation` now registers adapters up front (idempotent, installed-gated, exception-guarded).

**4. Tests** — 40 new tests in four files: registry mechanics with plain Python types (no numpy needed), per-adapter suites using the repo's `importorskip` + adapter-reload pattern, and an end-to-end regression test asserting a CodeAct run's prefill inspection renders `DataFrame(shape=...)` rather than `repr_len=...`.

## Validation

- `pytest tests/agentdoc tests/strategies` — 1641 passed (40 new)
- Full `-m "not integration and not stress"` suite green
- `ruff check`, `ruff format --check`, SPDX header check green
- Verified live: a CodeAct agent on Ollama `qwen3:1.7b` received the structural previews in its first `PythonOutput` event

Closes the design discussion in #86.

## Summary by CodeRabbit

- **New Features**
  - Added bounded value previews for large NumPy arrays, pandas DataFrames, and Series.
  - Added preview registration APIs with configurable formatting budgets.
  - Large values now show structural details and head/tail samples instead of excessively long representations.
  - Added automatic adapter registration when supported libraries are installed.
- **Bug Fixes**
  - Improved CodeAct inspections to display useful previews for large tabular values.
  - Preview failures safely fall back to standard rendering.

### DIFF
--- a/src/nooa/agentdoc/_docs.py
+++ b/src/nooa/agentdoc/_docs.py
@@ -262,6 +262,31 @@ class Spec:
 
         return register_type_info_extractor(target)
 
+    def define_preview(self, target: type) -> Any:
+        """Define a custom value preview for *target* in ``pformat``/``pprint``.
+
+        Replaces the truncated-``repr`` fallback for instances of *target*
+        (and subclasses) with a structural, bounded preview. The extractor
+        receives the instance and the active ``PreviewBudget``; returning
+        ``None`` declines, falling back to the default repr rendering —
+        return ``None`` whenever the plain repr fits the budget, so complete
+        values keep rendering as plain literals::
+
+            from nooa.agentdoc.ext import PreviewBudget
+
+            @spec.define_preview(np.ndarray)
+            def _(arr, budget: PreviewBudget) -> str | None:
+                r = repr(arr)
+                if budget.max_string is None or len(r) <= budget.max_string:
+                    return None
+                return f"ndarray(shape={arr.shape}, dtype={arr.dtype}, ...)"
+
+        ``doc()`` rendering is unaffected — use :meth:`define_doc` for that.
+        """
+        from nooa.agentdoc.registry import register_preview_extractor
+
+        return register_preview_extractor(target)
+
 
 spec = Spec()
 
--- a/src/nooa/agentdoc/_pformat.py
+++ b/src/nooa/agentdoc/_pformat.py
@@ -1402,6 +1402,23 @@ def _format_value(
         )
         return
 
+    # Registered value previews (numpy arrays, pandas frames, …): a preview
+    # extractor renders a structural, bounded preview instead of the truncated
+    # repr below. Returning None (or raising) declines and falls through.
+    from nooa.agentdoc.registry import PreviewBudget, get_preview_extractor
+
+    preview_extractor = get_preview_extractor(_object)
+    if preview_extractor is not None:
+        try:
+            preview = preview_extractor(
+                _object, PreviewBudget(max_length=max_length, max_string=max_string)
+            )
+        except Exception:
+            preview = None
+        if preview is not None:
+            _stream.write(preview)
+            return
+
     # Fallback to repr for other types (numpy arrays, pandas frames, custom classes, …).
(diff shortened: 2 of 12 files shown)


sklinglernv:
@CodeRabbit review full

sklinglernv:
@pierreolivierbonin thank you very much for the contribution and the new feature!

From my side two notes on the PR:

- NumPy previews do not actually remain bounded. Sampled elements use unrestricted repr(). With max_string=500, an array containing long strings produced a 100,091-character preview in my reproduction. Since these previews feed CodeAct context, this undermines the main safety property of the feature.
- Pandas previews corrupt valid DataFrames with duplicate column names. Both the dtype dictionary and to_dict("records") collapse duplicate labels, so the preview silently drops columns and emits warnings.


And then a general note (not related to the changes of the PR):
- I don't think we should handle registering preview and doc in codeact.py . Both features are valuable outside of codeact

pierreolivierbonin:
Hey there! Please let me know if there is anything left to do for this PR.

PR #88, opened 2026-08-04
ci: run sandbox containment tests, split off the sandbox marker

### LINKED ISSUE #78: Sandbox containment tests never run in CI (excluded by the integration marker)
`ci.yml:38` runs:

```
uv run pytest -q -m "not integration and not stress"
```

That excludes `tests/runtime/sandbox/`, 46 tests covering the per-cell OS sandbox: Landlock filesystem confinement, the seccomp network block, RLIMIT caps, cell timeouts, and the broker-pipe boundary. These are the end-to-end tests of the containment layer described in the paper's Appendix D.2 and pointed to by the safety note in `README.md`. It is the only pytest invocation in `.github/workflows/`.

They need no credentials and take 22 seconds:

```
uv run pytest tests/runtime/sandbox/ -m integration -q
46 passed, 23 deselected in 22.48s
```

**Why they look excludable but aren't.** Twelve test files carry `pytest.mark.integration`. Six are this sandbox suite. The other six are live-provider tests: `unifiedllm/test_responses_client.py`, `external/test_providers.py`, `integration/test_predict_return_types_live.py`, `integration/test_l4_eviction_e2e.py`, `integration/test_archival_on_error_e2e.py`, `provider_compat/test_provider_compat.py`. The marker means "needs credentials" for that group and "forks a real worker" for the sandbox group, and the blanket exclusion catches both.

**They are built to survive a runner without the guards.** Every `SandboxConfig` in `test_executor.py` passes `require=False`, and the four tests that need a specific mechanism carry `skipif` on `CAPS.rlimit`, `CAPS.seccomp`, and `CAPS.landlock_abi`. On a kernel lacking Landlock or seccomp they skip rather than fail. The one test using `require=True` monkeypatches a fake capability set.

**What I could not verify.** I have no access to a CI log for this repo, so I cannot confirm these pass on `ubuntu-latest`. I am reporting that they are structured to degrade gracefully there, not asserting the run is green.

Related to #65, which reports a different mechanism with the same effect: tests that quietly do not run.

Happy to submit a PR adding a sandbox step to CI if that is useful.

alessiodevoto:
Fixes #78.

## Problem
`ci.yml` only pytest step ran `-m "not integration and not stress"`. The `integration` marker was overloaded — it meant "needs live credentials" (6 files) *and* "forks a real sandbox worker" (6 files under `tests/runtime/sandbox/`). The blanket exclusion dropped both, so the 46-test sandbox containment suite — Landlock, seccomp, RLIMIT, cell timeouts, broker-pipe boundary.

## Fix
Split the marker:
- `integration` — unchanged, still means "needs live credentials."
- `sandbox` — new, means "forks a real worker for OS-level containment." No credentials needed; `skipif`-degrades on kernels without the relevant capabilities.

CI `test` job gains a second step:

```yaml
- run: uv run pytest -q -m "not integration and not stress and not sandbox"
- run: uv run pytest -q -m sandbox
```

Verification

- pytest -m sandbox --collect-only → 48 selected under tests/runtime/sandbox/.
- Default addopts → same 48 deselected; the 25 unmarked helper tests in that directory still run in the unit-test step.
- Ran locally on Darwin: 39 pass / 7 skip / 2 fail (the failures are
macOS-specific RLIMIT_AS behavior; the sandbox is Linux-only and the
executor already logs sandbox requires Linux). ubuntu-latest should
match the reporter's clean run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,15 @@ jobs:
       - name: Sync
         run: uv sync --all-extras --no-extra sandbox
       - name: Unit tests (no live API calls)
-        run: uv run pytest -q -m "not integration and not stress"
+        run: uv run pytest -q -m "not integration and not stress and not sandbox"
+      # Sandbox containment tests fork a real worker to exercise the OS-level
+      # sandbox (Landlock, seccomp, RLIMIT, cell timeouts, broker-pipe
+      # boundary). They need no credentials and skipif-degrade on kernels
+      # without the relevant capabilities, so ubuntu-latest runs them for
+      # real. Kept out of the unit-test step because forking a worker is
+      # heavier than a unit test. See #78.
+      - name: Sandbox containment tests
+        run: uv run pytest -q -m sandbox
 
   build:
     runs-on: ubuntu-latest
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -221,11 +221,12 @@ pythonpath = [
     "src",
     "packages/nooa-cli/src",
 ]
-addopts = "-v --tb=short --import-mode=importlib -m 'not integration and not stress'"
+addopts = "-v --tb=short --import-mode=importlib -m 'not integration and not stress and not sandbox'"
 markers = [
     "integration: marks tests as integration tests that make real API calls (run with '-m integration')",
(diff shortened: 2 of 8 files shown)


furgalep (approved):
LGTM

PR #89, opened 2026-08-04
fix(ci) - ripregrep skipping

### LINKED ISSUE #65: 145 tests silently skip without ripgrep, and the dependency isn't documented
Running the suite on a fresh Ubuntu setup:

```
uv run pytest -m "not integration and not stress"
6337 passed, 149 skipped
```

145 of those 149 skips are the same reason, all in `tests/tools/test_shell_tools_modern.py`:

```
SKIPPED [45] tests/tools/test_shell_tools_modern.py:334: needs rg and grep on PATH
```

After `apt install ripgrep`, with no other change:

```
6482 passed, 4 skipped
```

All 145 pass. The remaining 4 skips are legitimate (Slack tokens, a real-LLM test, a Python internals note).

The problem is that a skip is indistinguishable from a pass in the summary line. A contributor runs the suite, sees green, and doesn't know they are missing 145 tests of shell-tool coverage.

Nothing in `README.md`, `CONTRIBUTING.md`, `tests/README.md`, or `pyproject.toml` mentions ripgrep. The only references in the repo are in `nooa-cli` source, where `repo_tools.py:502` checks whether `rg` is available and falls back gracefully. The production code degrades; the tests hard-skip.

**What I could not verify.** `.github/workflows/*.yml` has no ripgrep install step, and I do not know whether `ubuntu-latest` ships it preinstalled. If it does not, these tests are not running in CI either. I have no access to a CI log for this repo to check, so I am reporting the local behavior and flagging the CI question rather than asserting it.

Happy to submit a PR adding a prerequisites note to `tests/README.md` and `CONTRIBUTING.md`, and an install step to CI, if that is useful.

alessiodevoto:
Fixes #65.

## Problem
`tests/tools/test_shell_tools_modern.py` skipif-drops all 145 tests when `rg` or `grep` is missing from `PATH`. `ubuntu-latest` does not ship ripgrep, and no workflow installs it — so on every CI run those 145 tests silently skipped and the `test` job passed vacuously.

## Fix
Two layers, so the failure mode can't recur silently:

1. **Install ripgrep in CI** and document it as a test prerequisite in `tests/README.md`. Fixes the current instance.
2. **Session-start check in `tests/tools/conftest.py`** — inspects `PATH`  for `rg`/`grep` at collection start. If either is missing:
   - in CI (`CI=true`): `pytest.exit(...)` → job fails red with a clear
     message rather than passing green with 145 silent skips;
   - locally: emits a `UserWarning` banner but does not block the run.

This closes the underlying "tests silently didn't run" pattern, not just the current instance — a future base-image regression would fail loud instead of erasing coverage.

## Verification
- `pytest tests/tools/test_shell_tools_modern.py --collect-only` → 145 collected with `rg` on `PATH` (matches the reporter's count).
- `CI=true PATH=/usr/bin:/bin pytest ...` → exits red with the prerequisite message. `PATH=/usr/bin:/bin pytest ...` (no CI) → session continues, warning surfaces.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,11 @@ jobs:
           python-version: "3.12"
       - name: Sync
         run: uv sync --all-extras --no-extra sandbox
+      # tests/tools/test_shell_tools_modern.py skipif-drops 145 tests when `rg`
+      # is not on PATH. ubuntu-latest does not ship ripgrep, so without this
+      # step those tests silently skip and the job passes vacuously. See #65.
+      - name: Install ripgrep
+        run: sudo apt-get update && sudo apt-get install -y ripgrep
       - name: Unit tests (no live API calls)
         run: uv run pytest -q -m "not integration and not stress"
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -90,6 +90,14 @@ Viewer main routes, journal and OTLP stores, OTLP ingest, stress coverage, image
 
 ## Running Tests
 
+### Prerequisites
+
+- `ripgrep` (`rg`) and `grep` must be on `PATH`. The 145 tests in
+  `tests/tools/test_shell_tools_modern.py` skipif-away without them, and a
+  clean skip is indistinguishable from a pass in the summary line. Install
+  with `apt install ripgrep` on Debian/Ubuntu or `brew install ripgrep` on
+  macOS.
+
 ```bash
 uv run pytest                          # all tests
 uv run pytest tests/runtime/ -v        # single directory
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -34,16 +34,13 @@ jobs:
           python-version: "3.12"
       - name: Sync
         run: uv sync --all-extras --no-extra sandbox
+      # tests/tools/test_shell_tools_modern.py skipif-drops 145 tests when `rg`
+      # is not on PATH. ubuntu-latest does not ship ripgrep, so without this
```
sklinglernv, line 38: nit: I think we remove the legacy shell tools, maybe we should rename at one point (no modern etc)

sklinglernv (approved):
Thanks for the catch and fix!

PR #91, opened 2026-08-04
chore(release): codify the release process in a single script

sklinglernv:
Releases had no defined procedure and nothing validated them. This adds one path that is the same every time.

## Cutting a release

```bash
git checkout main && git pull
uv run python scripts/make_release.py v0.0.9
```

The script runs, in order: preflight (on `main`, clean, in sync, tag unused) → lint, SPDX headers, unit tests → build 4 wheels and check version == tag → capability diff against the previous release → `gh release create --draft` → `gh release edit --draft=false`, which triggers `publish.yml`.

It stops at two prompts: after the capability report, and after the draft notes. Nothing irreversible happens before both.

## Capability diff

Both arms run fresh — HEAD in the working tree, the previous tag in a temporary worktree. A stored baseline cannot distinguish a real regression from the endpoint behind a model alias changing.

The only hard threshold is a floor on the stable tier (60%), which requires typing `OVERRIDE` to pass. Collapses, new error types and drops beyond ±5 points are reported for a human to judge, since LLM pass rates vary run to run. An arm where more than half the samples error is rejected as infrastructure failure rather than reported as a result. The report is appended to the release notes.

## Rehearsing

```bash
uv run python scripts/make_release.py v0.0.9 --checks-only \
  --models claude-haiku --runs 1 --limit 1
```

`--models`/`--runs`/`--limit` require `--checks-only` so they cannot wave a real release through. `--dry-run` prints the `gh` commands. Report logic is covered by `tests/test_make_release.py`; everything above was verified end to end against both arms.

## Also

- SPDX header check now skips `.worktrees/`, which scanned other branches' checkouts and failed only for developers running it locally.
- `RELEASING.md` trimmed from 277 lines to 76.
- `uv.lock` gains an `sdist` line for nemo-relay that `uv run` re-adds on every invocation; committing it stops pre-commit failing.

### DIFF
--- a/scripts/make_release.py
+++ b/scripts/make_release.py
@@ -547,12 +547,21 @@ class Diff:
     beyond_noise: list[str] = field(default_factory=list)
     added: list[str] = field(default_factory=list)
     removed: list[str] = field(default_factory=list)
+    models_changed: list[str] = field(default_factory=list)
     floor_breach: str | None = None
     markdown: str = ""
 
     @property
     def clean(self) -> bool:
-        return not (self.regressions or self.new_errors or self.beyond_noise or self.floor_breach)
+        # `removed` counts, `added` does not: a test that vanished may be hiding
+        # a regression, whereas a new test has no baseline to regress from.
+        return not (
+            self.regressions
+            or self.new_errors
+            or self.beyond_noise
+            or self.removed
+            or self.floor_breach
+        )
 
 
 def _mark(delta: float, *, inverse: bool = False) -> str:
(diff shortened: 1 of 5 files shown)


scripts/make_release.py
```diff
@@ -0,0 +1,965 @@
...
+            f"  {DIM}$ gh release edit {args.tag} --notes-file <notes + capability report>{RESET}"
+        )
+        print(f"  {DIM}$ gh release edit {args.tag} --draft=false{RESET}")
+        ok("dry run complete — nothing was created")
+        return 0
+
+    create_draft(args.tag, head_sha, report)
+    print(f"\n{DIM}Review the generated notes in the browser before continuing.{RESET}")
+    if not confirm(f"Publish {args.tag} to GitHub and PyPI?"):
+        print(f"Aborted. The draft remains; delete with: gh release delete {args.tag}")
```
alessiodevoto: `gh release delete {tag}` leaves the tag behind ? This one:
 `gh release create <tag> --target <sha> --draft` creates the remote tag, aborting here leaves a remote version tag without a release.

 We can change to

  ```bash
  gh release delete {tag} --cleanup-tag
```

  or have the script offer to clean up  tag automatically?

File: scripts/make_release.py [hunk 2]
```diff
@@ -0,0 +1,965 @@
...
+    new_errors: list[str] = field(default_factory=list)
+    beyond_noise: list[str] = field(default_factory=list)
+    added: list[str] = field(default_factory=list)
+    removed: list[str] = field(default_factory=list)
+    floor_breach: str | None = None
+    markdown: str = ""
+
+    @property
+    def clean(self) -> bool:
+        return not (self.regressions or self.new_errors or self.beyond_noise or self.floor_breach)
```
alessiodevoto: Here `clean` ignores `added` / `removed` tests.  If a stable test fails on HEAD, the test is deleted or renamed, it lands in `removed`, and no regression is counted. Maybe we have to keep track of this in the diff ? Not sure

alessiodevoto:
Thanks this was definitely needed and it looks good overall! Left a couple of comments !

alessiodevoto (approved):
Thanks LGMT!

scripts/make_release.py
sklinglernv: Fixed. Hint now uses `--cleanup-tag`.

scripts/make_release.py
sklinglernv: Good catch. Fixed. `clean` now counts `removed`; `added` doesn't, since a new test has no baseline to regress from.

Issue #84, opened 2026-08-04
`import nooa` fails on native Windows: unconditional `import fcntl` in `nooa/storage/sqlite.py`


v0idAtlas:
## Environment
- Windows 11 Pro (native, no WSL), Python 3.13.14
- Confirmed at `nooa==0.0.8` (PyPI release) and at main SHA `3bb36bd` — `sqlite.py`'s import is identical in both, so the bug ships in the release, not just the repo head

## Repro
```python
import nooa
```

## Actual
```
ModuleNotFoundError: No module named 'fcntl'
```

`fcntl` is POSIX-only. The import chain is `nooa/__init__.py` → `nooa/storage/__init__.py` → `nooa/storage/sqlite.py`, whose top-level `import fcntl` makes the **entire package** unimportable on native Windows — not just the SQLite storage backend.

## Root cause
`src/nooa/storage/sqlite.py` imports `fcntl` unconditionally for the session lock, used in exactly two places:

- `_acquire_session_lock`: `fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)`
- `SQLiteStorageManager.close()`: `fcntl.flock(self._lock_fd, fcntl.LOCK_UN)`

## Suggested fix
Either guard the import with a platform check and back the lock with `msvcrt.locking` on Windows, or lazy-import `fcntl` inside the two lock functions so `import nooa` succeeds and only session locking is platform-conditional. A minimal msvcrt-backed equivalent of the two calls (tested — the storage round-trip suite passes with it on Windows 11):

```python
# acquire (exclusive, non-blocking):
os.lseek(fd, 0, os.SEEK_SET)
msvcrt.locking(fd, msvcrt.LK_NBLCK, 1)   # raises OSError if held, like LOCK_NB
# release:
os.lseek(fd, 0, os.SEEK_SET)
msvcrt.locking(fd, msvcrt.LK_UNLCK, 1)
```

Windows releases the byte-range lock at process termination, matching `flock`'s process-exit semantics.

## Workaround used locally
A shim module named `fcntl.py` (msvcrt-backed, implementing `flock` + the four `LOCK_*` constants) placed ahead of stdlib resolution. With it, `SQLiteStorageManager.save_snapshot` / `restore_latest_snapshot` verified working across a real process restart on Windows 11 — including a cross-process kill/restore round-trip.

Happy to open a PR with the msvcrt-backed approach if that's welcome.

[...]

PR #94, opened 2026-08-04
fix(tools): bound memory in ShellTools file operations 🤖🤖🤖

### LINKED ISSUE #90: Harden ShellTools file operations against oversized files
## Problem

`ShellTools` file operations can load an entire file into memory:

- `read(path, lines=...)` calls `read_text()` before selecting the requested line range.
- Both `replace(Match, ...)` and `replace(path, old, new)` read and reconstruct the complete file.

A coding agent can therefore exhaust process memory by inspecting or editing a very large, generated, or otherwise untrusted file. This is a core tool-safety concern, independent of TUI, ACP, or other host rendering.

The shared `ActivityShellTools` observer should not duplicate complete before/after file contents, but bounding observer payloads does not protect the underlying file operation itself.

## Desired behavior

- Ranged reads consume only the requested region with bounded memory.
- Unbounded whole-file reads have an explicit size budget and return an actionable error or require an intentional opt-in when exceeded.
- Replacement operations either:
  - reject files beyond a documented safe size, or
  - use a bounded/streaming implementation with safe atomic replacement.
- Existing `Match` line-number semantics and path-containment checks remain intact.
- Reuse the existing agentdoc `pformat` / `TruncatingStringIO` policy for previews and messages; do not add another truncation format.
- Limits and truncation are visible to callers rather than silently losing editable content.

## Tests

- A ranged read from a large file does not call an unbounded `read_text()` path.
- Whole-file reads over the limit fail with recovery guidance.
- Match-based and string-based replacements have deterministic behavior at the size boundary.
- Large-file failures leave the original file unchanged.
- Memory remains bounded for a large or sparse-file fixture.

Hotragn:
## What does this PR do?

Bounds the memory a `ShellTools` file operation can consume, so a file's size no longer decides how much memory the agent's process needs. Implements the desired behavior in #90.

### The four unbounded paths

| Site | Before |
|---|---|
| `read(path, lines=...)` | `read_text()` the whole file, *then* slice to the range |
| `replace(Match, new)` | whole-file read + reconstruct + `write_text()` |
| `replace(path, old, new)` | whole-file read for a `.count()` uniqueness check |
| `_harvest_matches()` | `read_text()` per matched path, to use exactly one line of it |

The fourth is not listed in the issue and is the widest exposure, because it is not reached through `read()` or `replace()` at all: any `run("grep -rn ...")` whose hits land in a large file pulled that file fully into memory during anchor harvesting. The cost was incurred without the agent ever requesting a read.

I also found that `write_text()` truncates in place, so this was never purely a memory concern — a failing large-file `replace` could leave a half-written file. That directly contradicts the "large-file failures leave the original file unchanged" test listed in the issue, so it is fixed here too.

### Changes

- **Ranged reads are genuinely ranged.** New `src/nooa/tools/_bounded_io.py` walks a file in 64 KiB chunks and stops at the requested end. Memory scales with the slice.
- **Anchor harvesting keeps only what it needs.** Cost scales with the number of matches instead of the size of the file, and reading stops after the highest anchored line.
- **Whole-file reads and both `replace()` forms are budgeted** by a new `ShellTools(max_file_bytes=...)`, 8 MiB by default, checked with `stat()` before opening so an oversized file is never allocated. Over budget raises with the size, the limit, and the recovery step (`lines=(start, end)`, or how to raise the budget).
- **Replacements are atomic.** Contents are swapped through a sibling temp file, fsynced, then `os.replace()`d — the standard write-temp-then-rename idiom (`os.replace` is atomic on POSIX and Windows). The original's permission bits are copied across first, because a naive atomic swap silently drops the executable bit off a script, which would matter to `BenchAgent` editing a repo.

### Two design calls worth flagging

**Streaming reads, budgeted replacements.** The issue offers either "reject files beyond a documented safe size" *or* "a bounded/streaming implementation" for replacements. I took streaming for ranged reads, where the requirement is explicit and the semantics are clean, and the budget for replacements. Streaming `replace(path, old, new)` means reimplementing `str.count` and `str.replace` over a sliding window with an overlap buffer, and matching their non-overlapping semantics exactly — real risk of subtle divergence for a path that has to rebuild the file anyway. Happy to do the streaming version if you'd rather.

**Line boundaries still follow `str.splitlines()`, not file iteration.** This is the trap in the change. Iterating a file object splits on `\n` alone; `str.splitlines()` also breaks on `\v`, `\f`, `\x1c`–`\x1e`, `\x85`, U+2028 and U+2029. Since the old code sliced `read_text().splitlines(keepends=True)`, a naive `for line in f` rewrite would have **silently renumbered every `Match`** in a file containing any of those characters. `_bounded_io` reproduces the `splitlines()` boundary set exactly, and decoding matches `Path.read_text()` (locale default encoding, universal newlines).

Reading no further than `end` is sound rather than approximate: the old code clamped with `min(total, end)`, so having successfully read `end` lines already proves `total >= end` and the clamp is a no-op. The remaining lines cannot change the answer, which is why the file does not need to be counted.

### Behavior changes

- A negative `end` previously reached Python's negative-index slicing and returned a *reversed* region (`all_lines[0:-3]`). It now yields an empty region. I don't think any caller can have meant the old result, but it is a change, and it is asserted in a test rather than left implicit.
- An over-budget file encountered during anchor harvesting yields `.matches is None`. That is the existing fail-closed contract ("no structured anchors" is a supported outcome), not a new error path.

No new truncation format was added, per your note — the limit messages are plain errors, and nothing here renders previews.

## Related issues

Closes #90

## Checklist

- [x] Docs updated if behavior or public APIs changed — `read` / `replace` / `write_file` docstrings document the budget and the atomic swap, so it reaches the model through `doc(self)`
- [x] New source files carry an SPDX license header (`scripts/check_license_headers.py` passes: 880 files)

### Tests

40 new tests in `tests/tools/test_shell_tools_bounded_io.py`, covering the issue's list:

- ranged read from a large file does not take an unbounded `read_text()` path — proved by replacing `Path.read_text` with something that raises, so the test cannot pass while reaching it
- whole-file reads over the limit fail with recovery guidance in the message
- Match-based and string-based replacements are deterministic at the size boundary (exact-size passes, one byte over raises)
- large-file failures leave the original unchanged, with no temp files left behind
- memory stays bounded on a >15 MiB fixture, measured with `tracemalloc` while reading the file's *last* two lines, which forces the whole file to be streamed
- plus: differential equivalence against the old implementation across a range matrix and one case per exotic break character, chunk-boundary invariance down to `chunk_chars=1`, executable-bit preservation, symlink write-through, and `_resolve_path` containment still rejecting `../`

**These tests were verified to fail against the current implementation.** Reverting only `shell_tools.py` and re-running gives 13 failed / 27 passed — the 13 fail on the unbounded reads, the missing budget, and the non-atomic writes; the 27 that still pass are the differential ones, which assert *unchanged* behavior and so should pass on both.

Local runs on Linux (`uv sync --all-extras --no-extra sandbox`, matching CI): `ruff check .`, `ruff format --check .`, `scripts/check_license_headers.py`, and the unit suite. `pyright` on the touched files reports only a pre-existing error at `shell_tools.py:446` (`kind` `Literal`), which is present on `main` unchanged.

### Open questions

1. **Default budget.** 8 MiB is a bound, not a claim about the largest useful file — generous for source, still bounded, cheap to override. Say the word if you want a different house number.
2. **`FileTooLargeError`** lives in the private `_bounded_io` and subclasses `ValueError`, so existing `except ValueError` guards (including `_harvest_matches`) keep working and nothing new lands on the public surface. If you'd rather it be catchable by name, tell me where you want it exported.
3. Happy to split this into two commits/PRs — read + harvest bounds, then replace atomicity — if that reviews more easily.

🤖🤖🤖

## Summary by CodeRabbit

* **New Features**
  * Added configurable file-size limits for shell-tool file operations, with an 8 MiB default.
  * Added bounded-memory support for reading complete files, line ranges, and selected lines.
  * Large files and oversized individual lines now produce clear file-size errors before excessive memory is used.
  * File replacement and search-anchor processing now respect the configured limits.
* **Tests**
  * Added coverage for size limits, line handling, memory usage, range reads, replacements, and error behavior.

### DIFF
--- a/src/nooa/tools/_bounded_io.py
+++ b/src/nooa/tools/_bounded_io.py
@@ -1,0 +1,248 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Bounded-memory file primitives backing ``ShellTools`` file operations.
+
+A coding agent points its file tools at whatever the repository contains,
+including generated, minified, or otherwise untrusted files. Reading such a file
+whole — to hand back five lines of it, or to check that a search string occurs
+exactly once — lets the file's size, rather than the request's size, decide how
+much memory the agent's process needs.
+
+The primitives here bound that cost:
+
+- :func:`iter_lines_keepends` walks a file in fixed-size chunks, holding at most
+  one chunk plus one line.
+- :func:`read_line_range` reads only as far as the requested range.
+- :func:`read_specific_lines` keeps only the lines actually asked for, so cost
+  scales with the number of matches rather than the size of the file.
+- :func:`read_whole_file_checked` refuses an unbounded read above a byte budget
+  instead of discovering the size after allocating it.
+- :func:`atomic_replace_text` swaps contents through a sibling temp file, so a
+  failed write leaves the original intact.
+
+**Line boundaries follow** ``str.splitlines()``, not file iteration. Iterating a
+file object splits on ``\\n`` alone, while ``str.splitlines()`` also breaks on
+``\\v``, ``\\f``, ``\\x1c``-``\\x1e``, ``\\x85``, ``\\u2028`` and ``\\u2029``.
+The previous implementation sliced ``read_text().splitlines(keepends=True)``, so
+adopting file iteration would silently renumber every :class:`Match` in a file
+containing any of those characters. These helpers reproduce the
+``splitlines()`` boundary set exactly.
+
+Text decoding also matches the previous code path: files are opened with the
+locale default encoding and universal-newline translation, the same as
+``Path.read_text()``.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import os
+import tempfile
+from collections.abc import Iterator
+from pathlib import Path
+
+#: Default ceiling for a single unbounded file read or replacement.
+#:
+#: Source files are overwhelmingly under a megabyte; the headroom here is for
+#: checked-in generated output (bundles, lockfiles, fixtures) that an agent may
+#: legitimately need to edit. It is deliberately a bound rather than a guess at
+#: the largest useful file — callers that genuinely need more pass
+#: ``ShellTools(max_file_bytes=...)`` and make that choice explicit.
+DEFAULT_MAX_FILE_BYTES = 8 * 1024 * 1024
+
+#: Characters read per chunk while walking a file. Bounds peak memory together
+#: with the longest line encountered.
+CHUNK_CHARS = 64 * 1024
+
+
+class FileTooLargeError(ValueError):
+    """A file operation exceeded its byte budget.
+
+    Subclasses :class:`ValueError` on purpose: ``ShellTools`` already raises
+    ``ValueError`` for unusable file arguments, and ``_harvest_matches`` guards
+    its reads with ``except (OSError, ValueError)``. Inheriting keeps both
+    call sites behaving as they did without a new exception type to catch.
+    """
+
+    def __init__(self, path: Path | str, size: int, limit: int, hint: str, summary: str) -> None:
+        self.path = str(path)
+        self.size = size
+        self.limit = limit
+        super().__init__(f"{self.path} {summary}. {hint}")
+
+
+def iter_lines_keepends(
+    path: Path,
+    *,
+    max_line_chars: int,
+    chunk_chars: int = CHUNK_CHARS,
+) -> Iterator[str]:
+    """Yield lines exactly as ``read_text().splitlines(keepends=True)`` would.
+
+    Peak memory is one chunk plus the longest line, instead of the whole file.
+
+    Args:
+        path: File to read.
+        max_line_chars: Ceiling on a single line. A file with no line break in
+            this many characters would otherwise defeat chunking, so it raises
+            :class:`FileTooLargeError` rather than growing without limit.
+        chunk_chars: Characters per read.
+
+    Raises:
+        FileTooLargeError: A single line exceeded ``max_line_chars``.
+        OSError: The file could not be opened or read.
+    """
+    carry = ""
+    with open(path) as handle:
+        while True:
+            chunk = handle.read(chunk_chars)
+            if not chunk:
+                break
+            parts = (carry + chunk).splitlines(keepends=True)
+            # The trailing element may be an incomplete line, so it is held back
+            # until the next chunk confirms where it ends.
+            carry = parts.pop() if parts else ""
+            if len(carry) > max_line_chars:
+                raise FileTooLargeError(
+                    path,
+                    len(carry),
+                    max_line_chars,
+                    "A single line cannot be read in bounded chunks. Raise the budget "
+                    "with ShellTools(max_file_bytes=...) if this file is safe to load "
+                    "whole.",
+                    f"has a line over {max_line_chars:,} characters",
+                )
+            yield from parts
+    if carry:
+        yield carry
+
+
+def read_line_range(
+    path: Path,
+    start: int,
+    end: int,
+    *,
+    max_line_chars: int,
+) -> tuple[str, int]:
+    """Read lines ``[start, end]`` (1-indexed, inclusive) without reading past ``end``.
+
+    Mirrors the slice it replaces —
+    ``read_text().splitlines(keepends=True)[start - 1 : end]`` — including its
+    clamping: ``end`` is reported unchanged when the file has at least that many
+    lines, and lowered to the real line count only when the file ends first.
+    That equivalence is what makes reading no further than ``end`` sound: having
+    read ``end`` lines successfully already proves ``min(total, end) == end``.
+
+    Returns:
+        ``(text, clamped_end)`` — the joined region and the line number to
+        report as :attr:`Match.end`.
+    """
+    if end < 1:
+        # A non-positive end selects nothing. The old code reached Python's
+        # negative-index slicing here and returned a reversed region for
+        # end < 0, which no caller can have meant; an empty region is reported
+        # instead.
+        return "", end
+
+    pieces: list[str] = []
+    seen = 0
+    for seen, line in enumerate(  # noqa: B007 - final value is the line count
+        iter_lines_keepends(path, max_line_chars=max_line_chars), 1
+    ):
+        if seen >= start:
+            pieces.append(line)
+        if seen >= end:
+            break
+    return "".join(pieces), min(end, seen)
+
+
+def read_specific_lines(
+    path: Path,
+    wanted: set[int],
+    *,
+    max_line_chars: int,
+) -> dict[int, str]:
+    """Return ``{line_number: text}`` for ``wanted``, reading no further than needed.
+
+    Used for search-anchor harvesting, where a handful of line numbers are
+    needed out of a possibly enormous file. Memory scales with ``len(wanted)``
+    rather than the file's size, and reading stops after the highest requested
+    line.
+
+    Line numbers past the end of the file are simply absent from the result,
+    which lets the caller keep its fail-closed "no trustworthy anchors" path.
+    """
+    if not wanted:
+        return {}
+    highest = max(wanted)
+    if highest < 1:
+        return {}
+
+    found: dict[int, str] = {}
+    for seen, line in enumerate(iter_lines_keepends(path, max_line_chars=max_line_chars), 1):
+        if seen in wanted:
+            found[seen] = line
+        if seen >= highest:
+            break
+    return found
+
+
+def read_whole_file_checked(path: Path, max_bytes: int, *, hint: str) -> str:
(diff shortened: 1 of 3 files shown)


Hotragn:
Flagging four consequences of the atomic swap that I should have put in the description. Three are minor; the first is a real semantic change and I'd rather surface it than have it found in review.

**1. Hardlinked files diverge.** `write_text()` truncates in place, so every hardlink to an inode saw an edit. `os.replace()` swaps the directory entry, so the edited path gets a new inode and other links keep the old content. Verified:

```
before -> a: needle | b: needle    same inode: True
after  -> a: thread | b: needle    same inode: False
```

Options, in the order I'd pick them:

- **Document it.** Hardlinked source files are rare, and for a coding agent editing a checkout the atomic guarantee is worth more than hardlink write-through. This is also what `git`, `dpkg` and most editors do.
- **Detect and fall back.** `if path.stat().st_nlink > 1: write in place`. Preserves the old behavior exactly where it matters, at the cost of losing atomicity for precisely those files — which inverts the guarantee in the confusing direction.
- **Copy the inode back.** Write the temp file, then `os.replace` only if `st_nlink == 1`, else open the original `r+` and rewrite. Same tradeoff as above, more code.

I went with documenting it, and it is not currently asserted in a test. Happy to add either the fallback or an explicit test pinning the divergence, whichever you prefer — I'd lean toward a test that documents the choice.

**2. Replacement now needs a writable parent directory,** not just a writable file, since the temp file is created as a sibling. A read-only directory containing a writable file used to be editable and no longer is. Inherent to atomic replacement; the alternative is writing the temp file elsewhere, which breaks the same-filesystem requirement that makes `os.replace` atomic.

**3. `os.fsync()` on every replacement** adds a real per-edit cost. That is the deliberate price of "a failed edit leaves the original intact" — without the flush the swap is atomic with respect to the directory entry but not durable across a crash. If the latency matters more than crash durability for `BenchAgent` throughput, dropping the `fsync` still gets you the no-partial-file property; say so and I'll make it conditional.

**4. Mode preservation is effectively a no-op on Windows,** where `os.chmod` only honors the read-only bit. Not a regression — `write_text()` preserved nothing either — but the executable-bit test is meaningful only on POSIX. It currently runs everywhere; I can gate it if you add a `posix` marker (see #95).

Separately, and unrelated to the above: I verified on native Windows that the `ShellTools` *file* surface works there, including these changes — `read(path, lines=...)` and `write_file()` both round-trip correctly. Only the shell-session surface is POSIX-bound (`pass_fds` at `_bash_session.py:173`). So nothing here makes the Windows situation worse, for whenever #84/#85 land.

🤖🤖🤖

Hotragn:
Rebased onto current `main` — this had gone conflicting when #172 landed. All checks are green on `e8730d8` and it's mergeable again.

#172 removed the cwd containment check from `_resolve_path`, and it touches the same functions this PR does, so the resolution was more than textual. What the two changes each own, and how they compose:

- **`_extract_matches` cache.** #172 widened `file_cache` to carry the resolved `Path` so a `Match` stays bound to its file across a `cd`; this PR narrowed it to hold only anchored lines instead of whole files. Merged as `dict[str, tuple[Path, dict[int, str]]]` — both properties, neither dropped.
- **`Match` construction.** `resolved_path` is now a required keyword, so all four construction sites here pass it. `_resolve_path` returns an absolute path, which satisfies the new absolute-path check.
- **`replace()`, Match form.** Takes #172's `Path(target.resolved_path)` anchor rather than re-resolving `target.path`, with this PR's `read_whole_file_checked` budget applied to it.
- **`read()`.** Ranged and whole-file reads keep their bounds and now carry `resolved_path` through.

#172's `new_text=` on all three `FileWrite` returns, its `timed_out` plumbing, and the containment removal itself are all preserved as-is.

**One test changed meaning, so calling it out rather than leaving it in the diff.** `test_replace_still_rejects_paths_outside_cwd` asserted the containment #172 deliberately removed, so it could not survive as written. Rather than delete it I rewrote it as `test_replace_outside_cwd_is_still_bounded`: the parent-relative path now resolves instead of raising, and the assertion is that the size budget still applies to a file reached that way. If you'd rather it just be dropped, say so and I'll drop it — but it seemed worth keeping the half of the coverage that #172 doesn't invalidate.

Verification on `e8730d8`:

- `uv run pytest -q tests/tools` — 339 passed
- `uv run ruff check` / `ruff format --check` on `shell_tools.py`, `_bounded_io.py`, `test_shell_tools_bounded_io.py` — clean
- `git merge-tree --write-tree HEAD upstream/main` — single tree, no conflict

**On size, since +830 is a lot to ask anyone to review:** happy to split this into two PRs — part one the read bounds and the budget on both `replace` forms, part two `atomic_replace_text` and its tests, which is where the hardlink question in my earlier comment actually lives. That would let the bounds land without the atomic-swap semantics discussion blocking them. Say the word and I'll open it that way instead.

Note for whoever looks at this alongside #118: the two merged cleanly before today, and now conflict only because this branch sits on post-#172 `main` and that one doesn't yet. The underlying diff regions still don't overlap — #118's `shell_tools.py` changes are in `run_stream`, these are in the file-operation surface — so I'd expect it to come back clean once #118 rebases. I haven't run a combined test suite, so that's a diff-region read, not a verified merge.

🤖🤖🤖

furgalep:
Thanks for putting this together — the streaming/ranged-read approach is thoughtful, and I especially appreciate the care taken to preserve the existing `splitlines()` behavior and to document the hardlink tradeoff.

While reviewing it, I reproduced a few remaining contract gaps and added focused acceptance tests on [`review/pr94-regression-tests`](https://github.com/NVIDIA-NeMo/labs-OO-Agents/tree/review/pr94-regression-tests), commit [`5b88fa5`](https://github.com/NVIDIA-NeMo/labs-OO-Agents/commit/5b88fa5f00b61eca0828b7153e5d261314a4dd98).

The existing 41 PR tests still pass. The five added regressions currently fail independently for these reasons:

1. stale pathname metadata can permit a whole-file read beyond `max_file_bytes`;
2. atomic creation produces mode `0600` instead of normal umask-derived permissions;
3. replacement loses supported extended attributes;
4. the parent directory is not fsynced despite the crash-durability wording;
5. zero and negative `max_file_bytes` values are accepted.

The next step would be to cherry-pick `5b88fa5`, address those failures, and share the updated test results. The overall bounded-I/O design looks good; these tests are meant to make the remaining guarantees precise and help us get it over the line. Thanks again for the careful work here.

Hotragn:
Rebased onto current `main` (`d4d46f7`) and **cut down to the reads-only half**, which is the split I offered earlier in this thread. This is now one commit.

**What's still here:** the bounds. Three paths read a whole file regardless of what was asked for, and all three are fixed:

- `read(path, lines=...)` read the entire file to return a range; a ranged read now consumes only that range.
- `replace()`, both forms, read the whole file to rebuild it; both are now capped by `max_file_bytes` (8 MiB default, per-instance via `ShellTools(max_file_bytes=...)`) and raise `FileTooLargeError` with an actionable hint instead of loading it.
- Anchor harvesting in `_extract_matches` read every matched file whole to quote one line from each. This is the one I'd most want reviewed: the agent never requested a read, so any `grep` over a large file triggered it. Cost now scales with the number of anchors.

**What I removed:** `atomic_replace_text` and the five tests that covered it. Writes still go through `Path.write_text`, exactly as on `main`. My earlier comment in this thread listing four consequences of the atomic swap — hardlink divergence, the writable-parent requirement, per-edit `fsync` cost, and Windows mode preservation — applies to that deferred change, not to anything in the current diff. Nothing in this PR changes write semantics, so none of those questions need answering to review it. I'll open the atomic replacement separately once #90 is closed, and the hardlink choice can be settled there.

The size is now +676/-20 across 3 files: the test file is +392, `_bounded_io.py` is +200, and `shell_tools.py` is +84/-20.

**On line boundaries, which is the load-bearing detail.** The old code sliced `read_text().splitlines(keepends=True)`. `str.splitlines()` breaks on more than newlines — vertical tab, form feed, the file/group/record separators U+001C through U+001E, NEL (U+0085), and the Unicode line and paragraph separators U+2028 and U+2029 — where iterating a file object does not. A naive streaming rewrite would silently renumber every `Match` in a file containing any of those. So `_bounded_io` reproduces the `splitlines()` boundary set exactly, and the tests assert it differentially against the old implementation, which is kept verbatim in the test file as `_legacy_read`. Text decoding matches `Path.read_text()` too. The proof-of-absence tests replace `Path.read_text` with something that raises, so a passing ranged-read test cannot be reaching the unbounded path.

Verification on `6cd9d80`:

- `uv run pytest -q tests/tools` — 352 passed
- `uv run ruff check` / `ruff format --check` on the three files — clean
- `git merge-tree --write-tree HEAD upstream/main` — single tree, no conflict

Note it also picks up the two behaviours that landed while this sat: #172's removal of cwd containment (all four `Match` construction sites carry `resolved_path`, and `replace()` uses the anchor's own resolved path rather than re-resolving), and #308's EOF newline handling, which is untouched.

🤖🤖🤖

Issue #85, opened 2026-08-04
`install_debug_handler()` crashes at import on Windows: `signal.SIGUSR2` AttributeError escapes the platform guard


v0idAtlas:
## Environment
- Windows 11 Pro (native), Python 3.13.14
- Confirmed at `nooa==0.0.8` (PyPI release); code path unchanged from main SHA `3bb36bd`
- Only reachable once `import nooa` succeeds on Windows — see the companion fcntl issue: #84

## Repro
```python
import nooa   # with a local fcntl shim in place; see companion issue
```

## Actual
```
File "...\nooa\__init__.py", line 180, in <module>
    install_debug_handler()
File "...\nooa\runtime\debug_handler.py", line 429, in install_debug_handler
    signal.signal(signal.SIGUSR2, _debug_signal_handler)
                  ^^^^^^^^^^^^^^
AttributeError: module 'signal' has no attribute 'SIGUSR2'
```

## Root cause
`install_debug_handler` already anticipates platform problems — the registration is wrapped in:

```python
try:
    signal.signal(signal.SIGUSR2, _debug_signal_handler)
    _handler_installed = True
except (ValueError, OSError):
    # Can't set signal handler (not main thread, or platform issue)
    pass
```

but on Windows the failure happens at the **attribute access** (`signal.SIGUSR2` does not exist on win32), raising `AttributeError` — which the except clause doesn't cover. The guard catches the right idea one expression too late.

## Suggested fix (either is one line)
```python
except (ValueError, OSError, AttributeError):
```
or guard the attribute:
```python
if hasattr(signal, "SIGUSR2"):
    signal.signal(signal.SIGUSR2, _debug_signal_handler)
    _handler_installed = True
```

## Workaround used locally
Before importing nooa: `signal.SIGUSR2 = 0` — an invalid signal number, which routes the failure into the existing `ValueError` arm and no-ops cleanly. Verified: package imports and the storage suite runs on Windows.

Happy to fold this into the same PR as the fcntl fix if that's easier to review.

PR #98, opened 2026-08-05
fix(windows): resolve POSIX-only compatibility issues (#92, #95)

### LINKED ISSUE #92: [Bug]  Windows: two unconditional POSIX-only accesses crash the sandbox on construction and on worker death (guards.py + executor.py)
**Describe the bug**

Two related POSIX-only accesses in the sandbox module crash on Windows, both in the same class as [#84](https://github.com/NVIDIA-NeMo/labs-OO-Agents/issues/84) and [#85](https://github.com/NVIDIA-NeMo/labs-OO-Agents/issues/85).

**Bug A — `import resource` in `guards.py` (crashes on construction)**

`SandboxedExecutor` cannot be constructed on Windows at all. The moment `__init__` runs, it calls `_capabilities()` → `probe_capabilities()` in `src/nooa/runtime/sandbox/guards.py` (line 179), whose first statement is an unconditional `import resource` (line 181). The `resource` module is POSIX-only and does not exist on Windows, so this raises `ModuleNotFoundError` before any sandbox work begins.

A second unconditional `import resource` exists in `apply_rlimits()` (line 215 of the same file), currently dormant because its caller zeroes the arguments on non-Linux hosts before the import is reached — but it will crash if called directly or if the pruning logic changes.

**Bug B — `signal.SIGXCPU` / `signal.SIGKILL` in `executor.py` (crashes on worker death)**

`_classify_worker_death()` in `src/nooa/runtime/sandbox/executor.py` (line 409) checks the worker exit code against `signal.SIGXCPU` (line 412) and `signal.SIGKILL` (line 414) without any platform guard. Both constants are POSIX-only and do not exist in the `signal` module on Windows — accessing them raises `AttributeError` immediately, before the `==` comparison even runs. The result: whenever a sandboxed worker process dies on Windows, the classifier itself crashes with an unrelated `AttributeError`, masking the real failure entirely.

Bug B is currently unreachable because Bug A crashes first — but once Bug A is fixed, Bug B becomes immediately reachable on the `require=False` path.

I'm working on a PR covering both and will submit it shortly.

**To reproduce**

_Bug A_ — on any Windows host:

```python

# Crashes immediately, never gets past __init__
executor = SandboxedExecutor(
    agent=None,
    config=SandboxConfig(require=False),
    cell_timeout=None,
)
```

Actual output:

```
Traceback (most recent call last):
  File "...", line X, in <module>
    executor = SandboxedExecutor(...)
  File ".../executor.py", line 83, in __init__
    caps = _capabilities()
  File ".../executor.py", line 45, in _capabilities
    _CAPS_CACHE = probe_capabilities()
  File ".../guards.py", line 181, in probe_capabilities
    import resource
ModuleNotFoundError: No module named 'resource'
```

_Bug B_ — after Bug A is fixed, on a Windows host:

```python
from nooa.runtime.sandbox.errors import WorkerDiedError
import multiprocessing as mp

executor = SandboxedExecutor(
    agent=None,
    config=SandboxConfig(require=False),
    cell_timeout=None,
)

mock_proc = mp.Process(target=lambda: None)
mock_proc.start()
mock_proc.join()
executor._proc = mock_proc
executor._classify_worker_death(WorkerDiedError("worker died"))
```

Actual output:

```
Traceback (most recent call last):
  File "...", line X, in _classify_worker_death
    if code == -signal.SIGXCPU:
AttributeError: module 'signal' has no attribute 'SIGXCPU'
```

**Expected behavior**

- `probe_capabilities()` should detect that `resource` is unavailable on Windows — just as it already detects that `landlock` and `seccomp` are unavailable — and return a `Capabilities` object with `rlimit=False` rather than crashing. The sandbox should then either proceed in degraded mode (`require=False`) or raise a clean `SandboxUnavailable` (`require=True`), not an unhandled `ModuleNotFoundError`.

- `_classify_worker_death()` should check whether the signal constants exist before using them. On Windows, both `signal.SIGXCPU` and `signal.SIGKILL` are absent, so the guarded checks should be skipped and the function should fall through to `return exc`, surfacing the original `WorkerDiedError` cleanly.

**Environment**

- OS: Windows 11 Pro (native)
- Python version: 3.13
- NOOA version: latest published tag is v0.0.8 — note: `python -c "import nooa; print(nooa.__version__)"` cannot be run because the import chain crashes with Bug A before `__version__` is reachable. Tested against `main` at commit `2197e61ddbedb398a9926944b25329261ada08c7`, which may be a few commits ahead of the v0.0.8 tag.
- Model / provider: N/A — platform-level bugs in the sandbox module, unrelated to any specific model or provider

**Additional context**

_Bug A_ crash call chain (line numbers from commit `2197e61ddbedb398a9926944b25329261ada08c7`):

```
SandboxedExecutor.__init__()    executor.py  line  83
  → _capabilities()             executor.py  line  45
    → probe_capabilities()      guards.py    line 179
      → import resource         guards.py    line 181  ← CRASH
```

Second affected site in `guards.py` (dormant today, fragile):

```
apply_rlimits()                 guards.py    line 207
  → import resource             guards.py    line 215
```

_Bug B_ crash call chain:

```
run_cell()                   executor.py  line 245  (send-side worker death)
run_cell()                   executor.py  line 256  (recv-side worker death)
  → _classify_worker_death() executor.py  line 409
    → -signal.SIGXCPU        executor.py  line 412  ← CRASH
```

Confirmed live on Windows Python 3.13:

```
>>> import resource
ModuleNotFoundError: No module named 'resource'

>>> import signal
>>> hasattr(signal, 'SIGXCPU'), hasattr(signal, 'SIGKILL')
(False, False)
```

Suggested fixes:

- Bug A: gate the `import resource` in `probe_capabilities()` behind the existing `is_linux` check, returning `rlimit=False` when the import isn't available. Apply the same pattern to `apply_rlimits()`.
- Bug B: guard both signal checks with `hasattr`, matching the convention already used elsewhere in this codebase for the same class of problem.

There is no existing test coverage for `_classify_worker_death` — nothing in `tests/` calls it or exercises `WorkerDiedError` in a death-classification context. The PR should include a regression test that mocks a dead worker process with a Windows-style exit code.

### LINKED ISSUE #95: CI runs only on ubuntu-latest, so POSIX-only regressions reach PyPI before anyone notices 🤖🤖🤖
## The gap

Every job in `.github/workflows/ci.yml` and `publish.yml` is `runs-on: ubuntu-latest`. Nothing verifies that `import nooa` works on any other platform, so a module-level POSIX-only import reaches PyPI without anything failing first.

That is not hypothetical — it already happened, three times, and each time a user found it rather than CI:

- #84 — unconditional `import fcntl` in `nooa/storage/sqlite.py` makes the **entire package** unimportable on native Windows. Confirmed shipped in `nooa==0.0.8`, not just on `main`.
- #85 — `signal.SIGUSR2` is read outside its platform guard in `install_debug_handler()`, which `__init__.py` calls at import time.
- #92 — unconditional `import resource` in `sandbox/guards.py`, plus `signal.SIGXCPU`/`SIGKILL` in `executor.py`.

Two independent reporters, three issues, all in the same class. **I am not proposing to take any of them** — #92's reporter says they have a PR in progress, and #84/#85's reporter offered to submit both. This issue is about the missing guard that lets the class recur.

## What I measured

I ran a ladder on native Windows 11 (Python 3.13, no WSL), applying the minimum shim for each blocker to see what the next one is:

```
STEP 1  bare `import nooa`                  ModuleNotFoundError: fcntl                    <- #84
STEP 2  + msvcrt-backed fcntl shim          AttributeError: signal has no SIGUSR2         <- #85
STEP 3  + signal.SIGUSR2 = 0                imported OK, version 0.0.1.dev162
STEP 4  SandboxedExecutor(require=False)    ModuleNotFoundError: resource                 <- #92
STEP 5  ShellTools.read()/write_file()      OK
STEP 6  ShellTools.run()                    AssertionError: pass_fds not supported on Windows
STEP 7  eval_pipeline._memory_monitor       ModuleNotFoundError: resource
```

Three things in there are worth separating out from the known issues:

**STEP 3 is the important one.** Once #84 and #85 are fixed, `import nooa` *succeeds* on native Windows today. The two blockers are both one-line platform guards, and nothing deeper is preventing importability. That makes an import smoke test a genuinely cheap gate rather than the first step of a long port.

**STEP 6 and 7 are not reported in any open issue.**
- `src/nooa/tools/_bash_session.py:173` passes `pass_fds=(ctrl_w,)` to `create_subprocess_exec`, which asserts on Windows. Note this is a *different* failure from the `os.killpg`/`os.getpgid` calls at lines 79-80 and 642-661 — those are downstream and never reached, so an audit that only greps for `killpg` misses the real blocker.
- `util/eval_pipeline/src/eval_pipeline/_memory_monitor.py:19` imports `resource` unconditionally.

**STEP 5 is a useful negative result:** the `ShellTools` *file* surface (`read`, `replace`, `write_file`) is already platform-clean. Only the shell-session surface is POSIX-bound.

## Why a gate, and not just the fixes

The repo already has the right convention — `hasattr(signal, "SIGUSR1")` guards in `packages/nooa-cli/src/nooa_cli/commands/eval.py:47` and `util/eval_pipeline/src/eval_pipeline/__main__.py:13`. It is applied in two places and missed in four. That is exactly the pattern that a per-PR check fixes and code review does not: nobody reviewing a diff on `sqlite.py` is thinking about win32.

There's also a contributor cost, which is how I ran into this. On a Windows checkout the test suite cannot start at all — `conftest.py` imports `nooa.storage.sqlite` — so `uv run pytest` fails at collection before a single test runs. Any Windows-based contributor is blocked on their first command.

## Proposed design

Deliberately narrow, because full Windows support is a much bigger project and I don't think it's what's being asked for here. Running the whole suite on `windows-latest` is not viable: `BashSession` needs `bash` and `pass_fds`, and the Landlock/seccomp/RLIMIT sandbox is Linux-only by design.

**Tier 1 — import smoke test (this issue).** A `windows-latest` job that installs the built wheels and asserts each distribution imports:

```yaml
  import-smoke:
    strategy:
      matrix:
        os: [windows-latest, macos-latest]
    runs-on: ${{ matrix.os }}
    steps:
      - uses: actions/checkout@v4
      - uses: astral-sh/setup-uv@v5
        with:
          python-version: "3.12"
      - run: uv run --no-project --with . python -c "import nooa; print(nooa.__version__)"
```

Seconds of runtime, no credentials, and it fails on exactly the bug class that reached `0.0.8`. I'd suggest including `macos-latest` in the same matrix: macOS is POSIX so it wouldn't have caught #84, but it *would* catch a Linux-only assumption (`/proc`, `landlock`), and it's the platform most likely to be used by contributors after Linux.

**Tier 2 — a `posix` marker, later and incrementally.** `pyproject.toml` already carries `integration`, `stress`, and `sandbox` markers, so `posix` fits the existing convention. POSIX-bound tests get marked, and the Windows job grows to run the unmarked remainder. This should follow the fixes rather than lead them, and is a separate PR.

## Sequencing

The Tier 1 job goes **red until #84 and #85 land** — that's the point of it, but it also means it can't merge before those do. Three options, and I'd like a maintainer call rather than guessing:

1. Land the job after #84/#85 merge. Cleanest, but depends on other contributors' PRs.
2. Land it now with `continue-on-error: true`, then flip it once the fixes are in. Visible immediately, no red required check.
3. Fold the two one-line guards into the same PR as the job. Fastest to green, but duplicates work #84/#85's reporter already offered to do — which is why I'm not proposing it.

I'd suggest (1), with (2) as a fallback if #84/#85 stall.

## What I'm offering

Happy to submit Tier 1 — the workflow job, plus a short "Supported platforms" note in `CONTRIBUTING.md` recording that the suite currently requires a POSIX host, so the next Windows contributor isn't surprised by a collection error. I'll also file the two unreported findings (STEP 6, STEP 7) as their own issue if you'd rather track them separately from this one; say the word and I'll split them out.

Not filing this as a PR first because CONTRIBUTING asks for an issue on anything substantial, and adding runners to CI is a maintainer cost decision rather than a code question.

🤖🤖🤖

Harshitmishra001:
## What does this PR do?

Resolves unconditional POSIX imports and asserts across the codebase that crash the framework on Windows.

- Add platform guards for sandbox resource limits and executor death classification

- Handle missing fcntl and fallback to SQLite native locking

- Provide graceful fallbacks for resource memory limits in eval pipeline

- Prevent asyncio subprocess failure on missing start_new_session/pass_fds args

- Add regression tests for sandbox guards and executor

## Related issues

Fixes #92 
Fixes #95

## Checklist

### DIFF
--- a/src/nooa/agentdoc/_truncating_stream.py
+++ b/src/nooa/agentdoc/_truncating_stream.py
@@ -181,7 +181,7 @@ class FileBackedTruncatingStringIO(TruncatingStringIO):
         self._file: io.TextIOWrapper | None = None
         try:
             fd, path = tempfile.mkstemp(dir=dir, prefix=prefix, suffix=suffix)
-            self._file = os.fdopen(fd, "w")
+            self._file = os.fdopen(fd, "w", encoding="utf-8")
             self._file_path = path
         except OSError:
             _log.warning(
--- a/src/nooa/runtime/debug_handler.py
+++ b/src/nooa/runtime/debug_handler.py
@@ -426,8 +426,9 @@ def install_debug_handler(dump_dir: Path | None = None) -> None:
 
     # Install SIGUSR2 handler (less commonly used than SIGUSR1)
     try:
-        signal.signal(signal.SIGUSR2, _debug_signal_handler)
-        _handler_installed = True
+        if hasattr(signal, "SIGUSR2"):
+            signal.signal(signal.SIGUSR2, _debug_signal_handler)
+            _handler_installed = True
     except (ValueError, OSError):
         # Can't set signal handler (not main thread, or platform issue)
         pass
--- a/src/nooa/runtime/sandbox/executor.py
+++ b/src/nooa/runtime/sandbox/executor.py
@@ -409,9 +409,9 @@ class SandboxedExecutor:
     def _classify_worker_death(self, exc: WorkerDiedError) -> Exception:
         proc = self._proc
         code = getattr(proc, "exitcode", None)
-        if code == -signal.SIGXCPU:
+        if hasattr(signal, "SIGXCPU") and code == -signal.SIGXCPU:
             return CellTimeoutError("cell exceeded its CPU-time limit and was killed")
-        if code == -signal.SIGKILL:
+        if hasattr(signal, "SIGKILL") and code == -signal.SIGKILL:
             return CellMemoryError(
                 "worker was killed (out-of-memory or resource limit). "
                 "Reduce the cell's memory use or raise max_memory_mb."
--- a/src/nooa/runtime/sandbox/guards.py
+++ b/src/nooa/runtime/sandbox/guards.py
@@ -178,14 +178,20 @@ class Capabilities:
 
 def probe_capabilities() -> Capabilities:
     """Probe the host once for every enforcement mechanism the sandbox uses."""
-    import resource
-
     is_linux = platform.system() == "Linux"
+    rlimit = False
+    if is_linux:
+        try:
+            import resource
+
+            rlimit = hasattr(resource, "RLIMIT_AS") and hasattr(resource, "RLIMIT_CPU")
+        except ImportError:
+            pass
     return Capabilities(
         linux=is_linux,
         landlock_abi=landlock_abi() if is_linux else 0,
         seccomp=seccomp_supported() if is_linux else False,
-        rlimit=hasattr(resource, "RLIMIT_AS") and hasattr(resource, "RLIMIT_CPU"),
+        rlimit=rlimit,
     )
 
 
@@ -212,7 +218,10 @@ def apply_rlimits(*, max_memory_mb: int = 0, max_cpu_seconds: int = 0) -> None:
     so an absolute cap below that baseline would break it instantly. The cap is
     ``current VmSize + max_memory_mb`` — i.e. how much more the cell may allocate.
     """
-    import resource
+    try:
+        import resource
+    except ImportError:
+        return  # resource module is not available on this platform (e.g. Windows)
 
     if max_memory_mb and max_memory_mb > 0:
         baseline = _self_vmsize_bytes()
--- a/src/nooa/storage/sqlite.py
+++ b/src/nooa/storage/sqlite.py
@@ -7,7 +7,10 @@ Provides persistent storage using stdlib sqlite3 — no new dependencies.
 
 from __future__ import annotations
 
-import fcntl
+try:
+    import fcntl
+except ImportError:
+    fcntl = None
 import json
 import logging
 import os
@@ -643,7 +646,8 @@ def _acquire_session_lock(lock_path: str) -> int:
     """
     fd = os.open(lock_path, os.O_CREAT | os.O_RDWR, 0o644)
     try:
-        fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        if fcntl is not None:
+            fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
     except OSError:
         os.close(fd)
         owner_pid = _read_lock_pid(lock_path)
@@ -832,7 +836,8 @@ class SQLiteStorageManager:
                     conn.close()
         finally:
             if self._lock_fd is not None:
-                fcntl.flock(self._lock_fd, fcntl.LOCK_UN)
+                if fcntl is not None:
+                    fcntl.flock(self._lock_fd, fcntl.LOCK_UN)
                 os.close(self._lock_fd)
                 self._lock_fd = None
 
--- a/src/nooa/tools/_bash_session.py
+++ b/src/nooa/tools/_bash_session.py
@@ -18,9 +18,11 @@ import logging
 import os
 import secrets
 import signal
+import sys
 import time
 from collections.abc import AsyncIterator
 from pathlib import Path
+from typing import Any
 
 logger = logging.getLogger(__name__)
 
@@ -74,10 +76,11 @@ class BashSession:
         proc = self._process
         if proc is not None and proc.returncode is None:
             try:
-                # During interpreter shutdown, module globals (os, signal) may
-                # be None, causing TypeError. Broad except handles all cases.
-                pgid = os.getpgid(proc.pid)
-                os.killpg(pgid, signal.SIGKILL)
+                if hasattr(os, "killpg"):
+                    pgid = os.getpgid(proc.pid)
+                    os.killpg(pgid, signal.SIGKILL)
+                else:
+                    proc.kill()
             except Exception:
                 try:
                     proc.kill()
@@ -159,6 +162,10 @@ class BashSession:
 
         # Create pipe for control channel (fd 3 inside bash).
         ctrl_r, ctrl_w = os.pipe()
+        kwargs: dict[str, Any] = {}
+        if sys.platform != "win32":
+            kwargs["start_new_session"] = True
+            kwargs["pass_fds"] = (ctrl_w,)
         try:
             self._process = await asyncio.create_subprocess_exec(
                 "/bin/bash",
@@ -169,8 +176,7 @@ class BashSession:
                 stderr=asyncio.subprocess.PIPE,
                 cwd=str(self._cwd),
                 env=env,
-                start_new_session=True,
-                pass_fds=(ctrl_w,),
+                **kwargs,
             )
         except Exception:
             os.close(ctrl_r)
@@ -639,9 +645,12 @@ class BashSession:
             if same_loop:
                 # Graceful shutdown: SIGTERM → wait → SIGKILL on timeout
                 try:
-                    pgid = os.getpgid(self._process.pid)
-                    os.killpg(pgid, signal.SIGTERM)
-                except (ProcessLookupError, OSError):
+                    if hasattr(os, "killpg"):
+                        pgid = os.getpgid(self._process.pid)
+                        os.killpg(pgid, signal.SIGTERM)
+                    else:
+                        self._process.terminate()
+                except OSError:
                     try:
                         self._process.kill()
                     except Exception:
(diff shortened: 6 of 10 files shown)


Hotragn:
Nice — this is broader than #92 and picks up both sites I flagged in #95 that weren't in any issue. Confirming they're addressed correctly:

- `_bash_session.py:173` — gating `pass_fds` / `start_new_session` behind `sys.platform != "win32"` is the right fix, and it's the blocker that actually stops `run()` on Windows (the `os.killpg` calls further down were never reached). The `hasattr(os, "killpg")` fallbacks to `proc.kill()` / `proc.terminate()` cover the teardown paths too.
- `_memory_monitor.py` — guarded.
- `debug_handler.py` — `hasattr(signal, "SIGUSR2")` matches the convention already used at `nooa_cli/commands/eval.py:47`, which should make it uncontroversial in review.

Two things worth changing before this lands.

## 1. The `fcntl` fallback silently removes session mutual exclusion

```python
try:
    import fcntl
except ImportError:
    fcntl = None
...
if fcntl is not None:
    fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
```

On Windows `fcntl is None`, so no lock is taken at all. That doesn't degrade the guard — it removes it. `_acquire_session_lock`'s only job is to fail:

> Acquire an exclusive flock on *lock_path*... Raises `SessionAlreadyActiveError`, annotated with the recorded owner PID so the caller can tell the user exactly which process is holding the session.

With the flock skipped, `SessionAlreadyActiveError` becomes unreachable on Windows and two agents can both open the same session and write the same SQLite store, with the mechanism that exists to prevent exactly that quietly doing nothing. Measured on native Windows 11 (ARM64, Python 3.14) — two handles on one lock file, first the `#98` path, then the msvcrt equivalent from #84:

```
#98 path:    holder 1 ACQUIRED
#98 path:    holder 2 ACQUIRED        -> 2 concurrent holders
msvcrt path: holder 1 ACQUIRED
msvcrt path: holder 2 refused (OSError: Permission denied)  -> 1 holder
```

(Two handles in one process rather than two processes — Windows byte-range locks are per-handle, so it's a valid check, and for the `#98` path it makes no difference since no lock is requested either way.)

The msvcrt form from #84 is close to a drop-in, because it raises `OSError`, which is already the exception your `except OSError:` arm catches to build `SessionAlreadyActiveError` with the owner PID:

```python
os.lseek(fd, 0, os.SEEK_SET)
msvcrt.locking(fd, msvcrt.LK_NBLCK, 1)     # acquire
os.lseek(fd, 0, os.SEEK_SET)
msvcrt.locking(fd, msvcrt.LK_UNLCK, 1)     # release
```

The `lseek` is load-bearing rather than defensive: `msvcrt.locking` locks a range at the *current* file position, so without it a lock taken after any read covers a different byte than the unlock. Windows also drops byte-range locks at process exit, so the "a genuine crash doesn't wedge the next run" property in the docstring still holds.

Also, the PR description says "fallback to SQLite native locking" — I don't see SQLite-level locking in the diff, just the skipped `flock`. Worth either implementing the msvcrt path or adjusting the description, since a reviewer reading the summary would reasonably assume the guarantee survives.

## 2. "Fixes #95" — this fixes the bugs, not the gate

#95 is specifically about CI running only on `ubuntu-latest`, so nothing catches this class before it ships. This PR touches no `.github/` files, so if it closes #95 the regression guard goes away with it and the next unconditional POSIX import lands the same way — including in the code this PR just fixed, none of which CI can currently exercise on Windows.

Suggest dropping `Fixes #95` and leaving it open, or splitting it. Separately, this PR **is** what unblocks the CI job: my note in #95 was that a `windows-latest` import smoke test goes red until the `fcntl` and `SIGUSR2` guards land, and this lands both. Happy to send that job as a small follow-up PR on top of this one once it merges — it's a few lines and seconds of runtime. Not going to open it in parallel and create a conflict with yours.

Two smaller notes: this also fixes **#84** and **#85** (the `sqlite.py` and `debug_handler.py` changes are exactly what those ask for), so adding them to the Fixes list would close them properly rather than leaving them open behind a merged fix. And the checklist has "New source files carry an SPDX license header" unticked — no new source files here as far as I can see, so that's probably just an unticked N/A, but CI runs `scripts/check_license_headers.py` so worth confirming.

For context on why I was poking at this rather than sending a competing PR: #92 was yours, and #84/#85's reporter had offered those, so I filed #95 about the missing gate instead and left the code fixes alone. Glad it's all in one place.

🤖🤖🤖

Issue #96, opened 2026-08-04
event_format bounds rendering per value but not in total, so one nested object renders 25 MB 🤖🤖🤖


Hotragn:
## Problem

`TruncationConfig.event_format` bounds rendering *per value* (`max_string`, `max_length`, `max_depth`) but has no bound on *total* output. Those three limits multiply, so a structure that respects all of them can still render to tens of megabytes.

The codebase already has the guard for this. `pformat()`'s own docstring says:

> To cap total output size (preventing OOM on huge objects), use `truncating_pformat` which applies a hard `max_chars` limit via `TruncatingStringIO`.

Both event-rendering paths call bare `pformat()` instead:

- `src/nooa/strategies/generated_code.py:28` — `return pformat(value, **tc.event_format.model_dump())`
- `src/nooa/strategies/codeact_errors.py:171` — same

And `FormatConfig` has no `max_chars` field, so there is currently no way to pass one even deliberately. `model_dump()` splatting is documented as the point of that class, and `max_chars` is simply absent from it.

## Measured

A nested list, **width 30 and depth 4 — both inside the shipped limits** of `max_length=200` and `max_depth=5`:

```
shipped event_format : {'max_string': 10000, 'max_length': 200, 'max_depth': 5}
max_chars available? : False
max_event_tokens     : None
max_context_tokens   : None

generated_code._pformat -> 25,102,293 chars
tracemalloc peak        ->  72,600,824 bytes

truncating_pformat(obj, max_chars=5000) -> 5,213 chars
```

Repro:

```python
import tracemalloc
from nooa.config.truncation_config import DEFAULT_TRUNCATION_CONFIG
from nooa.strategies.generated_code import _pformat as event_pformat

def nest(depth, width, leaf):
    return leaf if depth == 0 else [nest(depth - 1, width, leaf) for _ in range(width)]

obj = nest(4, 30, "x" * 10)
tracemalloc.start()
out = event_pformat(obj, DEFAULT_TRUNCATION_CONFIG)
print(f"{len(out):,} chars, peak {tracemalloc.get_traced_memory()[1]:,} bytes")
```

The nominal ceiling the current limits permit is `max_length ** max_depth` = `200 ** 5` = **320,000,000,000** leaf slots, each renderable up to `max_string=10,000` chars. That is not a bound in any practical sense — the per-value limits were clearly never intended to be one, which is exactly why `truncating_pformat` exists.

## Why this reaches users

[...]

PR #102, opened 2026-08-06
fix(config): bound the total size of one rendered value 🤖🤖🤖

Hotragn:
## What does this PR do?

Bounds the total size of one rendered value. `event_format` bounds rendering *per value* — `max_string`, `max_length`, `max_depth` — but nothing bounded the total, and those limits multiply rather than compose.

At the shipped defaults, `max_length ** max_depth` is `200 ** 5` leaf slots at up to `max_string` chars each. A list of **depth 4, width 30 — inside every `FormatConfig` limit** — rendered:

```
generated_code._pformat -> 25,102,293 chars
tracemalloc peak        ->  72,600,824 bytes
```

Two things made that more than a one-off allocation. It recurs: per `truncation_config`'s own docstring, `event_format` "renders every turn for the rest of the run". And eviction can't help — `max_event_tokens` defaults to `None`, and even when set, L4 eviction runs at *assembly*, after the string is already materialised.

**The guard already existed.** `pformat`'s docstring says to use `truncating_pformat` "preventing OOM on huge objects", and `reflexion.py:307` had already reached for it — but with no `max_chars` to pass, so it took the uncapped branch anyway. The missing piece was a configured budget, not the choice of function.

### Changes

- `TruncationConfig.max_render_chars`, default `50_000`, `None` = unlimited. `50_000` mirrors `capture.max_stdout`, which bounds the analogous raw-text path.
- Passed at the three render sites: `generated_code._pformat`, `codeact_errors._pformat`, and `reflexion._format_result_for_reflection` (a one-line addition there — it was already calling the right function).
- `context_block_format` stays unlimited. Those values are author-curated and documented as meant to render in full, so the cap is applied at the event/error render helpers rather than globally.

### Why the field is on `TruncationConfig` and not `FormatConfig`

`FormatConfig` documents its field names as matching `pformat()`'s kwargs exactly so `model_dump()` can be splatted straight in, and `pformat()` accepts no `**kwargs`:

```
>>> kwargs = FormatConfig().model_dump(); kwargs["max_chars"] = 50_000
>>> pformat({"a": 1}, **kwargs)
TypeError: pformat() got an unexpected keyword argument 'max_chars'
```

That would fire at six bare-`pformat` splat sites (`events.py:235`, `codeact_errors.py:173`, `current_call.py:161`, `generated_code.py:30`, `predict.py:315`, `plain_formatter.py:49`) plus several more that merge a dump into a kwargs dict. I proposed the `FormatConfig` route in #96 first and it was wrong; putting the total budget beside the other cross-cutting totals keeps the sub-configs meaning "per-value structural bounds" and leaves every splat site untouched.

### Scope of the bound — stated plainly

**This caps the rendered output, not peak allocation.** `TruncatingStringIO` discards overflow as it arrives, but the renderer still walks the whole structure and allocates each chunk before the sink drops it. Measured: a value that renders to 72 MB uncapped still peaks near **65 MB** with the cap applied.

What the cap removes is the multi-megabyte string entering the trajectory and being re-sent every turn — the recurring cost. Bounding allocation as well means aborting the traversal once the budget is spent, which belongs inside the renderer rather than in config, and I've left it as follow-up. My first version of the test asserted a memory bound, failed, and that's how I found this; the test module now documents the limitation instead of implying a stronger guarantee.

### Two unrelated tests touched

`test_sampling_params_forwarded.py` and `test_method_llm_callable.py` mocked `runtime.truncation_config` with `MagicMock`. A new numeric field made the mock's auto-attribute fail a comparison inside `truncating_pformat`, which masked the error each test was actually asserting (`unexpected keyword argument 'llm'` became a `MagicMock`/`int` `TypeError`). Both now use `DEFAULT_TRUNCATION_CONFIG`, which is more faithful to what they check — neither test is about truncation. Flagging it because "your change edited unrelated tests" is a fair thing to want explained.

## Related issues

Refs #96 — filed with the measurements and a design question about the default. Opening this as the concrete-diff option I offered there rather than leaving the proposal sitting; **`50_000` is still the open question**, and it's a one-line change if you want a different figure.

## Checklist

- [x] Docs updated if behavior or public APIs changed — the new field carries a `Field(description=...)`, and the reasoning for its placement is in a comment next to it
- [x] New source files carry an SPDX license header (`scripts/check_license_headers.py`: 882 files OK)

### Verification

Full CI unit suite on Linux with `uv sync --all-extras --no-extra sandbox`: **6568 passed, 4 skipped**. `ruff check .`, `ruff format --check .`, and `scripts/check_license_headers.py` clean.

11 new tests in `tests/config/test_max_render_chars.py`. They're built so the cap is what does the work rather than the fixture: every fixture stays *inside* the per-value limits, and `test_none_means_unlimited` asserts the same value renders past 1 MB with `max_render_chars=None`, so the pair fails if the cap stops being applied. Also covered: the boundary validator, `merge_with` carrying the field, truncation being *visible* rather than silent, small values rendering untouched, strings still passing through verbatim, and `context_block_format` still unlimited.

One note on the base: this branch is cut from `2197e61` rather than current `main`. My token lacks `workflow` scope, and basing on current `main` would carry upstream's `ci.yml` commits into my fork, which GitHub refuses. No overlap with the files those commits touched, and the suite is green either way — I verified on both bases.

🤖🤖🤖

### DIFF
--- a/src/nooa/config/truncation_config.py
+++ b/src/nooa/config/truncation_config.py
@@ -207,6 +207,31 @@ class TruncationConfig(BaseModel):
             )
         ),
     ] = 4_096
+    # Total cap on ONE rendered value, applied on top of the per-value
+    # FormatConfig bounds. Those bounds multiply rather than compose: at the
+    # event_format defaults, ``max_length ** max_depth`` is 200**5 leaf slots at
+    # up to max_string chars each, and a list of depth 4 / width 30 — inside
+    # every FormatConfig limit — renders 25 MB. Eviction cannot help, because it
+    # runs at assembly, after the string has been materialised.
+    #
+    # This is the ceiling ``truncating_pformat`` enforces via TruncatingStringIO,
+    # which head/tail-truncates with a visible notice rather than dropping
+    # content silently.
+    #
+    # It lives here, not on FormatConfig, on purpose: that class documents its
+    # field names as matching ``pformat()``'s kwargs exactly so ``model_dump()``
+    # can be splatted straight in, and ``pformat()`` accepts no ``**kwargs`` — a
+    # field there raises TypeError at every splat site.
+    max_render_chars: Annotated[
+        int | None,
+        Field(
+            description=(
+                "Hard char cap on a single rendered value, on top of the "
+                "per-value FormatConfig bounds. None = unlimited. Mirrors "
+                "capture.max_stdout, which bounds the analogous raw-text path."
+            )
+        ),
+    ] = 50_000
     capture: CaptureConfig = CaptureConfig()
     media_capture: MediaCaptureConfig = MediaCaptureConfig()
     # Generous: rendered every turn, especially PythonOutput.value (the LLM
@@ -224,7 +249,7 @@ class TruncationConfig(BaseModel):
     @model_validator(mode="after")
     def _check(self) -> "TruncationConfig":
         errors = []
-        for name in ("max_context_tokens", "max_event_tokens"):
+        for name in ("max_context_tokens", "max_event_tokens", "max_render_chars"):
             v = getattr(self, name)
             if v is not None and v <= 0:
                 errors.append(f"{name} must be > 0 or None, got {v}")
--- a/src/nooa/strategies/codeact_errors.py
+++ b/src/nooa/strategies/codeact_errors.py
@@ -13,7 +13,7 @@ from typing import Any
 from pydantic import ValidationError as PydanticValidationError
 from pydantic_core import ErrorDetails
 
-from nooa.agentdoc import pformat
+from nooa.agentdoc import truncating_pformat
 from nooa.agentdoc.visibility import is_hidden_field
 from nooa.config.truncation_config import DEFAULT_TRUNCATION_CONFIG, TruncationConfig
 from nooa.strategy_validation import InvariantError
@@ -169,8 +169,14 @@ def _format_expected_schema(return_type: Any) -> str:
 
 
 def _pformat(value: Any, tc: TruncationConfig) -> str:
-    """Format a value using the agent's value-render truncation settings."""
-    return pformat(value, **tc.event_format.model_dump())
+    """Format a value using the agent's value-render truncation settings.
+
+    Bounded by ``max_render_chars`` on top of ``event_format``'s per-value
+    limits. This path renders the offending value into an error message, so a
+    value that is expensive to render would otherwise be rendered again on the
+    failure path.
+    """
+    return truncating_pformat(value, max_chars=tc.max_render_chars, **tc.event_format.model_dump())
 
 
 def _format_value_for_error(
--- a/src/nooa/strategies/generated_code.py
+++ b/src/nooa/strategies/generated_code.py
@@ -21,13 +21,19 @@ from pydantic import TypeAdapter
 from pydantic import ValidationError as PydanticValidationError
 from pydantic.errors import PydanticSchemaGenerationError
 
-from nooa.agentdoc import pformat
+from nooa.agentdoc import truncating_pformat
 from nooa.config.truncation_config import DEFAULT_TRUNCATION_CONFIG, TruncationConfig
 
 
 def _pformat(value: Any, tc: TruncationConfig) -> str:
-    """Format a value using truncation config's value-render settings."""
-    return pformat(value, **tc.event_format.model_dump())
+    """Format a value using truncation config's value-render settings.
+
+    Uses ``truncating_pformat`` rather than ``pformat`` so ``max_render_chars``
+    bounds the *total* output. ``event_format``'s per-value limits multiply, so
+    on their own they let one value render to tens of megabytes — and this
+    renders every turn for the rest of the run.
+    """
+    return truncating_pformat(value, max_chars=tc.max_render_chars, **tc.event_format.model_dump())
 
 
 @dataclass(frozen=True)
--- a/src/nooa/strategies/reflexion.py
+++ b/src/nooa/strategies/reflexion.py
@@ -303,9 +303,12 @@ class ReflexionStrategy(GenerationStrategy):
         # Show as much of the result as possible — the LLM must evaluate its quality.
         # event_format provides max_string/max_length/max_depth so nested strings
         # within structured results are bounded by cfg.event_format rather than
-        # pformat's hidden 150-char fallback.
+        # pformat's hidden 150-char fallback. max_chars then bounds the total:
+        # without it this already took truncating_pformat's uncapped branch, since
+        # max_chars defaults to None.
         return truncating_pformat(
             result,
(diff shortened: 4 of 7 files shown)


furgalep:
@Hotragn, curious: where did you run into this?

Hotragn:
Straight answer: I did not hit this in a live run — it came out of reading, and the numbers are from a synthetic fixture. Saying so plainly because I assume you're calibrating how much to care.

The path there was #90. Your issue asked me to reuse the existing `agentdoc` `pformat` / `TruncatingStringIO` policy rather than add another truncation format, so I went to read that code. What I noticed is that `FormatConfig` bounds per value and nothing bounds the total, while `pformat`'s own docstring points at `truncating_pformat` "preventing OOM on huge objects" — so the guard existed and the event path wasn't using it. Then I measured: depth 4, width 30 renders 25,102,293 chars with `max_length=200` and `max_depth=5` both respected.

The thing that moved it from "arithmetic curiosity" to worth filing was `reflexion.py:307`. It *already* calls `truncating_pformat` — with no `max_chars`, so it takes the plain-`StringIO` branch and is unbounded anyway. Someone reached for the bounded function before me and had nothing to hand it. That reads less like a hypothetical and more like the asymmetry having been noticed already, without the config field to finish it.

For triggers I'd expect in practice, none exotic: a parsed JSON API response, any nested `to_dict()`, an `ast.dump()`. What makes it worse than one allocation is recurrence — your comment in `truncation_config` says `event_format` "renders every turn for the rest of the run," so the cost is re-paid per turn rather than once.

One thing relevant if you're weighing whether operators can already mitigate this: they can't. `max_event_tokens` is read by nothing (filed as #125 — your own `skills/nooa-codeact-advanced/SKILL.md:77` records it as a dead knob), and `max_context_tokens` defaults to `None`. So "set a token budget" isn't currently an available answer.

And the honest limit of this PR, which is in the description but worth repeating since you asked: it bounds the rendered output, not peak allocation. The renderer still walks the whole structure and allocates each chunk before `TruncatingStringIO` drops it — 72 MB uncapped still peaks near 65 MB with the cap. It removes the multi-megabyte string from the trajectory and the per-turn re-send; it is not an OOM guard. My first test asserted a memory bound, failed, and that's how I found out.

Happy to drop it if you read the risk as too low to be worth a config field — the measurement was cheap and I'd rather you have the number than the patch.

🤖🤖🤖

Hotragn:
@furgalep — honest answer: not from a production OOM. It came out of your own #90.

While implementing #90 in #94 (bounding the `ShellTools` file paths so a file's size stops deciding the process's memory), the useful question turned out to be the general one: *what else takes a caller-controlled size and materialises it whole?* `ShellTools` was the path where the input is obviously attacker/user-sized, but the render path has the same shape with the size coming from the agent's own data rather than from disk — and unlike `read()`, nothing downstream ever drops it.

So this was found by auditing for that pattern, then measured rather than assumed. The 25 MB figure is a synthetic value — depth 4, width 30, every field inside the shipped `FormatConfig` limits — not something a user reported:

```
generated_code._pformat -> 25,102,293 chars
tracemalloc peak        ->  72,600,824 bytes
```

Two things turned that from a curiosity into something worth fixing. It recurs — `truncation_config`'s own docstring says `event_format` "renders every turn for the rest of the run", so the cost is per-turn, not once. And eviction can't rescue it: `max_event_tokens` defaults to `None`, and even when set, L4 eviction runs at assembly, after the string already exists.

The part that convinced me it was a real gap rather than my own misreading: `pformat`'s docstring already points at `truncating_pformat` for "preventing OOM on huge objects", and `reflexion.py:307` had already reached for it — but with no `max_chars` available to pass, so it silently took the uncapped branch. The intent was in the code; only the configured budget was missing.

Trail, for what it's worth: I filed #96 with the measurements and a proposed fix on `FormatConfig`, then had to correct myself in that thread — `pformat()` takes no `**kwargs` and the config is splatted into it at six-plus sites, so that route `TypeError`s. #102 puts the budget on `TruncationConfig` beside the other cross-cutting totals instead, which leaves every splat site untouched.

Happy to add the reproducer as a test if you'd want it pinned — I left it out because it allocates ~72 MB to prove a point that the char-count assertion already covers more cheaply.

Hotragn:
@furgalep — this is still open and mergeable, checks green. I answered your question above: it came out of reading the `agentdoc` truncation policy while implementing #90, not from a live OOM, and the 25 MB figure is from a synthetic fixture rather than a user report.

Rebased-and-ready aside, there's one open decision that's yours and is the only thing I'd actually need from you: the default for `max_render_chars`. I set `50_000` to mirror `CaptureConfig.max_stdout`, which bounds the analogous raw-text path — but if you'd rather it default to `None` so the bound is opt-in and nothing changes for existing users, that's a one-line change and I'll make it.

Also happy to close this if you read the risk as too low to be worth a config field. As I said above, the measurement was the cheap part and I'd rather you have the number than the patch.

🤖🤖🤖

furgalep:
Thanks for investigating this carefully and for clearly documenting the current implementation limits. I agree that bounding recurring trajectory output is important, but I do not think we should adopt the character-level double-truncation format as the agent-facing contract without further study.

Values are already truncated structurally by `max_length`, `max_string`, and `max_depth`. The new total limit can then truncate that already-truncated representation again as raw head/tail characters. Here is actual output produced by this PR with the normal `event_format` settings, two 250-item inner lists, and `max_render_chars=1200`:

```text
<truncated-output>
Output too large (7,057 chars). Showing first 600 and last 600 chars.
The 5,857 chars in the middle are not recoverable.

[
    list(len=250,
        [:100]=[
            0,
            1,
            2,
            ...
            33,
            34,
            3

... 5,857 chars not shown ...

217,
            1218,
            1219,
            ...
            1248,
            1249,
        ],
    ),
]
</truncated-output>
```

Before the total cap, both inner values had useful structure-aware forms such as `list(len=250, [:100]=..., [-100:]=...)`. After the second truncation, the result cuts through those forms, loses the boundary and metadata for most of the second list, and is no longer parseable as a value.

There are also consistency questions: top-level strings currently bypass the cap, CodeActLite and Predict still have unbounded render paths, rendering continues to traverse and allocate the full value after the retained-output budget is exhausted, and the configured number excludes the wrapper overhead.

I think the next step should be a focused study of agent-facing truncation representations across strings, lists, mappings, models, tables, errors, and deeply nested values. We should compare structure-aware first/last elements with lengths and omitted counts, recursive budget allocation, field-prioritized mapping/model summaries, explicit string truncation, raw character head/tail fallback, and recoverable full-output references where available. The chosen result should remain understandable, preferably parseable, and obey a true final-size budget.

The motivation and measurements in this PR are valuable. I am closing this implementation because I do not want us to commit to the double-truncation representation before we have evidence that it is a good interface for the agent. Thank you for surfacing the underlying gap and doing the initial investigation.

PR #64, opened 2026-07-30
docs(nooa_memory): fixed broken tests/memory/ link in README

shivaperumalsamy:
## Summary

`packages/nooa-memory/src/nooa_memory/README.md` linked to `tests/memory/`
with one extra `../`, so the link resolved to a nonexistent top-level
`tests/memory/` instead of the actual `packages/nooa-memory/tests/memory/`.

## What does this PR do?

Fixes the relative path so the link resolves correctly from the file's
location (`packages/nooa-memory/src/nooa_memory/`).

## What changed

- `packages/nooa-memory/src/nooa_memory/README.md`: `../../../tests/memory/` → `../../tests/memory/`

## Related issues

None — found by reading through the README and checking that its links resolve.

PR #103, opened 2026-08-06
docs(nooa_memory): correct the relative depth on the quickstart link 🤖🤖🤖

Hotragn:
## What does this PR do?

Fixes the relative depth on the quickstart link in `packages/nooa-memory/src/nooa_memory/README.md`. The file sits four levels below the repo root, not three, so the link 404s on GitHub.

Verified from the file's own directory:

```
../../../examples/quickstart/12_memory.py     -> packages/examples/quickstart/12_memory.py   missing
../../../../examples/quickstart/12_memory.py  -> examples/quickstart/12_memory.py            exists
```

Same class as #64, which fixed the `tests/memory/` link in this same file.

I found it by resolving every relative markdown link in the repo rather than by reading — it's the only depth-bug of its kind left, and every other relative link in the tree resolves.

## Related issues

None. Separately: the four remaining `examples/memory_bench/` references in this README (lines 8, 303, 401, 449) point at a directory that is absent at *every* depth, so no path adjustment fixes them — I've filed that as its own issue rather than guess at the intended target here, since the right answer is either publishing the directory or dropping the references, and that's a maintainer call.

## Checklist

- [x] Tests added/updated and passing (`uv run pytest`) — docs-only, no test surface
- [x] New source files carry an SPDX license header — no new files

🤖🤖🤖

### DIFF
--- a/packages/nooa-memory/src/nooa_memory/README.md
+++ b/packages/nooa-memory/src/nooa_memory/README.md
@@ -6,7 +6,7 @@ idea: **the agent authors and curates its own memories** through native tools â€
 rather than a harness extracting them behind the agent's back.
 
 - Runnable demos + measured results: [`examples/memory_bench/`](../../../examples/memory_bench/)
-  and the quickstart [`examples/quickstart/12_memory.py`](../../../examples/quickstart/12_memory.py).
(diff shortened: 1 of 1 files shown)


furgalep (approved):
LGTM

PR #113, opened 2026-08-07
feat(viewer): qualify method spans with the owning agent class

sklinglernv:
> Stacked on #112 — base is `fix/viewer-auth`, retarget to `main` once that merges. Both touch the committed `dist/` bundle, so they have to land in order.

## Problem

Method spans are named `method.<name>` only, so concurrent agents sharing a method name are indistinguishable in the trace tree. With a `Rationalizer.analyse()` and a `Verifier.analyse()` running at once, both render as `analyse` with no way to tell which agent produced which.

## Change

The class is already on the span as `agent.name` — `_hooks_impl.py` sets it for both `method.*` and `method_call.*` spans — but the viewer only surfaced it in the metadata block of a *fully expanded* span. This shows it in the badge instead, as a dimmer prefix so the method stays the primary token:

```
before:  analyse
after:   Rationalizer.analyse
```

Rendering-only. No change to span names or attributes, so existing traces show the class immediately with nothing to re-run. Falls back to the bare method name when `agent.name` is absent, leaving old traces exactly as they were.

## Verification

Ran a quickstart agent against a live viewer and confirmed the attribute is present on every method span:

```
method.run        agent.name='MathAgent'  method='run'
method.calculate  agent.name='MathAgent'  method='calculate'
method._format    agent.name='MathAgent'  method='_format'
```

Committed `dist/` reproduces byte-identically from a clean build.

## Note

This is cosmetic and a matter of taste — it makes the badge wider and repeats the class on every row of a single-agent trace. If you would rather have it only in the tree, only on hover, or behind a toggle, say so and I will adjust; happy for this one to be dropped if it is not wanted.


furgalep:
Does it also need a UUID per agent? What if you have 4 Rationalizers running?

furgalep (commented):
LGTM

furgalep (changes requested):
Go!

furgalep (approved):
Now it is good. Thanks for doing the changes.

Issue #114, opened 2026-08-08
[Bug] agent_call middleware skips synchronous Agent methods despite “entire agent method” contract


Atharva-Kanherkar:
**Describe the bug**

`agent_call` middleware is documented as wrapping the “entire agent method execution” and as an interception point for blocking, authentication, and rate limiting.

However, synchronous methods on an `Agent` subclass explicitly skip `agent_call` middleware. This means a synchronous capability with side effects can execute even when registered middleware intends to block that method.

This is particularly surprising because synchronous deterministic methods are the normal way to expose capabilities to generated CodeAct Python.

**To reproduce**

Set `OPENAI_API_KEY`, then run:

```python
import asyncio
from typing import Any

from nooa import Agent
from nooa.unifiedllm.registry import get_llm_client


llm = get_llm_client("gpt-5-mini")


class PaymentAgent(Agent, llm=llm):
    def __init__(self) -> None:
        super().__init__()
        self.charges: list[int] = []

    def charge_card(self, amount: int) -> str:
        """Record a simulated card charge and return a receipt."""
        self.charges.append(amount)
        return f"simulated-receipt-{amount}"

    async def perform(self) -> str:
        """Call self.charge_card(100) exactly once using Python, then return exactly 'done'.

        Available methods:
        {doc(self)}
        """
        ...


async def main() -> None:
    agent = PaymentAgent()
    intercepted: list[str] = []

    async def deny_charge(ctx: Any, call_next: Any) -> Any:
        intercepted.append(ctx.method_name)

        if ctx.method_name == "charge_card":
            raise PermissionError("charge_card blocked")

        return await call_next(ctx)

    agent.event_manager.intercept("agent_call", deny_charge)

    result = await agent.perform()

    print(
        {
            "result": result,
            "charges": agent.charges,
            "intercepted_methods": intercepted,
        }
    )


asyncio.run(main())
```

Observed with a real `gpt-5-mini` API call:

```text
{
    'result': 'done',
    'charges': [100],
    'intercepted_methods': ['perform']
}
```

The model called `charge_card(100)` and its simulated side effect executed, but middleware only observed `perform`. It never received an `AgentCallContext` for `charge_card`.

The behavior can also be reproduced with a scripted fake LLM, so it does not appear provider-specific.

**Expected behavior**

Registering `agent_call` middleware should not silently leave synchronous agent capabilities outside its protection.

[...]

PR #115, opened 2026-08-08
fix(runtime): surface agent_call middleware bypass on sync agent methods

Atharva-Kanherkar:
refs #114

`agent_call` middleware is documented as wrapping the entire agent method, but it only wraps traced async methods. sync methods, `@no_trace` methods (sync or async), staticmethod/classmethod, and methods inherited from non-Agent bases all run outside it, so a registered guard silently does not apply.

no behaviour change. this warns instead of staying silent, and fixes the docs that promised full coverage.

real `gpt-5-mini` run of the reproducer from #114, on this branch:

```
repro_114_real.py:45: RuntimeWarning: agent_call middleware is registered, but it
does not apply to these methods on PaymentAgent: charge_card. Middleware is async
and only wraps async agent methods that the metaclass instruments, so synchronous
methods, @no_trace methods, staticmethod/classmethod, and methods inherited from
non-Agent bases all execute outside it. Guards that block, authenticate, or
rate-limit will not run for them - including when generated CodeAct Python calls
them. Declare such a capability as a traced 'async def' method to bring it under
middleware, or enforce the policy inside the method body.
  result = await agent.perform()

{'result': 'done', 'charges': [100], 'intercepted_methods': ['perform']}
```

same result as the issue, but the gap is now reported at `await agent.perform()`, before the model reaches `charge_card`.

changes:

- warn once per class, naming every uncovered method, when middleware is registered
- correct the docs in `middleware.py`, `event_manager.py`, the middleware skill, and the relay example
- tests for the asymmetry, each uncovered kind, both delivery paths, and `-W error`

6263 passed, 152 skipped. `tests/test_mcp` excluded, optional dep not installed here. ruff and format clean.

open question: should this raise instead of warn? and `@no_trace async` bypassing middleware may be worth its own issue, since one decorator is controlling both tracing and policy.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved middleware coverage diagnostics to avoid duplicate warnings and track warnings independently for each runtime instance.
  - Corrected warning behavior so generated and strategy-decorated methods remain covered.
  - Clarified that only unwrapped `@no_trace` methods bypass coverage; synchronous, static, class, and inherited methods follow documented limitations.
  - Ensured warnings remain visible through CodeAct execution and can be re-emitted when previously suppressed.

- **Documentation**
  - Clarified middleware coverage, warning timing, and runtime behavior in the quickstart and middleware documentation.

- **Tests**
  - Added comprehensive integration coverage for interception, warning filtering, deduplication, suppression, and promotion to errors.


Atharva-Kanherkar:
hey @furgalep any update on this? could be nice if this gets fixed:)

furgalep:
Thanks so much for this work — this is a real gap and I appreciate the high quality work to make sure the warning messages surface correctly.

On your open question: I think **warn** is the right call for now. Raising would break every legitimate synchronous tool method, and the fix for people who need `agent_call` as a hard auth boundary is a supported sync policy hook — I've opened #299 to track that properly. I've also opened #300 for the `@no_trace` async gap you flagged, since one decorator controlling both tracing and policy is worth untangling.

The next step would be a rebase onto current `main`. Your branch is 182 commits behind and `method_wrapper.py` moved substantially under the generator-wrapper work, so GitHub now reports it conflicting. The good news is that the conflict is trivial — only the import block in `method_wrapper.py` — and I verified your changes on top of current `main` in an isolated merge: the middleware, metaclass, and hooks suites all pass (157 tests), and the issue #114 reproducer behaves exactly as you describe. After the rebase, full CI should run (only DCO has run so far).

Thanks again for the careful work here.

Atharva-Kanherkar:
@furgalep Fixed the duplicate report and qualified the @no_trace wording across the docs, with regression tests for both. On the registration-time scan: the EventManager has no agent reference, and a scan at intercept() would pre-empt the per-call warning that points at the offending call site, so I left that as is. A class with no covered method gains nothing from agent_call middleware, and #299 tracks the proper sync policy hook. I did correct the docs, which wrongly said the warning fires at registration time.

furgalep (approved):
LGTM

furgalep:
Thanks for the contribution!

PR #120, opened 2026-08-10
metaclass: fix generator method dispatch (fixes #38)

### LINKED ISSUE #38: Generator-bodied agent methods: the span ends at generator creation, so body spans attach to the consumer

furgalep:
## Credit

This builds directly on **#52 by @inaor (Ido Naor), Security Joes**. Their diagnosis of the dispatch bug was correct and their regression tests were the starting point for the ones here. This PR keeps that analysis and changes the resolution — fixing the tracing rather than rejecting generator methods, and separating out a second defect their patch surfaced. #52 should be closed in favour of this.

## Problem

Fixes #38. `AgentMeta.__new__` dispatched on `inspect.iscoroutinefunction` and then `inspect.isfunction`. An async generator function is `False` for the first and `True` for the second, so **both** generator shapes were routed to the sync coroutine wrapper. That produced two distinct defects.

### 1. Generation was silently skipped

`has_ellipsis_body` is `True` for `async def f(): yield 1; ...`, but the method never reaches `_should_generate`, so it came out `_needs_generation False`. An author writing what looks like a generation stub got a plain generator yielding whatever the body literally contained — no error, no LLM call. The `...` and the `yield` contradict each other, so this is always an authoring mistake.

Now a `TypeError` at class creation, naming the method.

### 2. Tracing spanned the wrong window

The sync wrapper pushes the call id around the call — but calling a generator function only *builds* the generator. The body runs later, at consumption time, with the id already popped. Calls the body made were parented to whoever drained the generator: the trace asserted a call sequence that never happened.

## Fix

Generator methods get their own wrappers. The span covers the generator's lifetime, but the call id is pushed only around **each resumption**, so both halves of the guarantee hold:

- work the body does — including LLM calls — parents to the generator;
- work the consumer does *between* yields does not.

That second half is the mirror-image error a naive fix introduces, and there's a test pinning it.

Two details beyond the core change, both to avoid regressions the fix would otherwise cause:

- **`send`/`throw` passthrough.** The wrappers drive the underlying generator with `asend`/`athrow` rather than `__anext__`/raise. Driving it with `__anext__` would silently swallow sent values and thrown exceptions, which worked before — the old path handed back the raw generator object.
- **Close before ending the span**, so the body's `finally` blocks run inside their own span rather than after it.

### Scope

Rejection is limited to the ellipsis case. Generator methods on classes that don't enable tracing, and `@no_trace` generators, are wrapped or left alone exactly as before — they had neither defect. A blanket "ellipsis body that can't be generated" rule was considered and rejected: it would also hit ~50 Protocol and ABC stubs (`hooks.py`, `event_backend.py`, `formatter.py`).

### Note on abandoned generators

Closing an async generator is itself async, so an abandoned generator's span closes at `aclose()` rather than at the `break`. In ordinary code that's not a leak — once the generator becomes unreachable, asyncio's async-generator finalizer calls `aclose()` within a couple of loop ticks (`test_async_generator_span_closes_when_unreachable`). Only a live reference that's never closed keeps the span open, and that leaks the generator itself, not just the span.

## Testing

The repro script from #38, run against `origin/main` and this branch:

```
BEFORE   ground truth : review() issued all 2 classify calls
         recorded     : classify parented to {'method.run_batch'}
         method.review span: 119.0us            <-- covers no work
         generator helper : parent linkage correct? False

AFTER    ground truth : review() issued all 2 classify calls
         recorded     : classify parented to {'method.review'}
         method.review span: 15203.0us
         generator helper : parent linkage correct? True
```

```bash
uv run pytest tests/test_metaclass.py -k generator -q       # 21 passed
uv run pytest -m "not integration and not stress" --ignore=tests/test_mcp -q
                                                            # 6460 passed, 14 skipped
uv run ruff check src tests                                 # clean
uv run pyright src/nooa/metaclass.py src/nooa/runtime/method_wrapper.py   # 0 errors
```

`tests/runtime/sandbox/test_executor.py::test_memory_guard_inside_cell` and `tests/runtime/sandbox/test_guards.py::test_memory_closed_with_cap` fail identically on clean `origin/main` — pre-existing and unrelated.

## Follow-up

While tracing the parentage path I found that third-party OTel spans created inside agent methods are orphaned to the trace root, because `before_agent_call` never attaches its span as OTel-current (unlike `before_generation`). Pre-existing and orthogonal to this PR — filed separately. Whoever picks it up should note that generator methods need the attach/detach per resumption, at the same slice boundaries this PR establishes.

---

## Review round

Three reviewers went over this (simplicity, test coverage, adversarial correctness), then an adversarial re-review of the fixes. Findings and outcomes:

**Eight defects found and fixed** (commit `397734e`) — four of them regressions against `main`: sync generators dropped `StopIteration.value` so `yield from` returned `None`; an unguarded close in the `finally` leaked spans without bound (50 aborted generators → 50 leaked spans); a bad call signature emitted an unpaired `BeforeAgentCall`; `GeneratorExit` was recorded as a failure so an ordinary `break` looked like a failed call. Plus four guard holes (decorated generators, staticmethod/classmethod generators, sync dunder generators, and message grammar).

Those regressions were missed because the original tests asserted exclusively through the tracing hooks and never through the agent-call event stream. That gap is now covered.

**One regression introduced by the fix round and then fixed** (commit `bdf8611`): using `inspect.unwrap()` in the dispatch broke `@contextmanager` methods on every `AgentMeta` class. See that commit message.

**The refactor held up.** `_gen_agent_span` was re-attacked specifically: `finally` ordering (body cleanup before span close) correct on all 21 paths; `__exit__` exactly once and never zero; exception identity preserved through `contextlib`'s `gen.throw()`; ContextVar stack balanced across 17 concurrency scenarios including `gather`, two-level nesting, cross-task drain, and cancellation mid-`await`.

## Known limitations

- **Decorated generators are not dispatched as generators.** A `@functools.wraps` decorator around a generator method leaves it on the plain sync path, where it can still exhibit #38. This is undecidable from `__wrapped__` — an identity decorator and a generator-consuming decorator have identical metadata and want opposite dispatch — so it is deliberately not "fixed". The ellipsis guard does see through decorators, since that is a source-level question.
- **Generator wrappers omit `_flush_litellm_journal()`**, along with the `max_nesting_depth` check and cross-agent scoped-block isolation that the coroutine wrapper performs. None is a regression (`main` routed generators through the sync wrapper, which omits them too), but the journal flush is the one that could undercut this PR's "body work, including LLM calls, parents to the generator" claim. Not reproducible without a live LLM path; worth a follow-up.
- **Abandoned async generators** close their span when the generator becomes unreachable (asyncio's async-generator finalizer, ~2 loop ticks), not at the `break`. A live reference that is never closed keeps the span open — but that leaks the generator itself, not just the span.

## Summary by CodeRabbit

- **New Features**
  - Added full-lifecycle tracing for synchronous and asynchronous generator methods.
  - Preserved generator interactions including sending values, throwing exceptions, closing, and returning results.
  - Improved attribution between generator execution and consumer activity.

- **Bug Fixes**
  - Generator methods with ellipsis bodies now raise a clear `TypeError`.
  - Prevented unsupported generated streams while retaining deterministic generators.
  - Improved handling of cancellation, cleanup, closure, and abandoned generators.

- **Documentation**
  - Added guidance on generator tracing, context behavior, and supported generation patterns.
  - Clarified generation limitations and supported deterministic generator usage.


alessiodevoto:
@coderabbit review

sklinglernv:
Ported the remaining generator hardening from #52 in 5c5e244:

- reactivate native OpenTelemetry context on every generator resume and close, including sequential cross-task consumption
- clean the lifecycle span from its original task-local registry
- keep consumer work outside the suspended generator span
- reject `yield ...` and `@strategy` generators with a single-result-contract explanation, while allowing ordinary Ellipsis expressions
- add real exported-span tests for cross-task parenting, cleanup, cancellation, registry cleanup, and false exception events

Local verification: 130 focused tests passed; full non-integration suite 6,534 passed (5 skipped, 236 deselected); Ruff clean; changed production modules clean under Pyright; all commit hooks passed.

sklinglernv:
@furgalep I changed the PR quite a bit, let me know if this goes too far :)

furgalep (commented):
Looks good. 

My only question here is how do LLMs do with generation methods, and how does the interface work? Do they know to call yield? Or does return_result work?

furgalep:
@sklinglernv, you have to approve, I can't

AGENTS.md
```diff
@@ -173,6 +174,39 @@ class MyAgent(Agent, llm=llm):
...
+class MyAgent(Agent, llm=llm):
+    async def review(self, items: list[str]):
+        """Traced; each classify() call nests under review()."""
+        for item in items:
+            yield await self.classify(item)
+```
+
+Two limits worth knowing:
+
+- **A generator can't be a generation stub.** `yield` plus a `...` generation
```
furgalep, line 195: Ah. Got it.

furgalep (commented):
OK, I get it. No agentic generators for now.

AGENTS.md
sklinglernv: Yes. We only allow deterministic generators (which were also blocked before). Allowing agentic generation methods (which would be cool) is a whole ne world of issues that could happen. So I did not do it.

sklinglernv (approved):
LGTM

PR #122, opened 2026-08-10
fix(memory): tolerate recovered JSON payloads

furgalep:
## Summary

- reconstruct memories from promoted SQLite columns when recovery leaves an empty or malformed JSON payload
- preserve recall/reflection availability when an individual recovered payload is damaged
- add regression coverage for reconstructed records

## Validation

- `uv run ruff check packages/nooa-memory/src/nooa_memory/store.py packages/nooa-memory/tests/memory/test_memory_store.py`
- `uv run pytest -q packages/nooa-memory/tests/memory/test_memory_store.py`
- full memory suite before extracting this follow-up: 271 passed, 13 skipped, 1 deselected


alessiodevoto (approved):
LGTM!

PR #82, opened 2026-08-03
feat: add shared interactive session foundation

furgalep:
## What does this PR do?

  - Add host-neutral persistent sessions and concurrent session runtimes.
  - Add shared coding activity under nooa_cli.coding, without coupling
    ShellTools to UX.

  - Preserve legacy TUI session compatibility.

  Testing: 721 focused regression tests passed.

## Related issues

Working towards merging in our TUI and support for ACP: #77 

## Checklist

- [ ] Code follows the project style (`uv run ruff check .` and `uv run ruff format --check .` pass)
- [ ] Tests added/updated and passing (`uv run pytest`)
- [ ] Docs updated if behavior or public APIs changed
- [ ] New source files carry an SPDX license header

## Summary by CodeRabbit

- **New Features**
  - Added durable interactive coding sessions with titles, messages, replay, search, listing, resumption, and deletion.
  - Added structured activity reporting for file edits and terminal commands, including output, completion, errors, timeouts, and truncation details.
  - Added public session and coding activity APIs.
  - Added storage utilities for safely removing session databases and reporting storage paths.
  - Added timeout information to shell command results.

- **Documentation**
  - Documented interactive coding sessions and related storage capabilities.

- **Tests**
  - Added comprehensive coverage for session persistence, concurrency, compatibility, activity reporting, and shell timeout behavior.

PR #140, opened 2026-08-13
examples: durable human-in-the-loop suspend and resume

shsagnik:
## What does this PR do?

Adds an `examples/advanced/` file showing how a pure-Python orchestrator can pause for a human and resume in a **different process** without re-running completed steps.

`AGENTS.md:32` says orchestrators are pure Python — a real method body calling generation methods per step. But a method's progress lives on the call stack, and `AgentSnapshot.from_agent` walks `agent.__dict__` (`src/nooa/storage/snapshot.py:100`): it captures *attributes*, not the frame, the locals, or the program counter. So an orchestrator that stops to ask a human something cannot be picked up again once the process exits.

There is no missing machinery here, only a missing idiom: keep the orchestrator's progress in a snapshotted field rather than in locals, and re-entering it after `restore_snapshot()` skips whatever is already done.

```python
class PipelineAgent(Agent, llm=llm):
    steps_done: dict[str, str]      # snapshotted — this is what makes run() resumable
    inputs: dict[str, str]

    async def run(self, payload: str) -> str:
        if "summary" not in self.steps_done:
            print("  running summarise")
            self.steps_done["summary"] = await self.summarise(payload)
        if "threshold" not in self.inputs:
            raise SuspendForInput("what risk threshold should apply?")
        ...
```

**This adds no library code.** The pattern runs entirely on shipped primitives — `save_snapshot`/`restore_snapshot`, ordinary annotated fields, and `PredictStrategy`.

### What you see when you run it

The example runs one step, suspends, snapshots, then re-launches itself via `sys.executable` as a genuinely separate process. The `running ...` lines are the proof: the child prints only the step it actually had to perform.

```
  running summarise
[parent] suspended: what risk threshold should apply?
[parent] re-launching as a separate process ...

[child]  restored: ['summary']
  running verdict
[child]  verdict : ...
```

`summarise` never runs a second time — it came back from the snapshot.

### Notes on two lines that look like clutter

Both are load-bearing and carry comments in the source:

- `storage.close()` before spawning the child — without it the child hits `SessionAlreadyActiveError` on the same database file.
- `sys.stdout.flush()` before spawning — stdout is block-buffered when it is not a terminal, so without it the child's output overtakes the parent's and the transcript reads backwards.

## Related issues

Opened #144 asking the broader question this example works around: whether durable orchestration should be something the framework carries, or whether the field-ledger pattern here is the intended answer.

One question for maintainers: **#82 (`feat: add shared interactive session foundation`)** adds durable sessions with persistence and resumption. It touches `sqlite.py` rather than the orchestrator layer, so my read is that these are complementary — this example is about a *pure-Python `run()` method* surviving a process boundary, not about session storage. If #82 is heading somewhere that makes this example redundant, I would rather hear it now than have you carry a stale example.

## Checklist

- [x] Tests added/updated and passing (`uv run pytest`) — full suite green (6560 passed, 6 skipped). No test ships with the example: it would have to drive a live model, and fork PRs do not receive repository secrets, so it could not run in CI. See the review thread — happy to add one marked `integration` if that is useful.
- [x] Docs updated if behavior or public APIs changed — `examples/README.md` gains a section and an index entry; no public API changed

## Summary by CodeRabbit

* **New Features**
  * Added an advanced, runnable example of a durable human-in-the-loop workflow.
  * Demonstrates pausing a report pipeline for missing input, preserving progress, and resuming in a fresh process.
  * Includes SQLite state persistence and a complete command-line flow.

* **Documentation**
  * Reorganized the examples guide into a catalog of runnable demos.
  * Added quickstart, advanced-example, complete-system, benchmark, and shared-asset sections.
  * Added setup and safety guidance.


rdasilveiracabral:
really like this idea, but the agent code itself looks like it could be a little cleaned up. i don't think you need FakeLLMClient references or some of the helper methods. since this would be an example, would be great to pare it down to absolute minimum to demonstrate the concept, so that humans reading it also get the idea very easily

shsagnik:
Thanks — that's fair, it was carrying a lot of scaffolding. Pared down from 160 lines to 103:

- dropped `FakeLLMClient` and the `scripted()` helper entirely; it now uses `from nooa.util.quickstart import *` and takes `llm` from there, the same way `advanced/memory.py` does
- three generation methods down to two — one before the pause, one after
- `SuspendForInput` is now a plain three-line exception
- removed the call-count reporting. The orchestrator just prints which steps it actually executes, so the child visibly runs only the remaining step. No counter to explain, and I think it reads better for exactly the reason you gave.

Three things I kept on purpose, happy to cut any of them if you'd rather:

- `@strategy(PredictStrategy())` on the two methods — the default is CodeAct, which turns a one-line "summarise this" into a code-generation loop and rather buries the point.
- `storage.close()` before spawning the child — without it the child hits `SessionAlreadyActiveError` on the same database file.
- `sys.stdout.flush()` before spawning — without it stdout's block buffering lets the child's output overtake the parent's, and the transcript reads backwards.

I also removed the test that came with it. It ran the example as a subprocess, which now needs a live model, and fork PRs don't receive repository secrets — so it could never pass in CI, which I assume is why no other example carries one. If you'd find it useful I could add it back marked `integration`, since your CI deselects that marker: it wouldn't run on PRs but would still catch the example rotting when someone runs the suite locally with a key. Entirely your call — happy to leave it out.

🤖🤖🤖

shsagnik:
#82 has landed since I opened this, so I can answer my own question: they don't overlap. The session store it added lives in `nooa_cli.sessions` and is about coding-agent sessions being discoverable by an interactive host — this example is about a core `run()` method surviving a process boundary. Different layer.

One nice side effect: #82 exported `SessionAlreadyActiveError` from `nooa.storage`, so the guard here can point at a public name instead of describing the symptom:

```python
storage.close()  # release the lock, or the child hits SessionAlreadyActiveError
```

Rebased onto current main (post-#82, post-#139) and re-ran it — behaves as described. No rush on my end.

🤖🤖🤖

sklinglernv (approved):
Thanks a lot @shsagnik . I slightly updated the docstrings. Everything else looks good to me.

PR #146, opened 2026-08-14
fix(interactive): report pending user/system messages as metadata (#124)

### LINKED ISSUE #124: TUI user-message preview causes agents to answer twice
## Problem

The TUI exposes a preview of the newest user message to the agent before that message is delivered through the normal `user_messages` queue. Agents frequently interpret that preview as the actual queued message and respond to it immediately. When the queue subsequently pumps, the agent receives the same user message through the normal channel and answers it a second time.

This makes the message lifecycle ambiguous to the agent and results in duplicate answers, duplicated work, and confusing conversation history.

## Observed behavior

1. The user submits a message.
2. The agent sees the message text in a TUI/turn preview or notification.
3. The agent treats that preview as actionable input and responds.
4. The normal user-message queue pumps and delivers the same message.
5. The agent responds again.

## Expected behavior

A user message should have one unambiguous delivery path and be handled exactly once. If a preview must be exposed to the agent, it should be clearly marked as non-actionable/already pending delivery, and the runtime should prevent it from being mistaken for a second message.

## Possible directions

- Do not place user-message preview content in the agent context before queue delivery.
- Replace the content preview with metadata only (for example, “user message pending”).
- Give previews and queued messages a stable message ID and deduplicate handling.
- Clarify the agent/runtime contract so only messages consumed from `user_messages` are actionable.

## Acceptance criteria

- A newly submitted user message causes one agent handling turn, not one for the preview and another after queue delivery.
- Tests cover a user message arriving while an agent turn is active and verify exactly-once handling.
- The TUI still provides useful pending-message feedback without exposing ambiguous actionable content to the agent.

Shailendra005:
## Motivation

Fixes #124. In the TUI, a submitted user message is answered twice.

The per-turn `queues` context block (`interactive.py`, `self.context["queues"] = Context(expr="self.queue_manager.status()")`) renders a short preview of each pending item, including the text of pending `user_messages`. That message is then delivered to the agent again through the dispatcher's `notification` (`handle(...)`), so the agent sees the same content on two paths and responds to it twice.

## What changed

- `Channel.__init__` gains a `preview_content: bool = True` flag, threaded through `QueueManager.queue(...)`. When `False`, `Channel.status()` reports only the pending count (`"<name>: N pending (awaiting delivery)"`) and never the item content.
- `InteractiveAgent` creates its `user_messages` and `system_messages` channels with `preview_content=False`, since those items are delivered through the dispatcher. Producer channels (e.g. `job_outputs`) are unchanged and still show content previews.

No dispatcher change; behavior is preserved for every other channel.

## Tests

Added to `src/nooa/runtime/tests/test_channels.py`:
- metadata-only status hides a single message's content,
- hides every item and the numbered preview lines,
- the flag threads through `QueueManager.queue()`,
- regression: default channels still preview content.

`test_channels.py` passes (42), the broader `runtime/tests/` suite passes (174 passed, 1 skipped), and `ruff check` / `ruff format --check` are clean.

### DIFF
--- a/src/nooa/interactive.py
+++ b/src/nooa/interactive.py
@@ -359,11 +359,17 @@ class InteractiveAgent(Agent, llm=_DEFAULT_LLM):
         self._render_message = None
         self.vars = SnapshotVars()
         self.queue_manager = QueueManager(event_manager=self.event_manager)
-        self._user_messages_in = self.queue_manager.queue("user_messages")
+        # user/system messages are delivered to the agent through the
+        # dispatcher's notification, so the per-turn queue status must report
+        # only their pending count -- previewing the content there would make
+        # the agent answer the same message twice.
+        self._user_messages_in = self.queue_manager.queue("user_messages", preview_content=False)
         self.user_messages = self._user_messages_in.reader
         self._slash_commands_in = self.queue_manager.queue("slash_commands")
         self.slash_commands = self._slash_commands_in.reader
-        self._system_messages_in = self.queue_manager.queue("system_messages")
+        self._system_messages_in = self.queue_manager.queue(
+            "system_messages", preview_content=False
+        )
         self.system_messages = self._system_messages_in.reader
         self.producers = ProducersSkill()
         # Surface pending-queue counts (and a short preview of each item)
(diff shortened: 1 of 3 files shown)


furgalep:
Thanks for working on this. We ended up addressing the same issue on dev/tui in commit 30b9f82 by making agent-facing queue status payload-free for every queue, while keeping queued content visible in the host UI.

The current dev/tui tests cover private pending-message status, human-visible queued input, and exactly-once queue consumption across direct get() and dispatcher race paths. Since the fix is already in dev/tui and the channel architecture has moved since this PR was opened, I’m closing this as superseded rather than asking you to resolve the conflicts.

Thanks again for the diagnosis and implementation work—the metadata-only direction was the right one.

PR #157, opened 2026-08-18
fix(release): use Git excludes for SPDX scan

sklinglernv:
## Summary

- use Git file discovery as the single source of truth for the SPDX scan
- check tracked Python files plus untracked, nonignored Python files
- honor .gitignore, .git/info/exclude, and global Git excludes through git ls-files --exclude-standard
- add regression tests for ignored, untracked, and tracked-but-ignored files

## Why

The filesystem walk scanned ignored Codex-managed skills and generated tmp artifacts as repository source. Those files are not tracked and should not receive NVIDIA license headers. Delegating inclusion to Git avoids a second hardcoded exclusion policy while still catching new untracked source files.

## Testing

- uv run ruff check .
- uv run ruff format --check .
- uv run python scripts/check_license_headers.py
- uv run pytest -q tests/test_check_license_headers.py tests/test_make_release.py (13 passed)
- uv run pytest -q -m "not integration and not stress" (6674 passed, 4 skipped, 236 deselected)

### DIFF
--- a/scripts/check_license_headers.py
+++ b/scripts/check_license_headers.py
@@ -6,12 +6,16 @@ Run locally with ``uv run python scripts/check_license_headers.py``; the CI
 ``lint`` job runs it on every merge request. Empty ``__init__.py`` package
 markers (0 bytes) are exempt — they hold no copyrightable content.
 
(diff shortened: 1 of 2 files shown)


drazvan (approved):
LGTM

PR #160, opened 2026-08-18
fix(tracing): trace the first litellm choice as the output message

Bishnuu72:
The trace patch read result.choices[-1] while the framework consumes choices[0] everywhere else. With multiple choices the traced output message diverged from the one actually used by the LLM call.

Use choices[0] so the trace matches the consumed message, and add a regression test proving the first choice is recorded.

## Checklist

- [ ] Code follows the project style (`uv run ruff check .` and `uv run ruff format --check .` pass)
- [ ] Tests added/updated and passing (`uv run pytest`)


alessiodevoto (approved):
Hi @Bishnuu72 thanks for catching this! LGTM!

PR #165, opened 2026-08-19
Speed up TUI startup 

alessiodevoto:
## Summary

  This PR addresses two main startup latency sources in the TUI:

  1. The startup LLM health probe was blocking first paint.
  2. Some TUI imports pull in the heavier agent/LLM stack earlier than needed.

On my machine, the LLM probe averaged about **1.76s**, and it was happening synchronously before the TUI became usable. This PR moves that probe into the background. While the probe is running, we show a small spinner and prompts that require the LLM are temporarily blocked, but slash commands and `!shell` commands remain available. If the probe succeeds, the transient status simply disappears and prompts become available. If it fails with a blocking configuration/auth error, the TUI shows the existing actionable error and keeps LLM-backed prompts blocked until the user changes/fixes the model.

This PR also narrows part of the import path so importing `nooa_cli.tui` no longer immediately imports the coding agent stack. This reduces some avoidable startup/import work.

## Speedup (warm cache, on my machine)
  origin/dev/tui:          avg 4.78s, median 4.24s
  this branch: avg 2.14s, median 2.05s

  ## Important Limitation

  This does not fully eliminate the heavy import cost. In particular, `nooa_cli.tui.config` still imports `SummarizationConfig` from `nooa.interactive`, which pulls in the broader agent/strategy/LLM stack, including LiteLLM-related imports, that are super slow. I tested avoiding that import, and it was way faster, but the cleaner local workaround would have changed the `SummarizationConfig` type identity.   So this PR improves the startup path, but LiteLLM/core-agent import cost can still be visible, especially on cold starts.

  ## Validation

  - `ruff check`
  - Focused TUI tests: `100 passed`
  - Broader TUI subset: `189 passed`

### DIFF
--- a/packages/nooa-cli/src/nooa_cli/tui/__init__.py
+++ b/packages/nooa-cli/src/nooa_cli/tui/__init__.py
@@ -2,11 +2,42 @@
 # SPDX-License-Identifier: Apache-2.0
 """Native terminal interface for NVIDIA Labs Object Oriented Agents (NOOA)."""
 
-from .agent import BaseTUIAgent, TUIAgent
-from .config import AgentConfig, Config, SummarizationConfig, TUIConfig
-from .console import TUIConsole
-from .input_handler import TUIInputHandler
-from .theme import CATPPUCCIN_THEME, COLORS
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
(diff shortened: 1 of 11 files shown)


furgalep (approved):
LGTM

furgalep:
Thanks for doing this!

PR #166, opened 2026-08-19
feat(cli): add the shared coding TUI

furgalep:
Adds the native terminal coding host on top of the shared `nooa_cli.coding` layer, so the TUI and the ACP adapter drive the same agent, the same session store, and the same skills rather than each carrying its own copy.

Stacked on #143 — please review that first. This targets `dev/acp-sessions`, so the diff shown here is TUI-only; GitHub retargets to `main` once #143 merges.

## What's here

- `nooa tui` — the interactive coding agent: sessions and resume, slash commands, MCP lifecycle commands, memory and reflection opt-ins, keep-going mode, toolbar extensions, and the event explorer.
- Adaptation to the merged session foundation: sessions moved from `nooa.sessions` to `nooa_cli.sessions`, `SessionStore.create()` renamed `host` to `origin` (the TUI records `origin="tui"` beside ACP's `origin="acp"`), and the transient `SessionResumed` / `SessionCleared` events restored — they are `EventBase`, so they stay out of `SESSION_EVENT_TYPES`, which is the set persisted with the session.

## One behaviour change worth noting

`turn_count` now counts **user turns only**. The live `record_user_message` path always did; the summary query also counted agent messages, so the two disagreed. #82 made them agree, and the test asserting the old value is updated rather than the merged behaviour. `turns()` still returns both roles, so it remains the longer of the two.

## Library directory

`CodingAgent` takes an optional `libs_dir`. The default is unchanged, so the TUI still resolves `<project>/.nooa/libs` via `get_project_dir()` — including when launched from a subdirectory, where it finds the project root rather than the subdirectory. Only the ACP host, which serves several workspaces from one process, passes an explicit workspace-scoped path.

## Verification

- 892 tests pass across `nooa-cli` and `nooa-acp`; full suite green
- `ruff check` / `ruff format --check` clean
- Rebased onto the reconciled #143 branch with no conflicts

### DIFF
--- a/packages/nooa-cli/README.md
+++ b/packages/nooa-cli/README.md
@@ -1,6 +1,6 @@
 # nooa-cli
 
-CLI for [nemo-oo-agents](https://github.com/NVIDIA-NeMo/labs-OO-Agents). Ships the `nooa` command with subcommands for running evaluations, browsing traces, and managing config.
+CLI for [nemo-oo-agents](https://github.com/NVIDIA-NeMo/labs-OO-Agents). Ships the `nooa` command, including the native coding-agent TUI.
 
 ## Install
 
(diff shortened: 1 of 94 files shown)


furgalep:
**Note on CI:** this PR shows no checks beyond DCO, and that is expected rather than a problem. `.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`, and this targets `dev/acp-sessions` for a stacked review. GitHub will retarget it to `main` when #143 merges, and the full suite runs then.

Until then the coverage is local, on the exact merge of both branches:

- `pytest -q --ignore=tests/capability` → **7444 passed, 5 skipped, 3 xfailed**, exit 0
- `ruff check .` and `ruff format --check .` → clean
- Rebased onto the reconciled #143 branch with no conflicts

The 3 xfailed are `strict=True` markers for the deliberately deferred external-package lifecycle work, described in #143.

furgalep:
Closing while the TUI work continues — reopening once it is ready for review.

The branch `dev/tui-on-acp` stays on origin and is unaffected: it is rebased onto the reconciled #143 branch, all commits signed, and the full suite is green (7444 passed). Nothing is lost by closing this; it can be reopened or re-created against the same branch at any time.

PR #171, opened 2026-08-19
docs: propose workspace trust for repository-supplied skills

furgalep:
Design proposal, for discussion. Adds `docs/proposals/workspace-trust.md` — no code changes.

## Why

Creating a coding-agent session imports Python from the workspace **before any prompt is sent**, so opening a folder executes code it contains, as the user, in a process holding model credentials. Verified end to end rather than inferred:

```
skill roots discovered: ['<repo>/.claude/skills']
EXECUTED VIA new_session: True
```

Three paths do it — conventional skill roots, additional roots named by the repository's *own* settings (which are not confined to the workspace), and `.nooa/libs`. A fourth, `SKILL.md`, is already inert.

This surfaced during review of #143 and is documented there as accepted behaviour. This proposal is about whether we want to keep it that way, and is deliberately **not** part of that PR.

## The core idea

Gate on **provenance, not file type**. Gating on type is the obvious move and it breaks immediately: agents write their own skills, and autonomous hosts load skills programmatically with no human present.

- **Host-declared** (`skills_dirs=`, user-level config) — trusted, no prompt. *This is what autonomous and embedded hosts already use; nothing changes for them.*
- **Agent-authored** (written via `SkillWriting` this session) — trusted, no prompt.
- **Workspace-discovered** — needs a decision.

Classes 2 and 3 share a directory, since an ACP host points `libs_dir` at `<workspace>/.nooa/libs`. They are separated **temporally**: pre-existing content is not imported at session start, while `libs.create()` and `reload()` keep working during the session. Same directory, different provenance, no new machinery.

## Grounding

Verified against Claude Code's documented model rather than recalled. Relevant findings:

- Skills load as **text**; bundled scripts run only through a permission-gated tool. No documented automatic-execution path at discovery — so our `SKILL.md` handling is already aligned and the `.py` import path is the outlier.
- Claude Code already gates on this axis: external imports in a *project* `CLAUDE.md` are gated, the same imports in the user's own `~/.claude/CLAUDE.md` are not, the documented reason being that you wrote them.
- Trust is keyed to the **git repository root**, does **not** inherit into nested repositories, and is **session-only** outside a repository. All three adopted here; each is a case I would otherwise have got wrong.
- There is **no re-approval when a trusted project's hooks change**. That is the argument for content-hashed trust (`direnv`-style) in a later phase, and it is called out rather than glossed.

## Phasing

1. **Confine repository-supplied paths** to the workspace — no UX, no trust model, worth doing regardless.
2. Provenance split, defaulting to trust, logging what *would* be gated.
3. Flip the default, add the prompt (`session/request_permission` already exists in ACP).
4. Consider content-hashed trust.

## Open questions in the doc

Whether agent-authored skills should survive across sessions (the temporal rule says no, without an HMAC-keyed marker), whether the repository root is right for a monorepo, and what a declined prompt should surface.

MCP servers are explicitly out of scope — they arrive from the ACP client rather than the repository, and are a separate trust question.

### DIFF
--- a/docs/proposals/workspace-trust.md
+++ b/docs/proposals/workspace-trust.md
@@ -1,0 +1,179 @@
+<!-- SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
+# Proposal: workspace trust for repository-supplied skills
+
+**Status:** draft, for discussion
+**Affects:** `nooa.skill_registry`, `nooa_cli.coding`, `nooa-acp`
+
+## The problem
+
+Creating a coding-agent session **imports Python from the workspace, before any
+prompt is sent**. Opening a folder is therefore enough to execute code it
+contains, as the user, in a process holding model credentials.
+
+This was verified end to end rather than inferred. A marker-writing file placed
+at `<repo>/.claude/skills/probe.py`, followed by an ACP `session/new` on that
+directory:
+
+```
+skill roots discovered: ['<repo>/.claude/skills']
+EXECUTED VIA new_session: True
+```
+
+The only action was opening the folder.
+
+### Three paths execute
+
+| Path | Trigger | Executes? |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


furgalep:
Closing this documentation-only PR in favor of tracking the security concern, design proposal, phased implementation, and open questions in #231. The issue preserves the verified reproduction and links back here for the full proposal history. No code from this PR was merged.

PR #172, opened 2026-08-19
fix(tools): allow file operations outside shell cwd

furgalep:
## Summary
- remove the cwd containment check from `ShellTools` file path resolution
- continue resolving relative paths from the persistent shell cwd while accepting parent-relative and absolute paths
- document the expanded path forms in `ShellTools` and `ActivityShellTools`
- cover read, write, path replacement, and `Match` replacement outside the current directory

## Why
`ShellTools.run()` already allows `cd` to any accessible directory and updates the tool cwd. Rejecting equivalent paths in `read`, `replace`, and `write_file` was inconsistent and did not provide a meaningful boundary.

## Verification
- `uv run pytest -q tests/tools/test_shell_tools_modern_behavior.py packages/nooa-cli/tests/test_coding_activity.py` — 20 passed
- `uv run pytest -q tests/tools` — 275 passed
- `uv run pytest -q packages/nooa-cli/tests` — 970 passed, 2 unrelated reproducible macOS `/var` vs `/private/var` failures
- `uv run ruff check src/nooa/tools/shell_tools.py packages/nooa-cli/src/nooa_cli/coding/activity.py tests/tools/test_shell_tools_modern_behavior.py`
- `uv run ruff format --check src/nooa/tools/shell_tools.py packages/nooa-cli/src/nooa_cli/coding/activity.py tests/tools/test_shell_tools_modern_behavior.py`
- `git diff --check origin/dev/tui...HEAD`

## Summary by CodeRabbit

* **New Features**
  * File operations now support relative and absolute paths, including locations outside the current working directory.
  * Reading, writing, and content replacement work reliably with external file locations.
  * Matches remain associated with their original files when the working directory changes.

* **Documentation**
  * Updated path descriptions to clarify support for relative and absolute paths.

### DIFF
--- a/src/nooa/tools/shell_tools.py
+++ b/src/nooa/tools/shell_tools.py
@@ -72,11 +72,21 @@ class Match:
 
     Pass any ``Match`` to ``self.shell.replace(match, new_text)``. Print it to
     view numbered lines. Slice with ``match[start:end]`` (1-indexed inclusive
-    line numbers) to narrow the region before replacing.
+    line numbers) to narrow the region before replacing. ``resolved_path`` is
+    required so the anchor remains bound to that file if the shell cwd changes.
     """
 
-    def __init__(self, path: str, start: int, end: int, text: str):
+    def __init__(
+        self,
+        path: str,
+        start: int,
+        end: int,
+        text: str,
+        *,
+        resolved_path: str | Path,
+    ):
         self._path = path
+        self._resolved_path = str(Path(resolved_path).resolve())
         self._start = start
         self._end = end
         self._text = text
@@ -85,6 +95,11 @@ class Match:
     def path(self) -> str:
         """File path."""
         return self._path
+
+    @property
+    def resolved_path(self) -> str:
+        """Canonical file path captured when this match was created."""
+        return self._resolved_path
 
     @property
     def start(self) -> int:
@@ -129,7 +144,13 @@ class Match:
             idx_start = start - self._start
             idx_end = stop - self._start + 1
             text = "".join(lines[idx_start:idx_end])
-            return Match(self._path, start, stop, text)
+            return Match(
+                self._path,
+                start,
+                stop,
+                text,
+                resolved_path=self._resolved_path,
+            )
 
         raise TypeError(f"indices must be int or slice, not {type(key).__name__}")
 
@@ -350,14 +371,8 @@ class ShellTools(Skill):
         await self._session.close()
 
     def _resolve_path(self, path: str) -> Path:
-        """Resolve a file-operation path and require it to remain inside cwd."""
-        root = self.cwd.resolve()
-        resolved = (root / path).resolve()
-        try:
-            resolved.relative_to(root)
-        except ValueError as exc:
-            raise ValueError(f"path escapes ShellTools cwd: {path}") from exc
-        return resolved
+        """Resolve a file-operation path relative to cwd when not absolute."""
+        return (self.cwd / path).resolve()
 
     async def run(
         self,
@@ -501,7 +516,7 @@ class ShellTools(Skill):
             return None
 
         anchors: list[tuple[str, int]] = []
-        file_cache: dict[str, list[str]] = {}
+        file_cache: dict[str, tuple[Path, list[str]]] = {}
         for raw in rg_out.splitlines():
             raw = raw.strip()
             if not raw:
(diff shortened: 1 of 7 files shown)


furgalep:
@coderabbit full review

furgalep:
@coderabbit full review

src/nooa/tools/shell_tools.py
```diff
@@ -570,7 +585,15 @@ async def _harvest_matches(self, command: str, displayed_stdout: str) -> list[Ma
...
             if not (1 <= line_no <= len(lines)):
                 return None
-            out.append(Match(mpath, line_no, line_no, lines[line_no - 1]))
+            out.append(
+                Match(
+                    mpath,
+                    line_no,
+                    line_no,
+                    lines[line_no - 1],
+                    resolved_path=resolved,
```
alessiodevoto, line 601: This can attach the wrong `resolved_path` when search results contain the same `mpath` non-contiguously. `file_cache` currently stores only `lines`, so on a cache hit `resolved` is not recomputed and still refers to whichever file most recently missed the cache.

A reachable case is overlapping/repeated search paths, e.g. `grep -rn 'needle' dir/subdir dir`. `rg --json` can emit duplicate file results for overlapping inputs, and output order is not something we should rely on for correctness. If `keep` is ordered like `a.txt`, `b.txt`, `a.txt`, the second `a.txt` Match can be constructed with `resolved_path` pointing at `b.txt`; then `replace(match, ...)` would edit the wrong file.

The fix should be small: mirror the Cybergym copy in this PR and cache `(resolved, lines)` instead of just `lines`, then unpack both before constructing the Match.

alessiodevoto (approved):

src/nooa/tools/shell_tools.py
furgalep: Good catch — fixed in 19082f3. The core cache now stores and unpacks `(resolved, lines)`, matching the CyberGym implementation, so non-contiguous repeated paths cannot inherit another entry’s canonical path. I added a regression with `a.py, b.txt, a.py`; it fails against the previous implementation (the final `a.py` is incorrectly bound to `b.txt`) and passes with this fix. The full affected suite passes: 306 tests.

furgalep:
Reviewed with the specific question in mind — can a `Match` end up pointing at the wrong file once asyncio is involved. **Short answer: no, and the design is stronger than the tests suggest.** One latent hole, fixed in 99580ba.

## The concurrency case holds

Verified by running it rather than reading it:

- `read()` in `a/`, then `cd b/` where a same-named decoy exists, then `replace(match, ...)` → edits `a/f.txt`, decoy untouched.
- The same interleaved under `asyncio.gather`, with the replace strictly after the `cd` → still correct.
- `match[3:4]` slicing inherits the anchor rather than re-deriving it.

Structurally it is better than "handled": `read()`, `replace()` and the ripgrep match-parsing each resolve the path and then do **synchronous** I/O, with no `await` between resolution and capture. Those coroutines cannot yield mid-operation, so there is no interleaving window to lose a race in — not merely one that happens to be covered.

## The hole

`Match.__init__` did `Path(resolved_path).resolve()`, and `resolve()` on a **relative** path resolves against the *process* cwd, which is not the shell cwd:

```
process cwd    = /private/var/.../tmppa9iwj7y
intended file  = /var/.../tmppa9iwj7y/a/f.txt
match.resolved = /private/var/.../tmppa9iwj7y/f.txt     <- silently wrong
```

No error, no warning, and the eventual `replace()` writes to the wrong file — exactly the failure this PR exists to prevent, reachable through a different door.

It is **latent, not live**: every caller passes an absolute path today. `shell_tools` goes through `_resolve_path()`, `repo_tools` resolves its root at line 417 and joins from it, and the cybergym copy does the same. But it was a convention rather than a rule, and the kind that a future `Match(str(p), ..., resolved_path=p)` with a relative `p` breaks silently.

99580ba makes it a rule — `Match.__init__` raises if the anchor is not absolute — with tests for both the rejected and the supported form.

Worth noting one thing this turned up: `test_shell_result_timeout_flag_preserves_positional_matches_argument` was itself constructing a `Match` with `resolved_path="example.py"`, a relative anchor used as a stand-in value. Harmless there, but it is the pattern the guard forbids, so it now uses an absolute path.

## Verification

- `pytest tests/tools packages/nooa-cli/tests/test_coding_activity.py` → 308 passed
- `pytest packages/nooa-cli/tests tests/test_mcp` → 127 passed
- `ruff check` / `ruff format --check` clean

Issue #144, opened 2026-08-14
Durable orchestrators: a pure-Python run() cannot resume across a process boundary 🤖🤖🤖


shsagnik:
## What I was doing

Exercising the persistence API to see whether a NOOA agent could drive a long-running workflow that pauses for a human — an operator supplying a value partway through — and then finishes later, possibly in a different process.

Two things came out of that, one small and one structural. The small one I've already opened a PR for (#139, the `StorageManager` docstring documents `agent.save()` / `MyAgent.load(...)`, neither of which exists). This issue is about the structural one, which I'd rather ask about than assume.

## The gap

`AGENTS.md:32` is explicit that orchestrators are pure Python:

> **Orchestrators are pure Python.** Workflow sequence methods have real bodies (no `...`), calling generation methods for each step. If a class has no `...` methods it doesn't need to subclass `Agent` at all.

That works well, but a Python method's progress lives on the call stack, and nothing captures the stack:

- `AgentSnapshot.from_agent` walks `agent.__dict__` (`src/nooa/storage/snapshot.py:100`) — it captures *attributes*, not the frame, the local variables, or the program counter.
- `CodeActStrategy` runs a live in-process loop (`while not session.is_exhausted()`, `src/nooa/strategies/codeact.py:827`) with no suspension point.
- Grepping `src/nooa/` for `suspend` / `pause` / `resume_from` finds nothing.

So an orchestrator that is three steps in and waiting on a human cannot be resumed after the process exits. `InteractiveAgent` (`src/nooa/interactive.py:310`) does give durable progress between *turns*, but that's a chat-shaped loop — it doesn't help a `run()` method that is halfway down its own body.

## The workaround, and why I'm not sure it's the intended one

Keep the orchestrator's progress in a snapshotted field rather than in locals, so re-entering the method after `restore_snapshot()` skips whatever is already done:

```python
class PipelineAgent(Agent, llm=llm):
    steps_done: dict[str, str]      # snapshotted — this is the resumable part
    inputs: dict[str, str]

    async def run(self, payload: str) -> str:
        if "summary" not in self.steps_done:
            self.steps_done["summary"] = await self.summarise(payload)
        if "threshold" not in self.inputs:
            raise SuspendForInput("what risk threshold should apply?")
        ...
```

[...]

PR #174, opened 2026-08-20
Add durable operation coordination for SQLite 🤖🤖🤖

KashyapSinh-Gohil:
## What does this PR do?

This PR adds durable operation coordination to SQLite-backed workflows that may be interrupted, retried, or resumed across multiple workers.

The new ledger persists logical operation identifiers, request fingerprints, effect-safety classifications, leases, fencing generations, terminal states, outcome transitions, and operation history. Expired operations are reclaimed only when retrying is safe. Ambiguous external outcomes remain UNKNOWN until explicitly reconciled, and stale workers cannot commit after their leases expire.

Existing SQLite databases are migrated additively while preserving event, tag, and snapshot data. Recorded events are persisted before subscribers are notified. If persistence succeeds but the acknowledgement is lost, retrying the same logical operation ID reuses its existing durable identity instead of creating a duplicate.

The change adds tests for migration, corruption handling, leases, fencing, concurrency, multi-process execution, reconciliation, event ordering, and deterministic recovery schedules.

Full repository validation passes with 6,632 tests passed, 7 skipped, and 284 deselected.

This change does not claim universal exactly-once behavior for arbitrary external services. Ambiguous external outcomes require external reconciliation or idempotency support.

## Related issues

No related issue.

## Checklist

- [x] Code follows the project style; lint and formatting checks pass.
- [x] Tests were added or updated and are passing.
- [x] Documentation was updated for the new public behavior.
- [x] New source files include SPDX license headers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ to follow semantic versioning.
 ## [Unreleased]
 
 - Initial public release of NVIDIA Object-Oriented Agents (NOOA).
+- Added a SQLite durable-operation ledger with request identity, renewable leases, fencing, transition history, and explicit unknown outcomes.
+- Added additive SQLite schema migration, fail-closed operation decoding, authoritative database time, and explicit unknown-outcome reconciliation.
 - Security: MCP server configurations no longer expand host environment variables
   from `${VAR}` placeholders. Trusted caller code must resolve secrets and pass
   their values explicitly.
--- a/README.md
+++ b/README.md
@@ -195,6 +195,53 @@ uv run nooa start-dev        # trace viewer on http://localhost:5001
 
 If the viewer isn't running, tracing is silently disabled — no configuration needed either way.
 
+### 4. Coordinate resumable operations
+
+For workflows that cross a process boundary, use the durable operation ledger
+to separate execution ownership from the event history:
+
+```python
+from nooa.storage import EffectClass, ExecutionKey, SQLiteStorageManager
+
+storage = SQLiteStorageManager("agent.db")
+store = storage.execution_store
+key = ExecutionKey("workflow-1", "tool-call-1")
+claim = store.claim(key, {"customer_id": "42"}, effect_class=EffectClass.IDEMPOTENT)
+
+if claim.executable:
+    try:
+        result = call_idempotent_service(customer_id="42")
+        store.complete_success(claim, result)
+    except TimeoutError:
+        store.complete_unknown(claim, "service response was ambiguous")
+else:
+    result = claim.record.result
+```
+
+The ledger shares the SQLite storage manager's transaction lock and provides
+stable operation identity, request conflict detection, renewable leases,
+fencing against stale workers, append-only transition history, and terminal
+replay. Expired pure or idempotent work may be reclaimed. Other effect classes
+become `UNKNOWN` for explicit reconciliation instead of being run again.
+
+An `UNKNOWN` operation is not automatically retried. After querying the
+external system or applying a domain-specific reconciliation procedure, resolve
+it explicitly:
+
+```python
+store.reconcile_success(key, {"remote_id": "result-123"})
+# or, only after proving the effect did not happen:
+store.reconcile_failure(key, RuntimeError("remote transaction was rolled back"))
+```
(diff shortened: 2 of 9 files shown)


furgalep:
Thanks for the substantial work here — the operation ledger design is genuinely well-built. Stable `(workflow_id, operation_id)` identity, canonical request hashes, append-only transitions, leases with fencing generations, terminal replay, effect-class conflict checks, and reclaim restricted to pure/idempotent work are all the right primitives, and the 10,000 deterministic fault-schedule tests show real rigor.

Looking closely at the integration, though, I noticed the new `durable_execution.py` module imports **zero NOOA code** — it's entirely standard-library. It's effectively a standalone library placed inside the package, and the only real coupling is the shared SQLite connection in `sqlite.py` and the `EventManager` change. That started a design conversation on our side.

**The core question:** whether this belongs in the core framework depends on a decision we haven't made yet — whether NOOA will support durable execution end-to-end (#144 territory: workflow state machines, checkpoints, suspend/resume), à la Temporal. If we commit to that roadmap, this ledger is the correct first building block and should be re-landed against the architecture spec as step one. If we don't — or not yet — this is better as a loosely coupled external package, where a decorator API (e.g. `@durable_op(store, "name", effect_class=...)`) on top of your ledger would give agents the same capability without baking schema and lease semantics into the core before the orchestration contract is decided.

Since that architectural decision is still open, I'm closing this PR rather than leaving it (and you) parked on a large, conflicting branch — none of this is a judgment against the design itself, which is exactly why I'd like to suggest:

1. **The `EventManager` persist-before-notify change is independently worth having** — if you'd like, that part could be split into its own smaller PR and reviewed on its own merits.
2. **If you'd like to proceed with the ledger independently**, we'd be glad to help extract it as a standalone package with your authorship fully preserved — the module is already written dependency-free, so the move is nearly free, and the decorator pattern would make a very natural public API. Let us know, and we can link to your external package from the main readme.
3. If we later commit to the durable-execution roadmap, this work would be the natural first step to revisit.

For completeness, the things our review found for whenever this moves forward (whichever path): a recursive-callback dedup bypass in the new `_emit_once` (the ID is recorded after callbacks run, so a recursive `add()` of the same event re-enters — we reproduced 333 callbacks / 330 duplicate events), the `_notified_event_ids` set grows without bound since `collapse()` never purges it, and the branch is 26+ commits behind with README conflicts.

Thanks again for putting this together — the engineering quality is genuinely high.

KashyapSinh-Gohil:
Hi Paul, (@furgalep) 

Thanks for taking the time to review this. I appreciate your comments on the ledger, and I understand why you don’t want to add it to core while the broader decision in #144  is still open.

I missed both of those EventManager issues. The fault-schedule tests didn’t cover callbacks adding the same event again, or the notification-ID set growing over time. Those need their own regression tests before I reuse any of that code.

I’d like to take the smaller EventManager path you suggested first, from a fresh branch off current main. The change would be to save recorded events before calling their subscribers. If storage raises an error, those subscribers shouldn’t run. I’d test callbacks that re-submit the same event separately from callbacks that add a new one, and check that the bookkeeping gets cleaned up on errors rather than accumulating indefinitely.

I’d also check what moving the callbacks changes about the event they see and whether their mutations get saved. I don’t want to quietly change that behavior. Any differences would be called out in the tests and documentation, while keeping notification-only/runtime events and the existing handling of subscriber errors intact.

This still wouldn’t guarantee delivery a process could crash after saving the event but before calling a subscriber. I’d keep retries, replay, and the ledger out of this patch.

Before preparing it, I’d check for overlap with the current event and SQLite work. My CodeAct conformance and semantic-isolation work would stay separate.

I’m also interested in extracting the ledger into a standalone package. Thanks for offering to help with that and preserve my authorship. The decorator approach seems worth exploring as a small API, with clear limits around operation coordination, safe retries, and resolving uncertain outcomes not full workflow recovery. I’d treat that as a separate effort rather than bundle it into the EventManager follow-up.

Does this match what you had in mind for the smaller PR? Is there any existing same-event recursion or callback-mutation behavior you’d specifically want preserved?

I’ll wait for your reply before starting the implementation. Thanks again.

PR #176, opened 2026-08-20
feat(tui): deliver direct-agent fullscreen experience

furgalep:
## Summary

Deliver a complete fullscreen TUI experience built around a direct, Python-native agent interface. This MR replaces the transport-shaped attachment/session protocol, makes fullscreen the default display mode, hardens runtime and terminal ownership, and adds the interaction, retention, Unicode, resize, and lifecycle behavior needed for production use.

## Agent/runtime architecture

- introduce immutable UI-facing `AgentState` and a structural `InteractiveAgent` protocol
- use ordinary Python methods for submit, interrupt, pending-input withdrawal, and stop
- provide atomic, coalescing `observe()` subscriptions with explicit terminal-failure reporting
- replace the former attachment controller with transactional, generation-filtered `AgentController`
- fold local state and observation ownership directly into `LocalAgentRunner`
- make stopped-loop submit and cross-thread interrupt admission truthful and rollback-safe
- remove the old attachment/session/action/receipt/capability/event protocol and migrate all consumers; no compatibility adapter or dual protocol remains

## Fullscreen UI

- make `fullscreen` the resolved default while retaining explicit `native` and `native-replay` product modes
- give one prompt_toolkit application sole ownership of transcript, composer, queue, status, mouse, and resize rendering
- add anchored viewport navigation, keyboard and mouse selection, copy/cut, selection-aware editing, and multiline composer behavior
- complete transcript drags released over status chrome or the composer, and recover when tmux drops mouse-up outside the pane
- clear visual selection after drag-copy while preserving the captured payload
- preserve valid Unicode and grapheme boundaries in source/export text and use a projection-only one-cell ellipsis when a glyph cannot physically fit

## Correctness and performance

- bound live fullscreen retention to 10,000 records or a conservative 16 MiB resident-text budget, including replay/model/projection copies
- evict source/model/cache state transactionally and preserve viewport and record-local selection coordinates
- make append bookkeeping O(1)
- avoid scanning or rebuilding 10,000 static records on width changes
- coalesce width-sensitive semantic replays and cache the latest width result
- isolate all Session teardown phases, preserve the primary application exception, and guarantee terminal restoration even when cleanup phases fail
- harden callback generations, observation termination, queue/loop shutdown, cancellation, and late-output handling

## Documentation and tests

- document ownership, lifecycle, retention, projection, display-mode, and remote-proxy boundaries
- add deterministic concurrency, composition, PTY, lifecycle, teardown fault-injection, Unicode screen-cell, mouse, selection, and cap-scale resize coverage

## Validation

- focused direct-agent/runtime/fullscreen/session/startup suite: **340 passed**
- complete `packages/nooa-cli/tests`: **1,182 passed**, with 2 known unrelated macOS `/var` vs `/private/var` temporary-path failures
- Ruff check: passed for all branch-changed Python files
- Ruff format check: passed for all branch-changed Python files
- scoped Pyright for changed interactive/TUI production modules: **0 errors, 0 warnings**
- `git diff --check`: passed
- cap-scale static resize probe: 10,000 retained records, 1,000 observations in 0.000326 s total on the test machine
- branch consolidated to one DCO-signed commit

## Checklist

- [x] Code follows project style
- [x] Tests added or updated
- [x] Architecture documentation updated
- [x] No legacy agent protocol retained
- [x] Review input files and unrelated `uv.lock` change excluded
- [x] DCO sign-off present on the complete MR commit

## Summary by CodeRabbit

* **New Features**
  * Added native, native-replay, and fullscreen display modes, configurable through the CLI and settings.
  * Fullscreen mode now supports transcript scrolling, selection, Unicode-safe rendering, resizing, and alternate-screen restoration.
  * Added improved interactive agent controls for submitting, interrupting, switching, and stopping tasks.
  * Job, todo, memory, and status views now use consistent presentation data.
* **Bug Fixes**
  * Improved terminal safety and handling of closed output streams.
  * External editor requests in fullscreen mode now provide a clear explanatory message.
  * Improved cleanup and shutdown behavior during session transitions and failures.


sklinglernv:
@CodeRabbit review full

sklinglernv:
## Codex Findings

  - P1 — UI job projection can crash background jobs. packages/nooa-cli/src/nooa_cli/interactive/local_agent.py:637 eagerly calls
    str() on every buffered job value inside the queue notification callback. Because that callback executes synchronously from
    Channel.put(), a value with a failing __str__ propagates into the producer and terminates the spawned task. A focused
    reproduction also caused _on_done to fail. Repeatedly converting the complete buffer on every notification is additionally
    quadratic for long-running buffered streams. Snapshot formatting should be exception-isolated and bounded to a recent-value
    window.

  - P2 — Hot-swap leaves host services attached to the old agent. packages/nooa-cli/src/nooa_cli/tui/session.py:1138 updates the
    runner, then assigns self._app.agent, but the rewritten TUIApplication uses _agent and never reads this new attribute.
    Meanwhile todo/memory/replay/toolbar services still close over Session.self.agent, and LocalTurnPolicy retains its
    constructor agent. After slash-inception, those views and reflection status/teardown therefore target the original agent; the
    replacement reflection runner may survive shutdown. Maintain one transactionally updated current-agent reference and use it
    across all host services and lifecycle policy.

  - P2 — Prefix eviction shifts a surviving viewport anchor. packages/nooa-cli/src/nooa_cli/tui/fullscreen_transcript.py:422
    removes the synthetic joining newline from the first surviving record but restores old_anchor unchanged. Selection endpoints
    are adjusted above, while the viewport’s record-local source_offset is not. At one-column width, a focused reproduction
    showed the pinned top row jumping from b to c after eviction. Shift the surviving anchor by the stripped prefix length.

  - P3 — Clipboard tests depend on ambient SSH variables. packages/nooa-cli/tests/tui/test_fullscreen_transcript.py:1384 mocks
    the local clipboard path without clearing SSH_CONNECTION/SSH_TTY; production intentionally bypasses that path remotely. Nine
    tests consequently fail under SSH. Clear these variables for local-transport tests and test remote behavior separately

packages/nooa-cli/docs/tui-rendering-architecture.md
sklinglernv, line 1: I like to keep this as docs, but can we make it an actual human facing documentation?

sklinglernv (approved):
Nice!

As a wishlist item independent of this PR, could we add a human-readable quick-start guide for the TUI? Covering how to install it, how to use it, and a few basic tricks? I've completely lost track of all the features! :)

PR #177, opened 2026-08-20
fix: Python-API eval metadata fallback and memory README

### LINKED ISSUE #104: nooa-memory README cites examples/memory_bench/ for its measured results, but that directory is not in the repo 🤖🤖🤖
## Problem

`packages/nooa-memory/src/nooa_memory/README.md` cites `examples/memory_bench/` as the source of its benchmark results, but that directory is not in the repository. It is absent at every relative depth, so this is not a broken-path bug — there is nothing to point at.

Five references:

| Line | Reference | Claim made |
|---|---|---|
| 8 | `examples/memory_bench/` | "Runnable demos + measured results" |
| 303 | `examples/memory_bench/reflecting.py` (`make_llm_reasoner`) and `longmemeval.py` (`make_llm_reconciler`) | "for working implementations" |
| 400–408 | `examples/memory_bench/` | "Measured with real gpt-5.4 + text-embedding-3-large", followed by a results table |
| 449 | `examples/memory_bench/` | "for runnable benchmarks and measured behavior" |

The table at 400–408 publishes quantified results, e.g.:

```
| `longmemeval.py` | LongMemEval + reconsolidation | memory **+70%**, reflection neutral (+0%) |
```

So the README states a measured improvement, names the script that produced it and the model it was run against, and points readers at a directory they cannot open. A reader who wants to check the +70% figure, or copy `make_llm_reasoner` as the docs suggest, has nothing to go to.

## Verification

```
$ find . -type d -name memory_bench          # nothing
$ ls examples/
README.md  advanced  arc_agi_3  assets  benchmarks  cybergym
nooa-slides-prototype.ipynb  quickstart  util
```

What does exist for memory: `examples/quickstart/12_memory.py` and `examples/advanced/memory.py`. Neither is `memory_bench`, and neither contains `make_llm_reasoner` or `make_llm_reconciler`.

`examples/benchmarks/` exists but holds the Harbor bench-agent adapter (`bench_agent.py`, `harbor_adapter.py`, `harbor_minimal.yaml`), not memory benchmarks.

## Why it seems worth fixing rather than ignoring

`nooa-memory` is a published PyPI distribution, and this README is its main documentation. The reproducibility claim is the load-bearing part: "brain-inspired long-term memory" with a **+70%** number attached is exactly the kind of assertion a reader will want to run themselves before adopting it. Right now they cannot, and there is no note saying why.

I would guess `memory_bench/` exists internally and did not make it into the open-source export — the fact that the README names two specific modules and two specific factory functions inside them reads like documentation written against real code rather than aspirationally.

## Possible resolutions

Ordered by what I would guess is most likely, but this is a maintainer call and I do not want to guess in a PR:

1. **Publish `examples/memory_bench/`.** Best outcome if it exists and is publishable — the claims become checkable and the docs need no change.
2. **Point the references at what does exist** (`examples/quickstart/12_memory.py`, `examples/advanced/memory.py`) and mark the results table as measured with a harness that is not currently published, so the numbers stay but stop implying a directory the reader can open.
3. **Drop the references and the table** if the harness is not going to be published.

Happy to send the PR for (2) or (3) — both are small and docs-only. I have not opened one because picking between them decides whether published numbers keep an implied source, which is not mine to decide.

Separately, #103 fixes the one link in this file that *is* just a wrong relative depth (`12_memory.py`, which does exist). That one is independent of this.

🤖🤖🤖

### LINKED ISSUE #152: Evaluator.run() raises pydantic ValidationError under documented plain-Python API (_model_metadata never populated)
## Bug description

`Evaluator.run()` raises a `pydantic.ValidationError` when using the plain-Python API documented in `util/eval_pipeline/README.md` ("Option 1: Python API"), i.e. constructing `Evaluator(models={...})` directly rather than via `Evaluator.from_config(...)`.

## Repro

Following the README's own example almost verbatim:

```python
from eval_pipeline import Evaluator, ExactMatchScorer
from nooa.unifiedllm import CompletionClient

evaluator = Evaluator(
    models={"gpt-4": CompletionClient(model="openai/gpt-4", api_key="...")},
    output_dir="experiments/results",
)
evaluator.add_test(
    name="sentiment",
    agent_class=SentimentAgent,
    method="classify_single",
    data=[{"kwargs": {"text": "I love this!"}, "expected": "positive"}],
    scorers=[ExactMatchScorer()],
)
results = await evaluator.run(models=["gpt-4"])
```

Raises:

```
pydantic_core._pydantic_core.ValidationError: 2 validation errors for EvalMetadata
models.0.ModelSpec.model_name
  Field required [type=missing, input_value={'id': 'gpt-4'}, input_type=dict]
models.0.str
  Input should be a valid string [type=string_type, input_value={'id': 'gpt-4'}, input_type=dict]
```

## Root cause

In `src/eval_pipeline/evaluator.py`, `Evaluator.run()` builds the run's metadata line from `self._model_metadata`:

```python
writer.start(
    ...,
    models=[self._model_metadata.get(m, {"id": m}) for m in model_ids],
    ...
)
```

`self._model_metadata` is initialized to `{}` in `__init__` with the comment "Optional metadata (set by config_adapter)" -- it's only ever populated by the YAML/`from_config` path. When `Evaluator` is constructed directly with a plain `models={...}` dict (the documented, "Recommended for flexibility" Python API), `_model_metadata` stays empty, so every model ID falls back to `{"id": m}` -- a dict that satisfies neither `ModelSpec` (missing the required `model_name` field) nor the `str` arm of `EvalMetadata.models: list[ModelSpec | str]`.

## Environment

- Pinned commit: `8622fc4d8744377e73a54521af8abd1ac36fb201`
- `eval_pipeline` version: `0.2.0`
- Installed via `pip install "eval_pipeline @ git+https://github.com/NVIDIA-NeMo/labs-OO-Agents.git@8622fc4#subdirectory=util/eval_pipeline"`
- `nooa==0.0.8` (PyPI)

## Workaround

Manually pre-populate `_model_metadata` before calling `run()`:

```python
evaluator._model_metadata = {m: {"id": m, "model_name": m} for m in model_ids}
```

## Suggested fix

Rather than defaulting the missing entry to `{"id": m}` (which can't satisfy `ModelSpec` or `str` either way) or making `model_name` a hard requirement (which would still leave the README's own example broken, since a plain-Python `Evaluator(models={...})` caller has no other place to supply it), `Evaluator.run()` already has everything it needs to build a complete `ModelSpec` on its own: `self.models[m]` is the actual client object, and every `UnifiedLLM` subclass (so `CompletionClient`/`ResponsesClient`) sets `self.model` in `__init__` (`unifiedllm.py` line ~1157). So the fallback in `run()` could be:

```python
self._model_metadata.get(m, {"id": m, "model_name": getattr(self.models[m], "model", m)})
```

This fixes the crash *and* produces more accurate metadata than a bare `{"id": m, "model_name": m}` guess would (`model_name` ends up as the real slug, e.g. `"openai/gpt-4"`, not the short dict key `"gpt-4"`). Falling back to `m` only applies if `self.models[m]` is some client type that doesn't expose `.model` (i.e. no worse than today's crash, but only for that edge case rather than the common path).

Happy to send a PR for this if useful -- it's a small, localized fix at `evaluator.py`'s `writer.start(...)` call in `run()`.

skundu42:
## What does this PR do?

Two small fixes:

**Eval Python API (`#152`).** `Evaluator.run()` crashed with a Pydantic `ValidationError` when constructed the way the README documents — `Evaluator(models={...})` plus `await evaluator.run()` — instead of via YAML/`from_config`. `writer.start()` fell back to `{"id": m}`, which is neither a valid `ModelSpec` (missing required `model_name`) nor a string. The fallback is now the model id string, which `EvalMetadata.models: list[ModelSpec | str]` already accepts. The YAML path still writes full `ModelSpec` dicts. Adds a regression test that runs the documented API and checks the JSONL metadata line.

**Memory README (`#104`).** `packages/nooa-memory/src/nooa_memory/README.md` pointed at `examples/memory_bench/` in four places. That directory is not in the public repo, so “runnable demos + measured results” 404’d. Links now go to `examples/quickstart/12_memory.py`. The measured-results table is kept, labeled as internal (harness not published). Adds tests that the README does not cite `memory_bench` and that every relative markdown link resolves on disk.

## Related issues

Closes #152
Closes #104

## Checklist

- [x] Docs updated if behaviour or public APIs changed

### DIFF
--- a/packages/nooa-memory/src/nooa_memory/README.md
+++ b/packages/nooa-memory/src/nooa_memory/README.md
@@ -5,8 +5,7 @@ Agents. Toggle it on per agent; it changes no core framework code. Its defining
 idea: **the agent authors and curates its own memories** through native tools —
 rather than a harness extracting them behind the agent's back.
 
-- Runnable demos + measured results: [`examples/memory_bench/`](../../../examples/memory_bench/)
-  and the quickstart [`examples/quickstart/12_memory.py`](../../../../examples/quickstart/12_memory.py).
+- Runnable demo: [`examples/quickstart/12_memory.py`](../../../../examples/quickstart/12_memory.py).
 
 ---
 
@@ -300,8 +299,8 @@ def my_llm_reconciler(cluster):                 # list[Memory] (old->new) -> (cu
     ...
 ```
 
-See `examples/memory_bench/reflecting.py` (`make_llm_reasoner`) and `longmemeval.py`
-(`make_llm_reconciler`) for working implementations.
+The stubs above are the contract. [`examples/quickstart/12_memory.py`](../../../../examples/quickstart/12_memory.py)
+shows the deterministic (no-LLM) reflection path.
 
 **When does reflection help?** Empirically (real gpt-5.4):
 
@@ -397,8 +396,9 @@ agent.event_manager.on("MemoryWritten", lambda e: print("wrote", e.memory_id, e.
 
 ## When does memory help?
 
-Measured with real gpt-5.4 + text-embedding-3-large (see
-[`examples/memory_bench/`](../../../examples/memory_bench/)):
+Measured internally with gpt-5.4 + text-embedding-3-large (the bench harness is
+not published in this repository). For a runnable offline demo see
+[`examples/quickstart/12_memory.py`](../../../../examples/quickstart/12_memory.py):
 
 | demo | what it isolates | result |
 |---|---|---|
@@ -446,5 +446,5 @@ Tests: [`tests/memory/`](../../tests/memory/).
 - `numpy`/`sqlite-vec` backends are exact/brute-force — fine to ~10⁵ vectors; beyond
   that prefer a Chroma backend.
 
-See [`examples/memory_bench/`](../../../examples/memory_bench/) for runnable
-benchmarks and measured behavior.
+See [`examples/quickstart/12_memory.py`](../../../../examples/quickstart/12_memory.py)
+for a runnable demo.
--- a/packages/nooa-memory/tests/memory/test_memory_readme.py
+++ b/packages/nooa-memory/tests/memory/test_memory_readme.py
@@ -1,0 +1,34 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""The packaged README must not point readers at files that are not in the repo."""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+_README = Path(__file__).resolve().parents[2] / "src" / "nooa_memory" / "README.md"
+_MARKDOWN_LINK = re.compile(r"\[[^\]]*\]\(([^)]+)\)")
+
+
+def test_readme_does_not_cite_missing_memory_bench():
+    """Issue #104: examples/memory_bench/ is not in the public repository."""
+    text = _README.read_text()
+    assert "memory_bench" not in text, (
+        "README cites examples/memory_bench/, which is not in this repository. "
+        "Point at examples/quickstart/12_memory.py instead."
+    )
+
+
+def test_readme_relative_markdown_links_resolve():
+    """Every relative markdown link from the README must exist on disk."""
+    text = _README.read_text()
+    missing: list[str] = []
+    for href in _MARKDOWN_LINK.findall(text):
+        target = href.split("#", 1)[0].strip()
+        if not target or target.startswith(("http://", "https://", "mailto:")):
+            continue
+        resolved = (_README.parent / target).resolve()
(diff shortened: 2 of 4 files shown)


furgalep:
Fresh local validation of the added regressions passes (`3 passed`: both README-link tests plus the documented Python-API evaluator test), and Ruff lint passes. Two current blockers remain: DCO is `ACTION_REQUIRED`, and `uv run ruff format --check packages/nooa-memory/tests/memory/test_memory_readme.py` reports that file would be reformatted (the `_README` assignment). Please sign off/amend the commit and run Ruff format before review.

skundu42:
Thanks for the comment. Apologies for missing out the signoff. Everything should be good now.

furgalep (approved):
OK to merge

PR #180, opened 2026-08-20
fix(arc_agi_3): exit non-zero when the harness produces no result.json

ShichengRao:
## What does this PR do?

`run_solver.main()` prints an error when the harness exits without writing `result.json` — and then falls through to `return 0`. Any wrapper or CI job reading the exit code sees success for a run that produced nothing. (We hit this integrating `run_solver` into an offline evaluation rig: a fully failed run looked successful to our automation until we checked artifacts.)

This returns 1 on that path so the printed failure and the exit status agree. The success path is unchanged, so `test_offline_e2e`'s `returncode == 0` assertion still holds — and now actually discriminates.

## Related issues

None filed; happy to open one if you prefer tracking it that way.

## Checklist

- [x] One-line change; `python -m ast` parses the file (ruff/pytest not run locally — no formatting-sensitive lines touched)
- [ ] Tests added/updated (none — existing e2e covers the success path; the failure path has no harness-free test hook)
- [x] No public API or docs impact beyond the corrected exit status
- [x] No new source files

### DIFF
--- a/examples/arc_agi_3/run_solver.py
+++ b/examples/arc_agi_3/run_solver.py
@@ -562,6 +562,7 @@ def main() -> int:
             "[run] harness exited without result.json — check "
             f"{run_dir}/harness.log and {run_dir}/launcher.err"
         )
+        return 1
 
     return 0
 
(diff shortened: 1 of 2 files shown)


esarafian (approved):
The production fix remains a single return 1 after the existing cleanup and durable-result check. Added a credential-free, normally collected regression covering missing/valid result.json with harness exit 0/1: all four cases pass; the original source fails both missing-result cases. Valid-result behavior, artifact copy-back, and cleanup are preserved. Ruff lint and format checks pass. Fork CI is approved and running; merge will wait for successful checks. 🤖🤖🤖

PR #181, opened 2026-08-20
fix(arc_agi_3): attach skip guards to the test, not a helper

ShichengRao:
## What does this PR do?

Both `skipif` markers in `tests/test_offline_e2e.py` (SDK-present and credentials-present) decorate `_traces_contain`, a plain helper, where pytest markers have no effect. The actual test, `test_offline_run_makes_progress`, sits below them unguarded — so running the suite without `ARC_LLM_API_KEY` / `ARC_API_KEY` fails the test on a missing artifact instead of skipping with the intended reason string.

This moves both markers onto the test function. Verbatim move, no other changes.

## Related issues

None filed; happy to open one if you prefer tracking it that way.

## Checklist

- [x] Marker move only; `python -m ast` parses the file (ruff/pytest not run locally — lines moved verbatim)
- [x] Tests: this PR is itself the test fix — without credentials the suite now skips with the documented reason instead of erroring
- [x] No docs or API impact
- [x] No new source files

### DIFF
--- a/examples/arc_agi_3/tests/test_offline_e2e.py
+++ b/examples/arc_agi_3/tests/test_offline_e2e.py
@@ -43,11 +43,6 @@ def _have_sdk() -> bool:
         return False
 
 
-@pytest.mark.skipif(not _have_sdk(), reason="ARC-AGI-3 SDK not installed (`uv sync --extra arc`)")
-@pytest.mark.skipif(
-    not os.environ.get("ARC_LLM_API_KEY") or not os.environ.get("ARC_API_KEY"),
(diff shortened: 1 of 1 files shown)


esarafian (approved):
The same two skip conditions now decorate the collected integration test; the helper and test bodies are unchanged. Verified an explicit credential-free integration invocation skips as intended. A mocked 36-check prerequisite/selection matrix confirms all 15 incomplete prerequisite combinations skip, while the complete combination still reaches the launch boundary. No live services were called. Ruff lint and format checks pass. Fork CI is approved and running; merge will wait for successful checks. 🤖🤖🤖

PR #184, opened 2026-08-21
feat(tui): add copy actions to Markdown code blocks

furgalep:
## Summary

- render a clickable Copy action above non-empty fenced Markdown code blocks in fullscreen mode
- retain exact source code as semantic metadata instead of reconstructing it from wrapped or syntax-highlighted terminal text
- preserve copy actions across resize replay and transcript retention eviction
- route clicks through the existing non-blocking clipboard path with standard success/failure feedback
- isolate native display modes and protect the internal action scheme from model-authored link spoofing

## Validation

- `uv run --project packages/nooa-cli pytest -q packages/nooa-cli/tests/tui/test_fullscreen_transcript.py packages/nooa-cli/tests/tui/test_agent_event_renderer.py` (142 passed)
- full `packages/nooa-cli/tests/tui` suite: 1106 passed; only the two known unrelated macOS `/var` vs `/private/var` path-alias failures
- Ruff check and format check
- scoped Pyright: 0 errors, 0 warnings
- `git diff --check`
- independent review: no blocker, high, or medium findings

### DIFF
--- a/packages/nooa-cli/src/nooa_cli/tui/copyable_markdown.py
+++ b/packages/nooa-cli/src/nooa_cli/tui/copyable_markdown.py
@@ -1,0 +1,91 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Markdown rendering with semantic copy affordances for fenced code blocks."""
+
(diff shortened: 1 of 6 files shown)


furgalep:
Superseded by #198, which now combines this source-backed code Copy action with the transcript prose-copy fix while retaining Rich's highlighted/indented code-block presentation.

PR #186, opened 2026-08-21
Hide skill lifecycle hooks from agent docs

furgalep:
## Summary
- mark the base `Skill.attach()` and `Skill.detach()` lifecycle hooks with the existing `@hidden` decorator
- verify inherited lifecycle hooks stay out of model-facing agent docs
- avoid adding any new agentdoc visibility mechanism or changing `spec(..., hidden=False)` semantics

## Verification
- `uv run ruff check src/nooa/skill.py tests/agentdoc/test_skill_lifecycle_visibility.py`
- `uv run pytest tests/agentdoc -q` (687 passed)


sklinglernv (approved):
LGTM!

PR #189, opened 2026-08-21
fix(agents): preserve Python error source context

### LINKED ISSUE #191: Sandbox backend drops traceback context from agent-facing errors: worker_traceback is captured but never consumed
### Summary

When `execution_backend="sandbox"`, the error text shown to the LLM after a failed cell contains only the exception type and message. The frame list, line numbers, source lines, and caret markers that the `inprocess` backend provides are all missing.

The worker's traceback *is* captured and carried across the process boundary, but nothing ever reads it.

### Environment

- Commit `97f52de`
- Python 3.12.3
- Ubuntu (WSL2)
- Baseline `uv run pytest -m sandbox -q`: 48 passed, 1 skipped

### Reproduction

    import asyncio

    from nooa import Agent
    from nooa.errors.formatting import format_error_for_llm
    from nooa.unifiedllm import FakeLLMClient

    CODE = "a = 1\nb = 2\nc = 3\nd = 4\ne = 5\nboom = None.strip()\n"

    class T(Agent, llm=FakeLLMClient()):
        pass

    async def run(use_sandbox: bool) -> str:
        agent = T()
        kw, ex = {}, None
        if use_sandbox:
            ex = SandboxedExecutor(
                agent,
                SandboxConfig(require=False, network=True, filesystem=False),
                cell_timeout=15.0,
            )
            kw["sandbox_executor"] = ex
        try:
            res = await agent.runtime.execute_code(
                CODE, execution_count=1, wrap_in_function=True, **kw
            )
            return format_error_for_llm(res.error, CODE, line_offset=res.wrapper_line_offset)
        finally:
            if ex is not None:
                await ex.aclose()

    async def main() -> None:
        for label, sb in (("inprocess", False), ("sandbox", True)):
            print(f"--- execution_backend={label} ---")
            print(await run(sb))
            print()

    asyncio.run(main())

### Actual output

    --- execution_backend=inprocess ---
      Cell In[1], line 6, in <module>
        boom = None.strip()
               ^^^^^^^^^^

    --- execution_backend=sandbox ---

### Expected

Both backends render equivalent guidance. `runtime/sandbox/serialization.py` states this is the intent:

> Preserves the original type name and formatted traceback so the CodeAct error formatter renders the same guidance the in-process path would.

### Cause

`serialization.py` captures the worker traceback (line 111) and reattaches it to the reconstructed exception (line 186) as `worker_traceback`. But that attribute has two write sites and no read sites:

    $ grep -rn "worker_traceback" src/ packages/ tests/ --include=*.py
    src/nooa/runtime/sandbox/serialization.py:77
    src/nooa/runtime/sandbox/serialization.py:186

`errors/formatting.py` derives frames solely from `error.__traceback__` (lines 113, 365, 369). The reconstructed surrogate has no real `__traceback__` — tracebacks aren't picklable — so the frame-rendering path is skipped and only the final line is emitted.

### Scope

Not limited to `AttributeError`. The same divergence appears with `ZeroDivisionError`, `NameError`, multi-frame errors through a nested `def` (in-process shows both frames, sandbox shows none), and cells containing blank lines or semicolon-compound statements.

Note that `wrapper_line_offset` is identical (2) on both paths, so the offset arithmetic is not implicated — only the frame rendering.

### Impact

Agents running the sandbox backend receive error feedback with no location information, which weakens the error-recovery loop precisely in the configuration intended for untrusted code.

### Question before a PR

Two possible directions, and I'd rather follow your preference than guess:

1. Extend `ErrorDTO` to carry structured frames (e.g. `traceback.extract_tb` output) so the existing renderer works unchanged on both paths.
2. Teach `format_error_for_llm` a string-traceback path that consumes `worker_traceback`.

Option 2 is smaller but needs its own offset handling, since the captured string already includes the `__repl_wrapper__` frames. Option 1 is a wider change but keeps a single renderer.

Happy to take this either way. I'm also interested in #187, and a shared conformance suite across backends is what led me here.

furgalep:
## Overview

This PR makes Python execution failures useful to the agent by preserving source-aware diagnostics from the point where an exception still has its traceback through event rendering and retry feedback. It applies to both in-process and process-sandbox execution; it does not change the TUI.

### Source-aware diagnostics

- Render tracebacks while source and traceback state are still available.
- Preserve chained exceptions, exception groups, syntax locations, multiline spans, tabs, wide characters, and combining marks.
- Keep persisted helper source in `linecache` so later failures point to the original cell.
- Bound malformed, hostile, and oversized diagnostics safely.

### Typed sandbox boundary

Sandboxed user-code failures now cross the process boundary as `SandboxExecutionError`, carrying:

- `original_type` — the worker exception type name;
- `original_error` — a safely reconstructed exception; and
- `diagnostic` — the worker-rendered source-aware traceback.

This avoids adding backend-specific fields to `ExecutionResult`. Strategy telemetry continues to report the original exception type.

### Deterministic return transport

Return values and `return_result(...)` payloads are serialized exactly once before transport. Non-picklable values become `CellSerializationError` results instead of killing the worker, while `return_result(None)` remains distinguishable from a missing payload.

### Unified execution events and cell identity

CodeAct and PurePython record failed executions as `PythonOutput` events with separate stdout, stderr, formatted error, status, and metadata. Prefill and generated cells receive stable, distinct `Cell In[N]` identities. `EventsApi` documents the distinction between failed execution outputs and standalone strategy/retry `Error` events and supports filtering by execution status.

### Formatter contract

Custom CodeAct formatters implement:

```python
def format(
    error,
    code=None,
    *,
    line_offset=0,
    max_error=None,
    tail_chars=None,
) -> str:
    ...
```

This is an intentional pre-1.0 breaking change; reduced legacy signatures are no longer adapted at runtime.

### CI coverage

The workflow uses read-only token permissions and adds dedicated Python 3.13 diagnostic and macOS sandbox IPC coverage.

## Scope

The SQLite/event-query iteration optimization discovered during review was split into #200. This PR contains no SQLite storage changes.

## Validation

- Full local suite: **6872 passed, 7 skipped, 298 deselected, 3 xfailed**
- Focused error/sandbox suite: **630 passed, 1 skipped, 45 deselected**
- Ruff and scoped Pyright passed
- `git diff --check` passed
- Latest pushed-head CI is tracked below

All review threads have been addressed and resolved.

Closes #186.
Closes #191.


furgalep:
@coderabbit full review

furgalep:
Addressed all comments from the latest review in signed commit `0c914eb`:

- restricted the two new CI jobs to `contents: read` and disabled checkout credential persistence;
- restored `linecache.cache` state in both runtime-marker tests;
- documented the `formatted_error` trust boundary and corrected sandbox serialization docs;
- retained the `execution_error` marker when prefill metadata is supplied;
- corrected the spoofing regression to exercise `_nooa_call_hint` and added explicit transported-diagnostic coverage.

I also verified the two unverified observations. `_run_one` captures the formatter in its default argument and is the only worker invocation path. PurePython emits `PythonOutput.execution_count` from `session.execution_count`, and `OutAccessor` indexes those emitted events by the same field, so there is no iteration/count mismatch to change.

Local validation: Ruff format/check, 413 focused tests, scoped Pyright (0 errors), workflow YAML and permission assertions, and `git diff --check`.

furgalep:
@coderabbit full review

furgalep:
Addressed the two fresh CodeRabbit comments in signed commit `3476b46`:

- clarified the agent-facing `EventsApi` docs: standalone `Error` events and failed `PythonOutput` events require different queries;
- extracted the four repeated sandbox failed-output lookups into one typed `_failed_output` helper.

Validation passed: Ruff format/check, 6 real sandbox integration tests, 17 EventsApi/query tests, scoped Pyright (0 errors), and `git diff --check`.

@coderabbitai full review

furgalep:
Addressed the remaining CodeRabbit nit in signed commit `abc1716`: the prefill metadata regression test now first asserts that synthetic events were emitted, so its `all(...)` assertion cannot pass vacuously.

Validation passed: all 203 tests in `test_codeact_pure_python_coverage.py`, Ruff format/check, and `git diff --check`.

@coderabbitai full review

furgalep:
Addressed both findings from the latest review in signed commit `633405b`:

- brokered control-flow signals whose payload cannot cross the process boundary now become a clear `CellSerializationError` instead of silently completing with `{}`; added a real process-boundary regression that also verifies worker recovery;
- all sandbox CodeAct failure assertions now select the exact `execute_python` tool-call ID.

Validation passed: Ruff, 37 sandbox tests (3 platform skips), scoped Pyright with zero errors, and `git diff --check`.

@coderabbitai full review

furgalep:
Addressed the latest review findings in signed commit `edad840`:

- CodeAct now assigns a distinct execution count to every real cell, including multiple `execute_python` calls in one model response. Synthetic-comment cells participate in the same sequence, preventing duplicate `Out[n]` keys.
- The existing CodeActLite recovery test now explicitly pins the first failure to execution count 1 and retains its `Cell In[1]` assertion.
- Failed CodeAct prefill `PythonOutput` events now carry `execution_error: true`; successful prefill events remain unmarked. Focused assertions cover both cases.
- Sandbox error-transport bounds now derive from the shared truncation configuration, including an explicit `None` check so a valid zero-tail setting remains correct.

Validation passed on Python 3.12 (535 affected tests plus 6 real sandbox tests), Python 3.13 (726 affected tests), Ruff, scoped Pyright (0 errors), and `git diff --check`.

@coderabbitai full review

furgalep:
Addressed both findings from the latest CodeRabbit review in signed commit `454b30f`:

- each CodeAct prefill cell now consumes its own session execution count, so execution, diagnostics, emitted `PythonOutput`, `Out[n]`, and the following model cell all use an unambiguous sequence;
- sandbox broker diagnostics now derive their bound from `DEFAULT_TRUNCATION_CONFIG.capture.max_error` instead of a separate literal.

Added focused regressions for successful/error prefill counts and two-prefill sequencing. PurePython's established prefill count remains unchanged.

Validation passed: Ruff format/check, 536 focused tests on Python 3.12 and Python 3.13, 37 real sandbox tests (3 platform skips), scoped Pyright with zero errors, and `git diff --check`.

@coderabbitai full review

furgalep:
Implemented the formatter contract and resolved truncation-policy follow-up in commit `eaab639`:

- preserved trusted worker-rendered diagnostics while enforcing caller and IPC bounds without nested truncation envelopes;
- threaded resolved `max_error` and `tail` through CodeAct, PurePython, sandbox executor, worker, and serialization;
- added a public formatter protocol while retaining legacy custom formatter signatures;
- hardened worker formatting against module-level formatter mutation and fixed tiny broker-limit handling;
- added focused unit and real-process sandbox regressions.

Independent final review identified two budget-bypass paths; both are fixed with regressions. Local validation passed: Ruff, scoped Pyright (0 errors), 326 focused Python 3.12 tests, 511 expanded Python 3.13 tests, 48 real-sandbox tests (10 platform skips), and `git diff --check`.

Broader architectural cleanup remains tracked in #190.

@coderabbitai full review

furgalep:
Addressed the remaining reduced-budget finding in `4fa0325`: valid pre-truncated envelopes are now parsed and preserved only when their retained head/tail fit the active policy; otherwise they are rebuilt to the smaller budget without nesting. Both public formatting and sandbox DTO paths have regressions.

Validation: Ruff and scoped Pyright pass; 307 focused Python 3.12 tests, 513 expanded Python 3.13 tests, and 48 real-sandbox tests pass; all 8 CI checks are green.

@coderabbitai full review

furgalep:
Addressed all verified findings from the latest review in `ec197d1`:

- distinguish legitimate `None` signal payloads from dropped non-picklable payloads;
- defensively normalize invalid sandbox error limits;
- bound PurePython syntax guidance as part of the final diagnostic;
- centralize hard text caps and preserve closing truncation markers;
- hard-cap rebuilt envelopes even with adversarially large numeric metadata;
- apply the requested focused test maintenance.

Validation passed: Ruff, scoped Pyright (0 errors), 333 focused Python 3.12 tests, and 52 real-sandbox Python 3.13 tests (10 platform skips). Independent follow-up review’s remaining medium finding is also fixed with a regression.

@coderabbitai full review

furgalep:
Addressed the two remaining actionable findings in `e5b9cbb`:

- invalid `tail_chars` fallback is now clamped below the active `max_error`;
- sandbox serialization now supports custom formatters whose only accepted call shape is `format(error)`.

Both paths have focused regressions. Local validation passed: Ruff, scoped Pyright (0 errors), 108 focused Python 3.12 formatter/serialization tests, and 520 expanded Python 3.13 tests. Independent review found no blocker/high/medium issues.

@coderabbitai full review

furgalep:
Addressed the final sandbox-tail normalization finding in `2642b82`. Raw sandbox error messages now use the same defensive tail-policy normalization as formatted diagnostics, rejecting boolean/negative values and falling back to a valid split. Added regressions for `tail_chars=-1` and `tail_chars=True`.

Validation passed: Ruff, scoped Pyright (0 errors), and 316 focused tests.

@coderabbitai full review

furgalep:
Follow-up `2642b82` addresses the remaining CodeRabbit finding: sandbox raw-error serialization now normalizes invalid `tail_chars` with the same policy as formatted diagnostics. Regressions cover negative and boolean values.

Local verification: Ruff, scoped Pyright (0 errors), 316 focused tests, and `git diff --check`.

@coderabbitai full review

furgalep:
Addressed the remaining custom-formatter compatibility finding in `0b261e4`: CodeAct now preserves worker `formatted_error` for formatters that support that keyword but not `line_offset`, including variants that accept the resolved budget. Added a focused regression.

Validation passed: Ruff, scoped Pyright (0 errors), 338 focused Python 3.12 tests, 523 expanded Python 3.13 tests, and `git diff --check`.

@coderabbitai full review

furgalep:
Addressed the remaining formatter-dispatch finding in `148f7f8` and follow-up `050887f`: both CodeAct and sandbox serialization now filter optional arguments from the inspected signature instead of relying on fixed combinations, while CodeAct still preserves the legacy source-code argument. Added regressions for partial keyword subsets and positional optional parameters.

Validation passed: Ruff, scoped Pyright (0 errors), 242 focused tests, and `git diff --check`. Independent review caught the positional ambiguity before the final follow-up, and that is now covered.

@coderabbitai full review

furgalep:
Implemented option 1 in `0d2affb`: custom CodeAct error formatters now have one required complete protocol, and the legacy signature-inspection/argument-filtering paths have been removed. The public contract and changelog now explicitly require `format(error, code=None, *, line_offset=0, formatted_error="", max_error=None, tail_chars=None)`. Internal sandbox formatter injection follows the same complete callable contract.

Validation passed on Python 3.12 and 3.13: 359 affected tests passed (62 sandbox tests skipped by platform/config), Ruff passed, scoped Pyright reported 0 errors, and `git diff --check` passed. An independent review reported no actionable findings.

@coderabbitai full review

furgalep:
Starting another independent review round on current head `0d2affb`.

@coderabbitai full review

furgalep:
Addressed every actionable finding from the additional review round in `f199bdd`:

- serialized return/signal payloads exactly once before DTO transport, while retaining legacy DTO decoding;
- rejected non-string formatter results with the safe bounded fallback;
- scoped wrapper line-offset adjustment to the current cell and preserved original source mappings for helpers defined in earlier cells;
- covered wrapped `SystemExit` cause chains;
- aligned the formatter-interface test with real CodeAct delegation; and
- added least-privilege workflow permissions.

Validation passed on Python 3.12 and 3.13, including Ruff, scoped Pyright, and the affected suites. A final independent review found no actionable findings.

@coderabbitai full review

src/nooa/errors/formatting.py
furgalep: Await shell doesn't math the signature of shell tools. And shell tools suggests another solution via args. However I'm not that excited about direct coupling of error messages and shell tools. 

We should update these examples.

src/nooa/runtime/events.py
```diff
@@ -45,9 +45,16 @@ class EventsApi(Skill):
         children = events[summary.children_tags]
 
     Examples:
-        # Find the last error
+        # Find the last standalone Error event
```
furgalep: What creates standalone error events? If nothing, we should remove from this example.

src/nooa/runtime/sandbox/serialization.py
```diff
@@ -19,16 +19,82 @@
...
+def _bounded_formatted_error(
+    value: str,
+    *,
+    max_error: int | None = None,
+    tail_chars: int | None = None,
+) -> str:
+    """Apply capture policy once and enforce a fixed IPC ceiling."""
+    content_limit = effective_error_limit(max_error)
+    bounded = _bound_preformatted_diagnostic(value, content_limit, tail_chars)
+    return _hard_bound_text(bounded, min(_MAX_ERROR_TRANSPORT, content_limit + 1_024))
```
furgalep: Why is this needed?

src/nooa/runtime/events.py
```diff
@@ -45,9 +45,16 @@ class EventsApi(Skill):
...
         errors = events.query(type="Error", limit=1)
 
+        # Execution failures are PythonOutput events, not Error events
+        failed_outputs = [
+            output
+            for output in events.query(type="PythonOutput")
+            if output.execution_status.value == "error"
+        ]
```
furgalep: Doesn't query have a better way to do this? This could be a very long operation for an agent with a long history.

src/nooa/events.py
```diff
@@ -663,6 +663,11 @@ class ExecutionResult(BaseModel):
     stdout: str = Field(default="", description="Captured stdout from execution")
     stderr: str = Field(default="", description="Captured stderr from execution")
     error: Exception | None = Field(default=None, description="Exception if execution failed")
+    formatted_error: str = Field(
```
furgalep: This is an abstraction leak.

Is there another way to do this?

Like SandboxException stored in error:

src/nooa/errors/formatting.py
furgalep: Addressed in b07e206. The heredoc recovery hint is now API-neutral: it shows only a valid triple-quoted command value and directs the model to inspect the available runner with doc(...). It no longer hard-codes shell.run or the nonexistent shell.write API.

src/nooa/runtime/events.py
furgalep: Addressed in b07e206. The EventsApi docs now identify standalone Error events as strategy/retry failures and distinguish them from failed cell executions, which are PythonOutput events.

src/nooa/runtime/sandbox/serialization.py
furgalep: Addressed in b07e206. Worker formatting now owns semantic capture/truncation once; serialization only applies an independent hard IPC ceiling. The prior envelope parser and rebounding path were removed, with focused transport-limit tests retained.

src/nooa/runtime/events.py
furgalep: Addressed in b07e206. EventsApi now supports execution_status, so the documented query is events.query(type="PythonOutput", execution_status="error", limit=1). Limited SQLite queries use indexed reverse keyset batches and stop at the first match; custom backends retain a compatible fallback.

src/nooa/events.py
furgalep: Addressed in b07e206. ExecutionResult.formatted_error was removed. Sandboxed user-code failures now use the typed SandboxExecutionError boundary, carrying original_type, original_error, and the worker-rendered diagnostic without leaking a backend-specific parallel field into ExecutionResult.

src/nooa/runtime/events.py
furgalep: Follow-up: the SQLite/backend iteration work has now been removed from this PR in bd25493 and moved to #200. This PR retains only the small `execution_status` query filter needed for the corrected failed-execution example.

drazvan:
Approved — the error-context work looks good. One follow-up: tighten `_FRAMEWORK_MARKERS` so bare `"nooa/"` is not matched as a substring in checkout paths; a repo under something like `.../repos/nooa/` can misclassify user/helper frames as framework code and drop them from diagnostics, so matching installed-package paths (e.g. `site-packages/nooa/`) would make the new regression test reliable everywhere.

drazvan (approved):

Issue #187, opened 2026-08-21
test(codeact): parameterize shared conformance tests across execution backends


furgalep:
## Problem

The general CodeAct strategy suite primarily exercises the default `inprocess` execution backend. Sandbox behavior is covered by separate executor and a small number of end-to-end tests, but backend-observable CodeAct contracts are not systematically run against both backends.

This leaves parity gaps where changes can pass the large in-process suite while differing in sandbox execution because of IPC, worker lifecycle, serialization, source registration, or hard-limit behavior.

## Proposed work

Create a shared CodeAct backend-conformance test harness parameterized over:

- `inprocess`
- `sandbox` (when host capabilities permit; use `SandboxConfig(require=False)` for portable semantic tests)

Run the same observable-contract tests for both backends where semantics are intended to match:

- successful return values and implicit/explicit returns;
- stdout and stderr capture;
- ordinary runtime exceptions;
- SyntaxError source/caret fidelity;
- source line and wrapper-offset fidelity;
- namespace persistence across cells;
- helper definitions across cells;
- inline `return_result()` success and failure;
- validation retry behavior;
- cancellation and timeout semantics where parity is intended;
- `PythonOutput` status, error, value, and ordering;
- nested agent/tool calls supported by both backends.

Keep backend-specific suites for behavior that should not be parameterized:

- sandbox security restrictions and OS capability probes;
- IPC serialization failures;
- broker/proxy behavior;
- worker death/restart;
- filesystem/network confinement;
- CPU/memory/hard-timeout enforcement;
- in-process-only implementation details and mocks.

## Acceptance criteria

- [ ] A reusable fixture/factory constructs equivalent CodeAct agents for both backends.
- [ ] A documented matrix identifies shared versus backend-specific contracts.
- [ ] Shared conformance tests run for both backends in CI where supported.
- [ ] Capability-dependent sandbox cases skip with an explicit reason rather than silently falling back.
- [ ] Failures identify the backend in pytest node IDs.
- [ ] Runtime remains reasonable; slow/security tests stay separately marked.
- [ ] Existing backend-specific suites are deduplicated where the new conformance suite supersedes them.

## Context

Follow-up from #185, which found that Python error source context had separate in-process rendering and sandbox IPC failure modes. The bug fix includes a targeted sandbox regression, but the broader CodeAct suite remains predominantly in-process.

H2908:
I'd like to pick this up if it's not already claimed.

[...]

PR #206, opened 2026-08-24
fix(sandbox): restore traceback frame context in agent-facing errors

### LINKED ISSUE #191: Sandbox backend drops traceback context from agent-facing errors: worker_traceback is captured but never consumed

H2908:
Fixes #191.

### Problem

Errors raised inside a sandboxed cell reach the LLM as a bare `Type: message` line, with no frame, line number, or source text. The in-process backend renders full IPython-style context for identical code.

Before:

    --- execution_backend=inprocess ---
      Cell In[1], line 6, in <module>
        boom = None.strip()
               ^^^^^^^^^^

    --- execution_backend=sandbox ---

Agents running the sandbox backend get error feedback with no location information, weakening the error-recovery loop precisely in the configuration intended for untrusted code.

### Cause

`serialization.py` already captured the worker traceback and attached it to the reconstructed exception as `worker_traceback` — but that attribute had two write sites and no read sites. `_format_runtime_error` derives frames solely from `error.__traceback__`, which is empty on a parent-side surrogate because tracebacks aren't picklable. It therefore returned at the early guard before any rendering ran.

### Approach

Marshal structured frames rather than relying on the formatted string.

`extract_tb` resolves source text eagerly, and it runs inside the worker where the cell is registered with `linecache`, so `(filename, lineno, name, line)` tuples carry everything the renderer needs across the boundary. `FrameSummary` objects are rebuilt on the parent and the existing filter/format pipeline runs unchanged — cell filenames already pass `_is_user_code_frame`, and `__repl_wrapper__` is the wrapper name on both paths.

I considered instead teaching the formatter to consume the `worker_traceback` string, but that would mean re-parsing formatted text to redo framework-frame filtering that already works on structured data.

The in-process branch is untouched; the fallback engages only when `__traceback__` is absent.

### Scope

Caret columns are deliberately excluded. They're computed in wrapper coordinates and the two backends anchor them differently, which is a separable concern from restoring frame context. The parity test asserts frame location and source text, and documents why it stops short of byte-identity. Happy to take that on in a follow-up if you'd like it.

### Tests

Adds `tests/runtime/sandbox/test_error_traceback_parity.py` with three cases: an in-process baseline guard (so a regression there isn't misread as a sandbox fault), the sandbox requirement, and cross-backend parity of frame lines. The error sits mid-file at line 6 so an off-by-one in offset handling would be caught.

- `uv run pytest -m sandbox -q` — 51 passed, 1 skipped (48 before this change)
- `uv run pytest -q` — 6641 passed, 152 skipped, 3 xfailed
- `uv run ruff format` / `ruff check` — clean

Verified on commit 97f52de, Python 3.12.3, Ubuntu.

### Note on #187

This surfaced while prototyping the shared conformance fixture that #187 asks for. The open CI question there still stands — `addopts` deselects the `sandbox` marker, so these tests don't run in the default suite. Happy to follow up once you've weighed in on whether that should be a separate job or a default-run subset.

### DIFF
--- a/src/nooa/errors/formatting.py
+++ b/src/nooa/errors/formatting.py
@@ -362,11 +362,16 @@ class IPythonErrorFormatter:
                     ~~^~~
             ZeroDivisionError: division by zero
         """
-        if not error.__traceback__:
-            return f"{type(error).__name__}: {error}"
+        # Frames come either from a live traceback (in-process) or, when the cell
+        # ran in the sandbox worker, from frames marshaled across the process
+        # boundary — a reconstructed exception has no live ``__traceback__``.
+        if error.__traceback__:
+            extracted = traceback.extract_tb(error.__traceback__)
+        else:
+            extracted = list(getattr(error, "worker_frames", ()))
 
-        # Extract all frames as FrameSummary objects (filterable)
-        extracted = traceback.extract_tb(error.__traceback__)
+        if not extracted:
+            return f"{type(error).__name__}: {error}"
 
         # Filter to user code frames only
         user_frames = [frame for frame in extracted if _is_user_code_frame(frame.filename)]
--- a/src/nooa/runtime/sandbox/serialization.py
+++ b/src/nooa/runtime/sandbox/serialization.py
@@ -39,6 +39,7 @@ class ErrorDTO:
     type_name: str
     message: str
     traceback: str
+    frames: list[tuple[str, int, str, str]] = field(default_factory=list)
 
 
 @dataclass
@@ -109,6 +110,10 @@ def result_to_dto(result: Any) -> ResultDTO:
             type_name=type(err).__name__,
             message=message,
             traceback="".join(tb.format_exception(type(err), err, err.__traceback__)),
+            frames=[
+                (f.filename, f.lineno or 0, f.name, f.line or "")
+                for f in tb.extract_tb(err.__traceback__)
+            ],
         )
         return dto
 
@@ -182,8 +187,15 @@ def _reconstruct_error(err: ErrorDTO) -> Exception:
                 exc = _SurrogateCellError(err)
         else:
             exc = _SurrogateCellError(err)
+
     if err.traceback:
         exc.worker_traceback = err.traceback  # type: ignore[attr-defined]
+    if err.frames:
+        import traceback as _tb
+
+        exc.worker_frames = [  # type: ignore[attr-defined]
+            _tb.FrameSummary(fn, ln, nm, line=src) for fn, ln, nm, src in err.frames
+        ]
     return exc
 
 
--- a/tests/runtime/sandbox/test_error_traceback_parity.py
+++ b/tests/runtime/sandbox/test_error_traceback_parity.py
@@ -1,0 +1,82 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Error-rendering parity between the in-process and sandbox execution backends.
+
+The sandbox backend marshals exceptions across a process boundary, so the live
+``__traceback__`` cannot survive. These tests pin the requirement that the
+agent-facing error text stays equivalent regardless of backend.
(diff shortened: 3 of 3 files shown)


H2908:
Thanks for the thorough work on #189 — I can see it covers the same core issue plus a much wider set of improvements. Happy to close #206 in favour of that once it lands. I'll look for other ways to contribute.

furgalep:
Yes, I think this is fixed in https://github.com/NVIDIA-NeMo/labs-OO-Agents/pull/189. Let me merge that now.

PR #207, opened 2026-08-24
feat(tui): improve fullscreen follow behavior

furgalep:
## Summary
- return the fullscreen transcript to the live tail whenever non-empty input is submitted
- show a semantic **New agent message** notice beside **Return to bottom** when agent prose arrives while the viewport is anchored
- cap the fullscreen input composer at 10 rows and route navigation to it while it overflows

## Implementation
- carry an explicit `agent_message` marker through both direct agent-event rendering and structured `AgentMessage` frontend rendering
- keep unseen-message state UI-owned and clear it when the user returns to the tail or clears the transcript
- leave activity/tool output unmarked so it does not trigger the notice
- while the capped composer has hidden rows, mouse-wheel gestures over it move a detached viewport without moving the edit cursor; typing or cursor movement resumes cursor following
- Page Up/Page Down intentionally move the composer cursor; short drafts retain transcript navigation

## Validation
- `uv run pytest -q packages/nooa-cli/tests/tui/test_fullscreen_transcript.py` — 177 passed
- `uv run pytest -q packages/nooa-cli/tests/tui` — 1204 passed, 2 pre-existing macOS `/var` vs `/private/var` path failures
- Ruff check and format pass on changed files
- Pyright passes on changed source and test files
- `git diff --check` passes

## Review
Two independent focused reviews were run. One identified missing semantic propagation through `TerminalFrontend`; that path is now covered. Final review reported no findings.

## Summary by CodeRabbit

- **New Features**
  - Added improved fullscreen composer scrolling for long drafts, including mouse-wheel and page-key navigation.
  - Limited the composer viewport to prevent it from overwhelming the transcript.
  - Added notifications when new agent messages arrive while viewing older transcript content.
  - Returning to the latest transcript or clearing it removes the notification.
- **Bug Fixes**
  - Restored cursor-following behavior after editing or moving through the composer.
  - Returning to the transcript tail after submitting a message now behaves consistently.

### DIFF
--- a/packages/nooa-cli/src/nooa_cli/tui/agent_event_renderer.py
+++ b/packages/nooa-cli/src/nooa_cli/tui/agent_event_renderer.py
@@ -203,7 +203,7 @@ class AgentEventRenderer:
                 Rule(Text("OO ", style=self._colors["mauve"]), style="dim", align="left"),
                 **emit_kwargs,
             )
-        self._emit_text(Markdown(str(text)), **emit_kwargs)
+        self._emit_text(Markdown(str(text)), agent_message=True, **emit_kwargs)
 
     # ── agent event handlers ───────────────────────────────────────────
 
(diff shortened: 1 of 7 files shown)


furgalep:
@coderabbitai review

PR #209, opened 2026-08-25
fix(agentdoc): preserve type docs for instances

furgalep:
## What does this PR do?

Makes `doc(instance)` preserve the complete type-level API contract from `doc(type(instance))`, then augment it with current values and public runtime-only fields.

Specifically, it:
- retains annotation-only fields, defaults, properties, methods, and docstrings in instance docs
- overlays current instance values and includes both `__dict__` and Pydantic `extra="allow"` fields
- documents plain classes with custom `__repr__` instead of replacing their API docs with repr output
- keeps `pformat()` repr semantics unchanged
- respects per-instance visibility overrides for both declared and dynamic fields
- preserves declared fields omitted by `SupportsInstanceValues` in `doc()`, while `pformat()` continues to omit them
- avoids executing arbitrary descriptors while reading inherited slots
- documents the resulting `SupportsInstanceValues` contract

## Rendered examples

Every output below was generated directly with this PR's committed code at `2da7420`.

### 1. Type parity, current values, dynamic fields, properties, and custom `__repr__`

```python
from nooa.agentdoc import doc, pformat

class Worker:
    """Runs queued work."""

    required_queue: str
    retries: int = 3

    @property
    def status(self) -> str:
        """Current worker status."""
        raise AssertionError("must not execute")

    def run(self, job_id: str) -> bool:
        """Run one job."""
        return True

    def __repr__(self) -> str:
        return "Worker(custom repr)"

worker = Worker()
worker.retries = 5
worker.region = "us-east"
```

`doc(Worker)`:

```python
class Worker:
    """Runs queued work."""

    required_queue: str
    retries: int = 3
    status: str  # Current worker status.

    def run(self, job_id: str) -> bool:
        """Run one job."""
```

`doc(worker)`:

```python
class Worker:
    """Runs queued work."""

    required_queue: str
    retries: int = 5
    status: str  # Current worker status.

    region: str = 'us-east'

    def run(self, job_id: str) -> bool:
        """Run one job."""
```

`pformat(worker)` still respects the author's representation:

```text
Worker(custom repr)
```

This demonstrates that instance docs retain the annotation-only field, property metadata, class/method docs, and method signature; overlay `retries`; add `region`; and never evaluate `status`.

### 2. Per-instance visibility for declared and dynamic fields

```python
from nooa.agentdoc import doc, spec

class Credentials:
    token: str = "class-token"

credentials = Credentials()
credentials.token = "runtime-token"
credentials.endpoint = "https://example.test"
[...]

PR #213, opened 2026-08-25
refactor(agentdoc): unify document rendering paths

furgalep:
## Summary

- route every `doc()` call through one typed subject-normalization and reference-graph pipeline
- separate neutral instance introspection from formatting, with one-way dependencies
- centralize bounded traversal, deduplication, `expand=False`, primary exclusion, and deterministic ordering
- extract every primary and referenced contract once so discovery and rendering cannot diverge
- make identity handling robust for equal/hash-colliding and unhashable metaclass types
- add focused parity, extractor, visibility, registry, cycle/diamond, structured-info, and compatibility coverage

## Why

Follow-up to #209. While reviewing the `inline_depth` fix, we found the single-object and multi-object implementations had independently accumulated different behavior. Adding an unrelated second object could expose `expand=False` types, truncate referenced APIs at 50 members, reorder references, or introspect `TypeInfo` itself.

This makes one-object documentation the one-element case of the same pipeline rather than maintaining two policy implementations.

## Architecture

- `_document.py` owns typed `DocumentSubject` normalization and reference-graph traversal.
- `_introspection.py` owns neutral, safe instance classification and value extraction.
- `_pformat.py` consumes those layers and remains responsible for rendering.
- Dependency direction is one-way: `_pformat` → `_document` / `_introspection`; `_document` does not import `_pformat`.
- Direct `pformat()` keeps its historical inline referenced-type behavior while reusing the shared graph policy.

## Behavior changes

- `spec(expand=False)` is honored in single and multi-object docs at every traversal level.
- Referenced APIs in multi-object docs are no longer silently capped at 50 members.
- Referenced types use deterministic `(name, module, qualname)` ordering.
- Repeated primary objects are deduplicated by identity while first-occurrence order is retained.
- Structured `TypeInfo`, `CallableInfo`, and `ModuleInfo` inputs are not misclassified as ordinary instances.
- Custom extractors run once per documented primary or referenced contract, and the extracted contract is reused for discovery and rendering.
- Distinct types remain distinct even when their metaclass overloads equality/hash; unhashable metaclasses are supported.

## Review

Two focused review rounds covered simplicity/maintainability and test coverage. Their architecture findings were incorporated, followed by two final read-only reviews. The final reviewers reported no remaining actionable architecture/correctness findings; the sole low-priority test assertion was strengthened to an exact-output check.

## Validation

- `uv run pytest tests/agentdoc src/nooa/agentdoc/tests -q` — **739 passed**
- full `uv run pytest -q` — **6,829 passed**, 7 skipped, 284 deselected, 3 xfailed
- changed-file pre-commit hooks — passed (`ruff`, format, Pyright, whitespace, SPDX, etc.)
- changed-source Pyright — **0 errors**
- `git diff --check` — passed

## Dependency

This is intentionally stacked on #209 (`fix/issue-199-doc-instance-parity`). Once #209 merges, this PR can be retargeted to `main`.

## Summary by CodeRabbit

* **New Features**
  * Added unified documentation rendering for single and multiple objects.
  * Added structured instance inspection for dataclasses, Pydantic models, NamedTuples, attrs classes, slotted classes, and custom objects.
  * Added metadata and runtime value extraction, including linked type and callable references.
  * Added bounded reference discovery with deterministic ordering and cycle handling.

* **Bug Fixes**
  * Improved support for unhashable or equality-customized types.
  * Preserved field ordering and override behavior across inheritance hierarchies.
  * Improved extractor registration and unregistration reliability.

### DIFF
--- a/src/nooa/agentdoc/_discover.py
+++ b/src/nooa/agentdoc/_discover.py
@@ -10,6 +10,23 @@ import inspect
 import typing
 from typing import Any
 
+
+class _IdentityTypeSet:
+    """Minimal add/iteration collection that deduplicates types by identity."""
+
+    def __init__(self) -> None:
+        self._items: list[type] = []
+        self._ids: set[int] = set()
+
+    def add(self, item: type) -> None:
+        if id(item) not in self._ids:
+            self._ids.add(id(item))
+            self._items.append(item)
+
+    def __iter__(self):
+        return iter(self._items)
+
+
 # Builtins to exclude from referenced types
 BUILTIN_TYPE_NAMES = {
     "str",
@@ -50,11 +67,17 @@ STDLIB_MODULES = {
 }
 
 
+def _identity_contains(items: Any, candidate: type) -> bool:
+    """Return whether *items* contains candidate by identity, without hashing it."""
+    return items is not None and any(item is candidate for item in items)
+
+
 def discover_referenced_types(
     obj: type | Any,
     *,
     seen: set[type] | None = None,
     field_names: set[str] | None = None,
+    type_info: Any = None,
 ) -> list[type]:
     """Discover all custom types referenced in a class or callable's interface.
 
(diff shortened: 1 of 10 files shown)


furgalep:
@coderabbit full review

PR #217, opened 2026-08-26
docs(config): mark event budgets as reserved

### LINKED ISSUE #125: max_event_tokens and min_preserved_events are documented, validated, and read by nothing 🤖🤖🤖
## Problem

`TruncationConfig.max_event_tokens` and `TruncationConfig.min_preserved_events` are documented as bounding events, accept values, validate those values — and are read by nothing. Setting either has no effect.

This is already known internally. `skills/nooa-codeact-advanced/SKILL.md:77`:

> **Dead/stale knobs**: `max_event_tokens` and `min_preserved_events` are read by nothing — events are never evicted by the renderer.

The gap is that the knowledge lives in a skill file for agents working on the repo, while the user-facing config still promises the opposite.

## What the public config tells a user

`config/truncation_config.py`, module docstring:

> Two layers of bounds:
> 1. **Whole-context token budget** (optional, applied at assembly): `max_context_tokens`, `max_event_tokens`.

Field and comment:

```python
max_event_tokens: Annotated[int | None, Field(description="Total event token budget")] = None
# L4 eviction: never evict fewer than this many recent events. Guarantees
# the model always sees its current Task + recent reasoning even when older
# events get evicted to fit the budget.
min_preserved_events: Annotated[
    int,
    Field(description="Minimum number of recent events preserved during eviction"),
] = 5
```

"applied at assembly", "Total event token budget", and "**Guarantees**" all describe behaviour that does not exist. There is no event eviction to preserve events from.

## Why it reads as working

Three things reinforce the impression, which is what makes this a trap rather than just a stale comment:

1. **Setting it succeeds.** `TruncationConfig(max_event_tokens=50_000)` stores `50_000`.
2. **Invalid values are rejected.** `_check` validates `max_event_tokens > 0`, so `max_event_tokens=-1` raises. A field that carefully validates its input does not look inert.
3. **A test exercises it.** `tests/runtime/test_truncation_config_integration.py::test_max_event_tokens_does_not_crash_on_generation` sets it and asserts generation does not crash. That passes on an inert field, so it reads as coverage without pinning any behaviour.

Meanwhile the eviction that *does* run budgets from `max_context_tokens` alone — `actor.py:2975`, `effective_context_limit = tc.max_context_tokens`, falling back to `usable_window // 2`. `max_event_tokens` never enters the calculation.

## Impact

An operator who sets `max_event_tokens` to protect against runaway event growth believes they have a bound and does not. It fails in the direction that looks fine until a long run blows the window — and `max_context_tokens` also defaults to `None`, so out of the box neither knob is doing anything.

Related: #96 measured one rendered value reaching 25 MB with the per-value limits respected. That case is not helped by these fields either, so "set a token budget" is not currently an available mitigation.

## Do not just delete them

Worth flagging because it is the obvious first instinct and it is the worst option here. `TruncationConfig.model_config` is `{'frozen': True}` with no `extra="forbid"`, so Pydantic's default applies:

```
>>> TruncationConfig(bogus_field=123)      # no error
>>> "bogus_field" in _.model_dump()
False
```

Removing the fields would turn `TruncationConfig(max_event_tokens=50_000)` from "accepted and ignored" into "silently swallowed and ignored" — same outcome, but now with no field to grep for and no error to notice. Anyone who already sets it in production would get zero signal.

## Suggested fix

Smallest honest change, no behaviour change:

1. Correct the module docstring so layer 1 lists only `max_context_tokens`.
2. Reword both fields to say they are currently not enforced, e.g. `Field(description="Reserved: not currently enforced — events are not evicted. Use max_context_tokens.")`, and drop the "Guarantees" comment.
3. Rename the existing test to something like `test_max_event_tokens_is_accepted_but_inert`, so it documents the state instead of implying coverage.

Optionally, and I would like a maintainer view on whether this is wanted: emit a one-time warning when either is set to a non-default, since the population most affected is people who already set them and think they are protected.

If the intent is to *implement* event eviction rather than document the gap, that is obviously the better answer and this issue becomes a tracking issue for it — but that is a design call and a much larger change, so I have not assumed it.

Happy to send the PR for 1–3; it is docs-in-code plus a test rename, and I would rather it be a deliberate choice than a guess. Just say which direction.

🤖🤖🤖

cpakkamisaac-sae:
## Summary

- list only `max_context_tokens` as the enforced assembly-time token budget
- document `max_event_tokens` and `min_preserved_events` as compatibility fields that do not currently evict events
- strengthen the existing integration test so an event is proven to survive an impossible reserved event budget

## Scope

This intentionally does not add event eviction or warnings. The existing fields remain accepted and validated for configuration compatibility while their current behavior is documented honestly.

## Testing

- `uv run ruff format --check src/nooa/config/truncation_config.py tests/runtime/test_truncation_config_integration.py`
- `uv run ruff check src/nooa/config/truncation_config.py tests/runtime/test_truncation_config_integration.py`
- focused tests: 166 passed
- `uv run pyright src/nooa/config/truncation_config.py`: 0 errors
- full suite: 6,806 passed, 7 skipped, 284 deselected, 3 xfailed

Fixes #125

## Summary by CodeRabbit

* **Documentation**
  * Clarified that `max_context_tokens` controls the enforced context budget.
  * Documented `max_event_tokens` and `min_preserved_events` as compatibility settings that are accepted but do not independently evict events.

* **Tests**
  * Updated token-budget coverage to verify that compatibility settings remain inert and existing events stay available when configured.

### DIFF
--- a/src/nooa/config/truncation_config.py
+++ b/src/nooa/config/truncation_config.py
@@ -5,7 +5,7 @@
 Two layers of bounds:
 
 1. **Whole-context token budget** (optional, applied at assembly):
-   ``max_context_tokens``, ``max_event_tokens``.
+   ``max_context_tokens``.
 2. **Sub-configs by mechanism**:
    - ``capture`` (head/tail truncation of raw text streams via
      ``TruncatingStringIO`` — used for execute_python's stdout / stderr /
@@ -25,6 +25,13 @@ Two layers of bounds:
      intentionally by the author, often large (e.g. ``doc(self)``,
      formatted state). Defaults to fully unlimited; whole-block eviction
      handles overflow at the assembly step.
+
+``max_event_tokens`` and ``min_preserved_events`` are retained for
+configuration compatibility, but are not currently enforced because events
+are not evicted separately. Event-level eviction is planned as part of the
+context API refactor; the fields remain so configuring them stays a visible
+no-op rather than becoming a silently swallowed keyword. Use
+``max_context_tokens`` for an enforced budget.
 """
 
 from typing import Annotated
@@ -181,13 +188,23 @@ class TruncationConfig(BaseModel):
     max_context_tokens: Annotated[int | None, Field(description="Total context token budget")] = (
         None
     )
-    max_event_tokens: Annotated[int | None, Field(description="Total event token budget")] = None
-    # L4 eviction: never evict fewer than this many recent events. Guarantees
-    # the model always sees its current Task + recent reasoning even when older
-    # events get evicted to fit the budget.
+    max_event_tokens: Annotated[
+        int | None,
+        Field(
+            description=(
+                "Reserved for compatibility; not currently enforced because events "
+                "are not evicted separately. Use max_context_tokens for an enforced budget."
+            )
+        ),
+    ] = None
     min_preserved_events: Annotated[
         int,
-        Field(description="Minimum number of recent events preserved during eviction"),
+        Field(
+            description=(
+                "Reserved for compatibility; not currently enforced because events "
+                "are not evicted separately. Use max_context_tokens for an enforced budget."
+            )
+        ),
     ] = 5
     # Output-token planning reserve. When a call sets ``max_tokens``
     # explicitly, THAT value is reserved out of the model window (the provider
--- a/tests/runtime/test_truncation_config_integration.py
+++ b/tests/runtime/test_truncation_config_integration.py
@@ -245,7 +245,7 @@ class TestConfigMergeEdgeCases:
 
 
 class TestTokenBudgetIntegration:
-    """Tests for token budget fields (max_context_tokens, max_event_tokens)."""
+    """Tests for enforced and reserved token-budget settings."""
 
     @pytest.mark.asyncio
     async def test_max_context_tokens_does_not_crash_on_generation(self):
@@ -272,7 +272,7 @@ class TestTokenBudgetIntegration:
         assert result == "done"
 
     @pytest.mark.asyncio
-    async def test_max_event_tokens_does_not_crash_on_generation(self):
+    async def test_max_event_tokens_is_accepted_but_inert(self):
         """Agent with max_event_tokens set should not raise ValueError during generation."""
         from nooa.unifiedllm import FakeLLMClient
 


Hotragn:
I filed #125, so flagging two places where this diverges from the implementation notes @alessiodevoto left on that issue. Both are small and I think worth closing before merge — the diff is otherwise the right shape, and I agree with the decision to leave the fields in place rather than delete them.

**1. The docstring note is missing its "why".**

The issue note asked for "a short note in the module docstring saying event-level eviction is planned as part of the context API refactor," in the context of the API being mid-refactor and event-level eviction likely landing as part of that work. This PR's note explains *that* the fields are inert:

> ``max_event_tokens`` and ``min_preserved_events`` are retained for
> configuration compatibility, but are not currently enforced because events
> are not evicted separately.

but not that enforcement is planned. Without that, the next reader can't tell whether these are deprecated leftovers on the way out or unfinished scaffolding on the way in — which is the distinction the note was meant to preserve. Suggested addition, wording open:

```
Event-level eviction is planned as part of the context API refactor; the two
fields are kept rather than removed so that setting them stays a visible
no-op instead of a silently swallowed keyword.
```

**2. The test assertion was changed, and I read the issue note as asking for it not to be.**

The note was explicit that the rename was welcome but "please keep the assertion as-is so it continues to pin the current behavior." This PR keeps the rename and rewrites the assertion — `max_event_tokens=50_000` becomes `=1`, `min_preserved_events=0` is added, a prior event is injected, and the new assertion is that it survives:

```python
assert prior_event in json.dumps(llm.last_messages)
```

The old test pinned "setting this is accepted and generation still succeeds." The new one pins "no event eviction happens," which is a claim about a feature's *absence*. That inverts the moment the planned eviction lands: a correct future implementation with `max_event_tokens=1, min_preserved_events=0` should drop that event, so this test will fail and read like a regression rather than an expected behaviour change. Given that the same reviewer said the eviction work is likely coming as part of the refactor, leaving the assertion untouched seems like the point.

Reverting to the original body and keeping only the rename plus the docstring gets both — the name stops implying budget coverage, and the assertion stays valid across the refactor.

**Disclosure, since it's relevant:** I had a branch for this ready on 2026-08-13 after @alessiodevoto scoped it on the issue, and didn't open the PR — that's on me, and this comment isn't a claim on the work. @cpakkamisaac-sae, happy to hand over the two snippets above as a patch if that's easier than reworking it yourself, or to leave it entirely with you. Either is fine by me.

One naming nit, take or leave: the issue note suggested `test_max_event_tokens_is_accepted_but_inert`; this uses `test_event_budget_settings_are_accepted_but_inert`. The latter is arguably more accurate now that it covers both fields, so I mention it only so the difference is deliberate rather than missed.

🤖🤖🤖

alessiodevoto (approved):
LGTM thanks!

PR #233, opened 2026-08-28
fix(repo): support current Tree-sitter query API

furgalep:
## Summary
- compile Tree-sitter queries through `Language.query(...)` when the installed binding exposes the legacy API
- use `tree_sitter.Query(language, source)` for Tree-sitter 0.26+
- add branch-level compatibility tests and a real 0.26 integration regression for symbols and references

Capability detection is used instead of parsing package version strings, so compatible backports and future releases follow the API they actually expose.

## Validation
- focused compatibility suite: 3 passed with the locked Tree-sitter 0.26.0 AST extra
- complete `packages/nooa-cli/tests`: 1514 passed, with only 2 known unrelated macOS `/var` vs `/private/var` failures
- real Tree-sitter 0.26 symbol extraction and reference filtering verified
- Ruff format/check passed
- Pyright: 0 errors, 0 warnings
- `git diff --check` passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with legacy and current Tree-sitter APIs.
  * Restored reliable symbol extraction and reference detection for JavaScript, while correctly ignoring comments and strings.
  * Improved TypeScript class definition detection.

* **Tests**
  * Added coverage for Tree-sitter query compilation and capture handling.
  * Added integration tests for JavaScript symbol and reference analysis.
  * Added validation for TypeScript definition queries.

### DIFF
--- a/packages/nooa-cli/src/nooa_cli/tools/_tree_sitter_backend.py
+++ b/packages/nooa-cli/src/nooa_cli/tools/_tree_sitter_backend.py
@@ -139,6 +139,14 @@ def _get_parser(lang: str) -> ts.Parser | None:
         return None
 
 
+def _compile_query(language: ts.Language, source: str) -> ts.Query:
+    """Compile a query using the API exposed by the installed Tree-sitter version."""
+    legacy_query = getattr(language, "query", None)
+    if callable(legacy_query):
+        return legacy_query(source)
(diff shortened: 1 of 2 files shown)


alessiodevoto (approved):

alessiodevoto:
Thanks! LGTM!

PR #238, opened 2026-08-29
Fix JSON parser matching non-JSON Markdown fences

ryana:
## What does this PR do?

Restricts `extract_and_parse_json()` to unwrap Markdown fences only when the entire trimmed response is enclosed by a bare or explicit `json` fence. Non-JSON fences such as `bash` and `python` are left in the complete response, so they cannot mask valid JSON appearing later or produce errors based only on the fenced command.

### Where this came from

This surfaced when I asked OO to translate a Markdown file into another language. The document contained `bash`, `json`, and bare code blocks, and the JSON parser treated the first non-JSON fence as a JSON wrapper, causing havoc during translation.

### Design note

IMHO, we should not bend over backwards to accommodate models that do not return JSON when asked for JSON. Current models are capable of following that contract, and every additional recovery heuristic creates more ambiguous edge cases and a never-ending game of whack-a-mole. This PR is deliberately narrow: it fixes the incorrect fence match without adding new recovery behavior.

## Related issues

None.

## Validation

- `uv run pytest tests/unifiedllm/test_json_parsing.py` — 12 passed
- `uv run ruff check src/nooa/unifiedllm/unifiedllm.py tests/unifiedllm/test_json_parsing.py` — passed

## Checklist

- [x] Code follows the project style for the changed files
- [x] Tests added/updated and passing
- [x] No documentation changes needed; this narrows internal parser cleanup behavior
- [x] No new source files

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON parsing for Markdown-fenced responses.
  * Fenced JSON is now recognized only when it constitutes the complete response, reducing accidental parsing of unrelated content.
  * Non-JSON fences embedded in prose no longer interfere with valid JSON found later.
  * Invalid complete fences now preserve the original content while reporting a parsing error.
  * Added clearer tracking for removed JSON fences.

### DIFF
--- a/src/nooa/unifiedllm/unifiedllm.py
+++ b/src/nooa/unifiedllm/unifiedllm.py
@@ -343,8 +343,8 @@ def extract_and_parse_json(text: str) -> dict[str, Any]:
     original_text = text
     text = text.strip()
 
-    markdown_pattern = r"```(?:json)?\s*\n?(.*?)\n?```"
-    markdown_match = re.search(markdown_pattern, text, re.DOTALL)
+    markdown_pattern = r"```(?:json)?[ \t]*\r?\n(.*?)\r?\n?```"
+    markdown_match = re.fullmatch(markdown_pattern, text, re.DOTALL)
     if markdown_match:
         _record_llm_metric("json_fence_removed")
         text = markdown_match.group(1).strip()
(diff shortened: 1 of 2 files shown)


furgalep (approved):
LGTM

PR #243, opened 2026-09-01
feat(tracing): compose scoped instrumentation hooks

furgalep:
## What does this PR do?

Adds a generic, composable instrumentation seam so embedding hosts can observe NOOA lifecycle events without replacing native OpenInference tracing.

The immediate consumer is the NeMo Gym adapter, which needs one rollout-local collector for native NOOA invocation identity, nesting, status, and timing. The API is intentionally Gym-independent so other hosts can attach observers without coupling NOOA core to their schemas.

## Why this is needed

Before this change, NOOA had one active `InstrumentationHooks` object per async context. Calling `set_hooks()` replaced that object. This did **not** prevent trace fan-out: one `OpenInferenceHooks` already emitted spans through an OpenTelemetry provider with multiple exporters (OTLP, JSONL, journal, Langfuse, and so on).

The missing capability was different: two independent **hook-level lifecycle observers** could not safely coexist unless callers built their own composite. A host such as Gym either had to replace OpenInference instrumentation or reconstruct lifecycle data from a separate event stream. Replacement loses native tracing; reconstruction duplicates NOOA semantics and can lose parentage, failure, cancellation, or timing details.

This PR separates those concerns:

- OpenTelemetry exporter fan-out remains unchanged.
- Independent hook implementations can now be composed.
- A host can add a task-local observer for one invocation and restore the prior observer set afterward.

### Composable hook observers

`CompositeInstrumentationHooks` dispatches every lifecycle callback to child hooks in registration order:

- agent calls
- generations
- code execution
- method invocation
- tool execution
- rendered-message observation

Each `before_*` result is retained separately and only returned to the matching child's `after_*` callback. Observer failures are isolated so instrumentation cannot interrupt the workload or prevent later observers from receiving the event.

`compose_hooks()`:

- ignores `None`
- flattens existing composites
- deduplicates observers by identity
- preserves callback order
- avoids unnecessary wrappers for zero or one hook

### Scoped, task-local installation

`hooks_scope()` uses the existing hook `ContextVar` with token-based set/reset semantics:

```python
with hooks_scope(rollout_hooks):
    result = await agent.run(...)
```

By default this composes `rollout_hooks` with the current observer set. On exit it restores the exact prior value, including on exception or cancellation. Because the storage is context-local, concurrent asyncio rollouts can use independent collectors without global multiplexing or cross-rollout leakage.

This helper is not conceptually mandatory—a caller could reproduce it with `get_hooks()`, `compose_hooks()`, `ContextVar.set()`, and token reset—but centralizing the behavior prevents each host from implementing subtly different nesting and restoration rules.

### Before/after ownership

Both singular and composite `before_*` calls now return an internal ownership envelope. `call_after_hook()` dispatches through the observer that created the context, not whichever observer happens to be active later. This matters when scopes change while an async operation is suspended.

### OpenInference coexistence

`NemoOOAgentsInstrumentor` now adds its hook to the current composition rather than overwriting it. Uninstrumentation removes only that instrumentor's hook by identity. Multiple instrumentors can therefore be torn down out of order without resurrecting stale observers or removing unrelated ones.

### Session attribution

`session_scope()` provides matching token-safe OpenTelemetry session binding:

```python
with session_scope(rollout_id), hooks_scope(rollout_hooks):
    result = await entrypoint(...)
```

This keeps trace session IDs isolated across concurrent tasks and restores nested values exactly.

### Cancellation and failure semantics

Lifecycle exception fields now accept `BaseException` so `asyncio.CancelledError` can be reported accurately while still being re-raised. Agent calls, generation, method invocation, `return_result`, and code execution—including sandbox execution—now report cancellation as failure rather than success.

Observer-originated `CancelledError` is treated like an observer failure and isolated, consistent with the existing contract that instrumentation callbacks must never affect agent execution. Process-control exceptions such as `KeyboardInterrupt` and `SystemExit` are not broadly swallowed.

## Public API

Exported from `nooa.runtime`:

- `CompositeInstrumentationHooks`
- `compose_hooks()`
- `hooks_scope()`

Exported from `nooa.tracing`:

- `session_scope()`

Existing `InstrumentationHooks`, `get_hooks()`, and `set_hooks()` remain available. Existing single-hook implementations continue to work.

## Compatibility notes

- This is additive for normal hook consumers.
- Exception annotations widen from `Exception | None` to `BaseException | None`; runtime callback shape is otherwise unchanged.
- Existing OTLP/file/journal exporter behavior is not replaced or redesigned.
- `set_hooks()` retains replacement semantics for callers that explicitly want it; `hooks_scope()` is the safe additive/scoped path.

## Review hardening

Independent review identified and this revision fixes:

- stale hook resurrection when multiple OpenInference instrumentors are uninstrumented out of order
- singular-hook context misrouting if the active hook changes between `before_*` and `after_*`
- `return_result` cancellation being reported as success
- an observer's `CancelledError` escaping and cancelling the workload

Regression tests cover each case.

## Validation

On the pre-rebase feature tree:

- full suite: **6839 passed, 7 skipped, 285 deselected, 3 xfailed**
- only explicit deselection: unrelated host-font Matplotlib test

After rebasing onto current `main` and resolving the two textual conflicts:

- focused runtime/tracing suite: **257 passed**
- review-fix suite: **260 passed**
- Ruff on changed runtime/tracing/test files: passed
- Pyright on changed runtime/tracing modules: passed
- `git diff --check origin/main...HEAD`: passed
- independent final review: no release-blocking correctness or compatibility findings

## Related work

This enables a NeMo Gym integration to attach a rollout-specific native NOOA observer while retaining OpenInference tracing. That downstream integration should depend on the first NOOA release containing these APIs.

## Checklist

- [x] Code follows the project style
- [x] Tests added/updated and passing
- [x] Public API/changelog updated
- [x] New source files carry SPDX headers

## Summary by CodeRabbit

- **New Features**
  - Added composable instrumentation hooks that can be scoped and combined with existing hooks.
  - Added scoped tracing sessions for isolating trace data across nested and concurrent executions.
  - Added tracing support for synchronous and asynchronous generators, including lifecycle events and context restoration.

- **Bug Fixes**
  - Improved cancellation handling while preserving cancellation behavior and reporting failures correctly.
  - Preserved existing instrumentation when tracing integrations change.
  - Improved execution error reporting with clearer sandbox details and consistent formatting.


furgalep:
Addressed both review comments in f8b10b1: session and hook test state now use unconditional try/finally cleanup. Focused tracing tests pass (15/15).

furgalep:
Addressed in c8a7190: composite activation now unwinds each child context manager defensively and preserves the original workload result/exception if a child cleanup fails. Added direct cleanup-failure coverage; generator span-tree and hook-composition tests pass (22/22). The prior CI generator span failures were fixed by forwarding activation through composites and unwrapping singular owned contexts; the subsequent full Linux test job passes. The separate actor.py and codeact.py observations are pre-existing origin/main code outside this PR's changed hunks.

alessiodevoto:
@coderabbit review

alessiodevoto:
One minor thing: we still have `NemoOOAgentsInstrumentor` under the `nooa.tracing` public API (and this PR adds new uses) . Since the project has moved to NOOA/nooa naming, should we add a new alias/name such as `NOOAInstrumentor` , keep `NemoOOAgentsInstrumentor` as a backwards-compatible alias, and update the new tests/usages to exercise the new spelling? We can keep this for a separate PR as I see we still have some of these forgotten variables around. Otherwise, LGTM thanks!

alessiodevoto (approved):

PR #248, opened 2026-09-02
TUI fullscreen follow-ups: perf, correctness, code sharing

furgalep:
# TUI fullscreen follow-ups

Follow-up backlog from the review rounds of the merged select-copy stack (#244, #245). Seven commits, each independently verified; full TUI suite **1428 passed** (previously flaky suite runs are now clean at the root).

## Performance
- **Event detail rendering memoized** (`_styled_detail_lines`): the search chain walked the same markdown up to six times per keystroke (match lines, occurrences, highlighted output, measurement, paint). One Rich render per `(row, width)`, with the row retained in the cache entry so a recycled `id` can never alias stale lines.

## Correctness
- **`/model` probe cancellation gap**: `except Exception` cannot catch `CancelledError` (a `BaseException` in Python 3.13). A cancelled probe after `_begin_model_validation` stranded deferred startup prompts forever. Ownership now resolves on cancellation before re-raising. Regression cancels a gated probe and asserts `reject_deferred_messages` fires.
- **Memory arm/confirm gesture window**: the docstring promised a window that never existed — an armed `f` from minutes earlier still fired. Arms now carry a timestamp and expire after 10s; a stale arm re-arms instead of firing.

## Code sharing
- **Test-only snapshot renderer retired**: `render_resume_picker`/`_preview_lines`/`_wrap` had no production callers (the picker renders through the shared browser). They moved into the test package as a deterministic fixture, and the superseded `ResumePickerModel.scroll_preview`/`preview_offset` scroll state went with them (−116 production lines).
- **Shared browser construction template-methoded**: `ExplorerBrowser.__init__` gained `_create_list_control`/`_option_window_width` factory hooks and shared search chrome; `ResumePicker.__init__` collapsed to its model, its view facade, and `super().__init__` (−42 net lines, ~85 duplicated construction lines gone). The duplicate `invalidate()` deleted — the base one covers every control.

## Test-suite stability (both flaky tests fixed at the root)
- **Cancel-status race**: the test asserted cleanup had not started between the cancel and the next line — a scheduler interleaving that only holds on an idle machine. The mid-flight ordering assertion is gone; the synchronous-acknowledgement contract assertions remain.
- **Fake-session leak**: `test_resume_list_fills_its_pane` assigned `SessionManager` classmethods with bare `classmethod` objects — no monkeypatch, no teardown — leaking 12 fabricated sessions (`s00000001…`) into every later test (why the memory-sidecars test failed intermittently in full-suite runs). Now patched through monkeypatch.

## Verification
- Full TUI suite: **1428 passed**, zero flakes across repeated runs
- Ruff clean
- Every commit replayed to GitHub via the Git Data API with trees verified identical to local

## Summary by CodeRabbit

- **Bug Fixes**
  - Cancelling model validation now correctly resolves pending prompts.
  - Expired confirmations no longer display misleading hints and require a fresh confirmation.
  - Resume-picker preview scrolling, filtering, sorting, and selection are more reliable.
  - Event details load more efficiently when revisiting content.
  - Search and close controls now update correctly when displayed options change.

- **Tests**
  - Added coverage for confirmation expiry, cancellation handling, event rendering and caching, and resume-picker behavior.

### DIFF
--- a/packages/nooa-cli/src/nooa_cli/tui/commands.py
+++ b/packages/nooa-cli/src/nooa_cli/tui/commands.py
@@ -562,6 +562,21 @@ class ModelCommand(Command):
                     )
                 )
             return CommandResult.err(f"Failed to switch model: {e}")
+        except BaseException as e:  # noqa: BLE001 — ownership must resolve on cancel too
+            # Cancellation is a BaseException in Python 3.13: if the probe is
+            # cancelled after _begin_model_validation took over the startup
+            # prompt ownership, nothing would ever reject or release the
+            # deferred prompts. Resolve ownership on the way out, then
+            # propagate the cancellation to the caller.
+            if model_validation_started:
+                self._mark_model_check_failed(
+                    HealthCheckResult(
+                        ok=False,
+                        error_message=f"Model validation cancelled: {e}",
+                        blocking=True,
+                    )
+                )
+            raise
         self.config.default_model = selected
         self._mark_model_ready(selected)
         try:
--- a/packages/nooa-cli/src/nooa_cli/tui/commands.py
+++ b/packages/nooa-cli/src/nooa_cli/tui/commands.py
@@ -511,7 +511,11 @@ class ModelCommand(Command):
                 TextOutput(f"Current model: {self.config.default_model}", "info")
             )
         selected = args[0]
-        model_validation_started = False
+        # Ownership of the deferred startup prompts needs resolving only
+        # while the probe is in flight. The switch phase below has already
+        # produced a healthy probe; cancelling THERE must not publish a
+        # failed health (the agent may already be on the new model).
+        probe_in_flight = False
         try:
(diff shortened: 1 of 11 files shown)


furgalep:
The latest review pass is addressed in 8e213b61:

- `_query_changed()` now resets the cached current preview after `set_query()`, so a typing-swapped selection follows the transcript tail like the filter/sort handlers.
- The memory arm/confirm hint now expires on its own: arming schedules a `loop.call_later` deadline that clears `_pending_confirm` and runs expiry callbacks (wired to `ExplorerBrowser.invalidate`), cancelled on every consume/disarm/fire path.
- The `_fragment_cache` invalidation pattern is the established project-wide prompt-toolkit idiom; leaving as-is per discussion.

Verified: full TUI suite 1429 passed locally, Ruff clean, new no-keypress expiry regression included.

PR #259, opened 2026-09-04
fix(tui): bound resume preview workers

furgalep:
## Summary

- serialize actual `/resume` preview renderer work so cancelled `asyncio.to_thread` awaiters cannot leave multiple GIL-heavy chunks running concurrently
- retain and drain both superseded preview tasks and their executor wrappers, including repeated-cancellation during picker close
- restore the main UI/focus immediately when the picker closes, then finish bounded worker cleanup before the dialog coroutine returns
- omit the currently executing TUI coroutine from exit diagnostics while preserving visibility of executor threads that may still be blocked

## Why

Rapid session paging could outpace the 120 ms preview dwell gate. Cancelling the asyncio task did not stop a chunk already executing in the default thread pool, so stale chunks accumulated, starved the UI, and were later joined by `asyncio.run()` after the goodbye message. The reported `Task-1: main` was the diagnostics function observing its own caller rather than leaked work.

## Verification

- `uv run pytest packages/nooa-cli/tests/tui -q` — 1,443 passed
- focused resume-picker/session-plumbing tests — 100 passed
- Ruff check — passed
- `git diff --check` — passed

New regressions cover worker serialization, close-time draining, repeated cancellation during drain, and actionable exit diagnostics.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resume preview stability by preventing overlapping renders and ensuring cancelled preview work finishes before the picker closes.
  * Improved session shutdown diagnostics by excluding the active diagnostic task from pending-task reports.
  * Ensured focus and picker cleanup complete reliably when closing the resume dialog.

* **Tests**
  * Added coverage for preview rendering, cancellation, worker cleanup, and shutdown diagnostics.

### DIFF
--- a/packages/nooa-cli/src/nooa_cli/tui/resume_picker.py
+++ b/packages/nooa-cli/src/nooa_cli/tui/resume_picker.py
@@ -632,6 +632,17 @@ class ResumePicker(ExplorerBrowser):
         # reuses their transcript instead of growing without limit.
         self._preview_models: OrderedDict[tuple[str, int], Any] = OrderedDict()
         self._preview_tasks: dict[tuple[str, int], Any] = {}
+        # Cancellation of an asyncio.to_thread awaiter does not stop its worker.
+        # Serialize actual render calls so superseded chunks cannot accumulate
+        # and contend for the GIL while the user keeps navigating.
+        import asyncio
+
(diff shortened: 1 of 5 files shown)


furgalep:
@coderabbitai review

PR #261, opened 2026-09-04
fix(llm): retain OpenAI Responses reasoning state 🤖🤖🤖

furgalep:
## Summary

- preserve native Responses output items and replay encrypted reasoning state across direct and event-sourced turns
- pass `reasoning`, `store`, and `include` from registry entries so stateless Responses sessions can request replayable reasoning items
- make `/reasoning off` send the explicit Responses effort `none` while preserving `reasoning.context`
- add GPT-5.6 `xhigh` and `max` controls for Responses clients while preserving existing Completion-client behavior

## Scope

This revision removes the earlier usage normalization, cache accounting, and token-parameter translation. A general model/provider capability layer remains follow-up work. Append-only CodeAct correction is tracked in #264.

## Validation

- 151 focused tests passed
- Ruff lint passed
- Ruff formatting check passed

🤖🤖🤖

### DIFF
--- a/src/nooa/config/model_config.py
+++ b/src/nooa/config/model_config.py
@@ -16,9 +16,9 @@ dicts (see ``unifiedllm/registry.py``); this model is the first step.
 
 from __future__ import annotations
 
-from typing import Any
+from typing import Annotated, Any
 
-from pydantic import BaseModel, ConfigDict
+from pydantic import BaseModel, ConfigDict, Field
 
 
 class ModelConfig(BaseModel):
@@ -43,9 +43,46 @@ class ModelConfig(BaseModel):
     max_tokens: int | None = None
     temperature: float | None = None
     top_p: float | None = None
-    reasoning: dict[str, Any] | None = None
-    store: bool | None = None
-    include: list[str] | None = None
+    # OpenAI Responses API reasoning controls, forwarded verbatim to
+    # ``litellm.responses(reasoning=...)``. Typically ``{"effort": "medium",
+    # "context": "all_turns"}``; ``effort`` is the thinking budget
+    # (none/low/medium/high/xhigh/max, model-dependent) and ``context``
+    # selects how much prior reasoning the model may attend to.
+    reasoning: Annotated[
+        dict[str, Any] | None,
+        Field(
+            description=(
+                "OpenAI Responses reasoning controls forwarded to the provider, "
+                "e.g. {'effort': 'medium', 'context': 'all_turns'}."
+            )
+        ),
+    ] = None
+    # OpenAI Responses API server-side state policy. ``store=False`` runs the
+    # session stateless (no server-retained response chain) so the client must
+    # replay the full output — the ZDR-friendly mode this framework uses. Leave
+    # unset to accept the provider default (``store=True`` on OpenAI).
+    store: Annotated[
+        bool | None,
+        Field(
+            description=(
+                "Responses API state policy: False keeps the session stateless "
+                "(client replays history); unset uses the provider default."
+            )
+        ),
+    ] = None
+    # OpenAI Responses API ``include`` list — extra output fields the provider
+    # should return. For stateless reasoning continuity set
+    # ``["reasoning.encrypted_content"]`` so encrypted reasoning items come back
+    # for replay on the next turn.
+    include: Annotated[
+        list[str] | None,
+        Field(
+            description=(
+                "Responses API output fields to include, e.g. "
+                "['reasoning.encrypted_content'] for stateless reasoning replay."
+            )
+        ),
+    ] = None
 
     @classmethod
     def from_registry(cls, name: str, raw: dict[str, Any]) -> ModelConfig:
--- a/src/nooa/config/model_config.py
+++ b/src/nooa/config/model_config.py
@@ -43,6 +43,9 @@ class ModelConfig(BaseModel):
     max_tokens: int | None = None
     temperature: float | None = None
     top_p: float | None = None
+    reasoning: dict[str, Any] | None = None
+    store: bool | None = None
+    include: list[str] | None = None
 
     @classmethod
     def from_registry(cls, name: str, raw: dict[str, Any]) -> ModelConfig:
--- a/src/nooa/runtime/actor.py
+++ b/src/nooa/runtime/actor.py
@@ -1242,7 +1242,15 @@ class ActorRuntime:
         elif not isinstance(content, str):
             # Other non-string types - convert to string representation
             content = str(content)
-        event = LLMOutput(content=content)
+        assistant_message = getattr(response, "assistant_message", None)
+        reasoning_items = (
+            assistant_message.get("reasoning_items")
+            if isinstance(assistant_message, dict) and not response.tool_calls
+            else None
+        )
+        if not isinstance(reasoning_items, list):
+            reasoning_items = None
+        event = LLMOutput(content=content, reasoning_items=reasoning_items)
         event_id = self.event_manager.add(event)
 
         return response, event_id
--- a/src/nooa/runtime/actor.py
+++ b/src/nooa/runtime/actor.py
@@ -1165,6 +1165,15 @@ class ActorRuntime:
                         "cached_tokens": getattr(_prompt_details, "cached_tokens", None),
                     }
                 )
+            _input_details = getattr(_usage_raw, "input_tokens_details", None)
+            if _input_details is not None:
+                _usage_dict["input_tokens_details"] = (
+                    _input_details
+                    if isinstance(_input_details, dict)
+                    else {
+                        "cached_tokens": getattr(_input_details, "cached_tokens", None),
+                    }
+                )
             _completion_details = getattr(_usage_raw, "completion_tokens_details", None)
             if _completion_details is not None:
                 _usage_dict["completion_tokens_details"] = (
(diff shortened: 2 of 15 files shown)


src/nooa/config/model_config.py
```diff
@@ -43,6 +43,9 @@ class ModelConfig(BaseModel):
     max_tokens: int | None = None
     temperature: float | None = None
     top_p: float | None = None
+    reasoning: dict[str, Any] | None = None
+    store: bool | None = None
```
furgalep: Need some info about what "store" means. `store_reasoning`?

src/nooa/config/model_config.py
```diff
@@ -43,6 +43,9 @@ class ModelConfig(BaseModel):
...
+    include: list[str] | None = None
```
furgalep: What does `include` mean? 

We should Annotate all of these parameters so they are self documenting.

furgalep (commented):
Not ready yet

sklinglernv:
Codex said:

/reasoning accepts effort levels unsupported by the active model. packages/nooa-cli/src/nooa_cli/tui/commands.py:1076 writes none,
    xhigh, or max without checking model capabilities. For example, /reasoning off configures none for GPT‑5 Pro, which only supports high;
    /reasoning max is invalid for GPT‑5.4 Pro, which supports only medium, high, and xhigh. The command reports success, but the next request
    fails provider validation. Capability-gate the choices and reject unsupported values; where none is unavailable, “off” should either restore
    the model default or report that reasoning cannot be disabled. GPT‑5 Pro documentation
    (https://developers.openai.com/api/docs/models/gpt-5-pro), GPT‑5.4 Pro documentation
    (https://developers.openai.com/api/docs/models/gpt-5.4-pro).

sklinglernv:
Does this work for models where LiteLLM switches from completion client to response client behind the scenes? This is happening, for example, for GPT 5.6.

src/nooa/runtime/actor.py
```diff
@@ -1242,7 +1242,15 @@ async def _core_llm(ctx: LLMCallContext) -> LLMCallContext:
...
-        event = LLMOutput(content=content)
+        assistant_message = getattr(response, "assistant_message", None)
+        reasoning_items = (
+            assistant_message.get("reasoning_items")
+            if isinstance(assistant_message, dict) and not response.tool_calls
+            else None
+        )
+        if not isinstance(reasoning_items, list):
+            reasoning_items = None
+        event = LLMOutput(content=content, reasoning_items=reasoning_items)
```
alessiodevoto: This stores `reasoning_items` on text-only/non-tool `LLMOutput`, but CodeAct can later remove that event before the next request in two text-only recovery paths:

  - `src/nooa/strategies/codeact.py:1033`: the `finish_reason == "stop"` / `text_only_stop_behavior == "return_result"` branch removes `event_id`, then calls `_process_tool_calls(...)` at line 1047 without passing `reasoning_items`. If synthetic `return_result` validation fails, execution continues to
  another LLM request after line 1065 with the retained reasoning state gone.
  - `src/nooa/strategies/codeact.py:1103`: the `synthetic_comment` branch removes `event_id`, then creates a synthetic `ToolCallEvent` at line 1106 without `reasoning_items` and continues at line 1151.

  Normal real tool-call turns do preserve this via `_process_tool_calls(..., reasoning_items=...)`, so this looks specific to synthetic text-only replacement.

furgalep:
## Review: retain OpenAI Responses reasoning state

I reviewed this against the OpenAI Responses reasoning-continuity guidance and checked it out locally. **LGTM — I recommend merging.** It implements the highest-value items from the earlier audit, and it is well-scoped and well-tested.

### What it does well

- **Complete native output replay.** `_responses_output_items` + `_responses_assistant_message` retain the full `_batch` of typed output items and expose the `reasoning` subset, and `_transform_messages` replays `_batch` verbatim. This is exactly the stateless (`store:false`) full-item-replay pattern OpenAI documents for reasoning continuity, and it future-proofs against new output-item types.
- **Text responses now preserve reasoning too.** Previously only tool-call turns kept native output; storing `reasoning_items` on `LLMOutput` and re-emitting them before the assistant text in `ResponsesProviderFormatter` closes the gap for `reasoning.context: all_turns`.
- **Correct provider isolation.** The reasoning-item replay is confined to the Responses paths. `OpenAIProviderFormatter`'s text branch does not emit `reasoning_items`, so Completion-client models safely ignore the new field.
- **`/reasoning off` semantics are right.** Sending explicit `effort: "none"` while preserving `context` avoids silently restoring GPT-5.6's `medium` default, and Completion clients keep their prior behavior.
- **State-policy config plumbed.** `store` / `include` / `reasoning` flow from the registry, enabling stateless replayable sessions.

### Verification I ran (worktree `fix/gpt56-responses-reasoning`)

- Targeted suite: **151 passed** (`test_responses_reasoning_state`, `test_model_registry`, `test_formatters`, `test_llm_complete_event`, `test_reasoning_command`, `test_completer`).
- Broader regression: **428 passed** across `test_litellm_responses_bridge` + `tests/strategies` reasoning/responses/codeact.
- `ruff check` and `ruff format --check`: clean.

### Non-blocking observations

1. **Actor guard is intentional but subtle.** `reasoning_items` is stored on `LLMOutput` only when `not response.tool_calls`. That's correct (tool turns already replay the batch through CodeAct), but a one-line comment on the `and not response.tool_calls` condition would help future readers.
2. **Event-sourced text replay reconstructs the assistant message rather than replaying the stored native `message` item.** The reasoning item (the actual provider state) is preserved and re-emitted, and the text is regenerated — functionally correct for continuity. Worth a brief note that only `reasoning_items`, not the full `_batch`, survives on the event-sourced text path.
3. **`cached_tokens` from `input_tokens_details` is still not parsed** (the scope note already defers usage/cache accounting). That remains the main follow-up so Responses cache hits are observable.

Nice work — this is the right foundation for moving `gpt-5.6-sol` onto Responses once the cache telemetry follow-up lands.

furgalep:
## Update — addressed all three review notes

Pushed `924e4f0` to `fix/gpt56-responses-reasoning`:

1. **Responses cache telemetry (substantive).** `actor.py` now reads `usage.input_tokens_details.cached_tokens` (the Responses API's cache-hit location) and includes it in the `cached_tokens` fallback chain — previously Responses cache hits were recorded as `0`. Added `test_cached_tokens_from_responses_input_tokens_details` covering the native usage shape.
2. **Actor guard comment.** Documented why `reasoning_items` is stored on `LLMOutput` only for non-tool turns (tool turns already replay the full native batch via CodeAct tool events, so re-storing would double-replay).
3. **Event-sourced text-replay comment.** Clarified that only `reasoning_items` (not the full `_batch`) survive on the event-sourced text path, so `ResponsesProviderFormatter` re-emits reasoning before a reconstructed assistant message.

### Verification
- Targeted: **152 passed** (incl. the new cache test).
- Regression: **442 passed** across `test_litellm_responses_bridge`, `test_responses_cache_control`, and `tests/strategies` (reasoning/responses/codeact/cache).
- `ruff check` + `ruff format --check`: clean.

furgalep:
## Update — addressed Severin's and Paul's comments

Pushed `12dedcd` to `fix/gpt56-responses-reasoning`.

### Paul (inline, `src/nooa/config/model_config.py`)
`reasoning`, `store`, and `include` are now `Annotated` with self-documenting `Field(description=...)`:
- **`reasoning`** — OpenAI Responses reasoning controls forwarded to the provider, e.g. `{"effort": "medium", "context": "all_turns"}`.
- **`store`** — Responses state policy; `False` keeps the session stateless (client replays history), unset uses the provider default.
- **`include`** — Responses output fields to include, e.g. `["reasoning.encrypted_content"]` for stateless reasoning replay.

### Severin — Codex: `/reasoning` accepts effort levels the model can't support
Added a per-model reasoning-effort capability table and gated `validate_args` against it:
- An effort the model doesn't accept is now rejected up front (e.g. `/reasoning max` on GPT-5.4 Pro) instead of failing provider validation on the next request.
- Where a model can't disable reasoning (e.g. GPT-5 Pro, high-only), `/reasoning off` is reported as unsupported rather than silently sent as an invalid effort.
- Unknown models stay permissive (classic `off|low|medium|high`); `xhigh|max` keep the Responses-client gate.
- Added capability regression tests.

### Severin — "does this work when LiteLLM switches completion→responses (GPT-5.6)?"
Verified against the live registry:
- `gpt-5.6-sol` and `gpt-5.6-responses` resolve to **`ResponsesClient`**, so they get the native `reasoning` dict and the full effort ladder including `xhigh`/`max`.
- The plain `gpt-5.6` alias resolves to **`CompletionClient`** — that's the LiteLLM-bridged path. There, `/reasoning` writes the Chat-style `reasoning_effort` (which LiteLLM translates when it bridges to Responses), and `xhigh`/`max` remain gated because they're Responses-only controls. So `/reasoning` works for the bridged case for `off|low|medium|high`; it just doesn't expose the two Responses-only levels on a Completion client. If you'd like those two levels enabled for bridged GPT-5.6 specifically, that's a follow-up once we can detect the bridge reliably.

### Alessio
His concern about synthetic text-only recovery paths dropping reasoning state is tracked in **#264** (append-only CodeAct correction), so I left it for that issue rather than duplicating it here.

### Verification
- Targeted: **155 passed**; config: **64 passed**; bridge/cache regression: green.
- `ruff check` + `ruff format --check`: clean.

src/nooa/config/model_config.py
furgalep: @-

src/nooa/config/model_config.py
furgalep: @-

furgalep:
## Rebalanced: #264 work split into stacked PR #268

To keep this PR focused on **retaining OpenAI Responses reasoning state**, I moved the append-only text-only recovery and model-family provenance work into a separate stacked PR: **#268** (base `fix/gpt56-responses-reasoning`).

This branch is now back to just the reasoning-retention scope:
- `2fef2b3` retain OpenAI Responses reasoning state
- `924e4f0` parse Responses cache tokens; clarify reasoning-state comments
- `12dedcd` annotate Responses config; capability-gate `/reasoning`

(Force-pushed to drop the two #264 commits; they now live on `fix/codeact-append-only-264` in #268.) Targeted suite green (93 passed).

PR #268, opened 2026-09-05
fix(codeact): preserve text-only model turns

furgalep:
## Summary

- preserve every text-only `LLMOutput` unchanged in history
- append a separate `TextOnlyReply` metadata event for tracing and diagnostics
- recover through a sync-or-async callback instead of replacing the provider turn
- default to model-visible correction and retry; offer text-as-result as an explicit opt-in
- reject provider error/truncation finishes before the callback can accept partial text
- remove the old text-only behavior modes rather than carrying alpha compatibility code

## Extension point

`TextOnlyResponseAction` is the callback result, not an argument to `@strategy`:

```python
@strategy(CodeActStrategy(on_text_only=return_text_as_result))
async def summarize(self, text: str) -> str: ...
```

The handler receives `TextOnlyResponseContext` with the untouched response, normalized text, current call, and declared return type. It returns one of:

| Action | Meaning |
|---|---|
| `return_result(value)` | Validate through the normal CodeAct result path. |
| `retry(*events)` | Append feedback and ask the model again. |
| `tool_calls(*calls)` | Run synthetic tool calls while retaining the assistant turn. |

Built-ins:

- `retry_text_only_response` (default): append an `Error` asking for `return_result(value)` or `execute_python(code)`, then retry.
- `return_text_as_result` (opt-in): validate non-empty prose as the method result.

Applications can supply their own sync or async policy. On `main`, synthetic execution targets `execute_python`; TUI-only tools are outside this PR.

## Code walkthrough

1. `src/nooa/strategies/codeact.py` defines the public callback contract: `TextOnlyResponseContext` is the input and the frozen `TextOnlyResponseAction` is the handler's decision token.
2. `CodeActStrategy.__init__` accepts `on_text_only`; the default handler returns a retry action carrying a model-visible `Error`, while `return_text_as_result` is the opt-in result policy.
3. The generation loop leaves the runtime-created `LLMOutput` in place, invokes the handler only for a complete text-only response, and records a separate `TextOnlyReply` metadata event for diagnostics.
4. The loop then applies the returned action through existing CodeAct paths: normal return validation, appended retry events, or synthetic tool execution. Provider errors and truncated responses never enter this recovery hook.
5. `src/nooa/config/strategy_config.py` rejects the removed behavior flags with an actionable migration error instead of preserving dead alpha compatibility.
6. The public exports, strategy docs, and CodeAct Skills show both built-in policies and the custom callback form.

## Why

Recovery is append-only. NOOA retains the turn that actually happened while keeping recovery policy as an application extension point.

## Documentation and validation

- public API docstrings and strategy docs cover the callback contract and built-ins
- agent-authoring and advanced CodeAct Skills cover the common and custom cases
- CI, CodeRabbit, and DCO are green
- stack-top full suite: 7,152 passed, 7 skipped, 298 deselected, 3 xfailed
- focused regressions cover non-string results, provider error finishes, safe diagnostics, and empty output
- Ruff and targeted Pyright are clean

### DIFF
--- a/docs/concepts/strategies.md
+++ b/docs/concepts/strategies.md
@@ -54,6 +54,57 @@ variables, helper definitions persist across cells, and generated code can call
 visible methods and tools on `self`. The loop ends when a value validates
 against the return annotation.
 
+### Recovering from a text-only CodeAct turn
+
+CodeAct expects every model turn to call `execute_python` or `return_result`,
+but models sometimes emit plain prose instead. NOOA keeps that original
+assistant turn in history, then asks a configurable callback what to do:
+
+```python
+from nooa import CodeActStrategy, return_text_as_result, strategy
+
+
+class Summarizer(Agent, llm=llm):
+    @strategy(CodeActStrategy(on_text_only=return_text_as_result))
+    async def summarize(self, text: str) -> str:
+        """Summarize the text."""
+        ...
+```
+
+`TextOnlyResponseAction` is the callback's return value; it is not passed to
+`@strategy`. The wiring is:
+
+1. `@strategy(...)` attaches a `CodeActStrategy` to the method.
+2. `CodeActStrategy(on_text_only=handler)` registers a sync or async callback.
+3. When a text-only response occurs, the handler receives a
+   `TextOnlyResponseContext` and returns a `TextOnlyResponseAction`.
+
+The default handler, `retry_text_only_response`, appends a model-visible
+`Error` telling the model to use one of its tools and retries. The opt-in
+`return_text_as_result` handler validates the bare text against the method's
+return type. Custom handlers can return `TextOnlyResponseAction.return_result`,
+`.retry`, or `.tool_calls`; all paths preserve the provider's original turn.
+
+For example, an application can supply its own model-visible correction:
+
+```python
+from nooa import TextOnlyResponseAction, TextOnlyResponseContext
+from nooa.events import Error
+
+
+def require_tool(context: TextOnlyResponseContext) -> TextOnlyResponseAction:
+    return TextOnlyResponseAction.retry(
+        Error(content=f"Plain text cannot finish {context.call.method_name}; call a tool.")
+    )
+
+
+class Investigator(Agent, llm=llm):
+    @strategy(CodeActStrategy(on_text_only=require_tool))
+    async def investigate(self, question: str) -> str:
+        """Investigate the question."""
+        ...
+```
+
 ## Strategy selection is not model selection
 
 The strategy controls the interaction pattern. The LLM setting controls which
--- a/skills/nooa-agent-authoring/SKILL.md
+++ b/skills/nooa-agent-authoring/SKILL.md
@@ -195,7 +195,16 @@ async def classify(self, text: str) -> Intent: ...
 async def implement(self, task: str) -> str: ...
 ```
 
-**Constructors take `config=` only.** `PredictStrategy(max_retries=3)` and `CodeActStrategy(max_iterations=10)` are errors — wrap options in `PredictConfig(...)`/`CodeActConfig(...)`. Useful `CodeActConfig` fields: `max_iterations`, `max_retries`, `cell_timeout`, `max_tokens`, `temperature`, `max_consecutive_text_only`, `restrictions`.
+**Configuration fields go through `config=`.** `PredictStrategy(max_retries=3)` and `CodeActStrategy(max_iterations=10)` are errors — wrap options in `PredictConfig(...)`/`CodeActConfig(...)`. Strategy-level extension points such as `CodeActStrategy(on_text_only=...)` remain direct constructor arguments. Useful `CodeActConfig` fields: `max_iterations`, `max_retries`, `cell_timeout`, `max_tokens`, `temperature`, `max_consecutive_text_only`, `restrictions`.
+
+CodeAct preserves a model response that contains prose but no tool call. By
+default it appends an `Error` asking the model to use `execute_python` or
+`return_result`, then retries. For a method where prose is a valid final value,
+use `CodeActStrategy(on_text_only=return_text_as_result)`. `on_text_only`
+receives a sync or async callback; the callback receives
+`TextOnlyResponseContext` and returns `TextOnlyResponseAction`. The action is
+the callback result—it is not passed to `@strategy`. See
+`nooa-codeact-advanced` for custom retry and synthetic-tool examples.
 
 `max_iterations` is a safety net, not the main tuning dial — decompose the task instead of raising the cap. For prefill control, truncation tuning, code restrictions, and the full config surface, see `nooa-codeact-advanced`.
 
--- a/skills/nooa-codeact-advanced/SKILL.md
+++ b/skills/nooa-codeact-advanced/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: nooa-codeact-advanced
-description: Advanced tuning of NOOA strategies — CodeAct prefill (understanding, disabling, custom, pre-ellipsis code), loop guards (max_iterations, retries, text-only stop), truncation tuning (TruncationConfig/CaptureConfig/FormatConfig), code restrictions (RestrictionsConfig), execution environment internals, and PredictStrategy tuning (retries, param guards, output_serialization). Use when configuring CodeActConfig or PredictConfig beyond defaults, writing a custom prefill, restricting generated code, or debugging truncation/eviction behavior.
+description: Advanced tuning of NOOA strategies — CodeAct prefill, loop guards, text-only recovery callbacks, truncation, code restrictions, execution internals, and PredictStrategy tuning. Use when configuring CodeActConfig or PredictConfig beyond defaults, handling models that return prose instead of tool calls, writing a custom prefill, restricting generated code, or debugging truncation and eviction.
 compatibility: nooa package
 ---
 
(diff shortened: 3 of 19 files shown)


furgalep:
@coderabbit full review

src/nooa/events.py
```diff
@@ -126,7 +126,7 @@ class TextOnlyReply(EventBase):  # type: ignore[misc]
     but ``Role.METADATA`` — never rendered to the model, so capturing a
     text-only reply does not change generation. This is the durable, structured record that
     powers ``/bug`` capture and time-travel replay; the model-visible
-    correction is a separate ``Error``/``Feedback`` event (``Role.USER``).
+    recovery action is recorded separately and may be supplied by the host.
 
     Replaces the lossy ``DebugTrace`` previously written on the CodeAct
```
furgalep: Remove this "replaces the..." That's not info about the current code.

furgalep:
CodeRabbit follow-up is complete.

- Removed text-only config keys now fail explicitly with migration guidance to CodeActStrategy(on_text_only=...) (61b27489).
- Empty stop responses remain persisted for tracing/resume but no longer become empty assistant messages in the next provider request; state-bearing or reasoning-bearing events remain replayable (61b27489).
- Whole-file Match provenance now reaches diff generation, including correct no-final-newline markers (c49a5be8).
- The earlier Bedrock consecutive-assistant and custom-dispatch findings were superseded by the callback rewrite: retry appends user feedback after the original assistant turn, and the single sync/async handler is invoked before action dispatch.
- The old reasoning_provenance thread targeted a superseded version of this PR. Provider-state isolation is intentionally owned by child PR #310, where opaque state uses a private issuer-scoped envelope and fails closed on incompatible routes.

Validation: 7,019 passed, 7 skipped, 3 xfailed in the full repository suite; focused post-cleanup tests 35 passed; Ruff passes; changed production files pass Pyright.

src/nooa/tools/shell_tools.py
```diff
@@ -122,6 +124,11 @@ def text(self) -> str:
         """Raw file content (no line numbers)."""
         return self._text
 
+    @property
```
sklinglernv: this seems unrelated?

sklinglernv (approved):
Besides the shell tool changes that seem unrelated, LGTM

furgalep:
## Full review: preserve text-only model turns

**Verdict: the redesign is a clear improvement — append-only recovery with a real extension point — but there is one blocker to fix before merge, plus one open provider-compatibility question.** Everything else held up under execution-based review.

### What holds up (verified by running it, not just reading)

- **Append-only recovery is real on all three action paths.** The original `LLMOutput` is preserved for `return_result` (validation failure appends a correction rather than deleting the turn), `retry`, and `tool_calls` (via `preserve_llm_output=True` in `_process_tool_calls`). The only remaining `event_manager.remove()` is the empty-response branch, which is correctly scoped: `stop`-with-empty-content goes through the handler branch instead (the `_has_text or finish_reason == "stop"` condition), gets preserved in the event log, and is filtered from provider context.
- **The empty-turn fix is complete.** `context_builder.py` now drops empty `LLMOutput` from provider context (event kept for persistence/diagnostics) — this resolves the earlier review finding about empty assistant messages reaching APIs.
- **The config break is loud, not silent.** `text_only_stop_behavior`/`text_only_correction`/`text_only_correction_fn` now fail `CodeActConfig` validation with a pointed message directing users to `CodeActStrategy(on_text_only=...)`, rather than being silently ignored. The removed-options rejection fires on the `model_validator(mode="before")` path.
- **`TextOnlyReply` schema migration is safe.** Old stored events with `route`/`recovered` fields deserialize cleanly (extra fields dropped); the `handler`/`action` fields record what actually ran.
- **Callback contract behaves:** sync-or-async dispatch works; a handler raising propagates (correct — recovery must not swallow application errors); a non-`TextOnlyResponseAction` return raises a clear `TypeError`.
- **`bench_agent` and the skills/docs are all migrated off the removed API coherently.**

**Verification I ran:** 192 tests (text-only suite + codeact + config), 1,325 (runtime + context_blocks) — all green. Ruff check + format clean on the four core files. DCO present on all 4 commits; CI green on the head (secret-scan, build, lint, 3.13-compat).

### Blocker — fix before merge

**`codeact.py:1061` crashes for non-string `return_result` values.** The `return_result` path calls `get_harness_metrics().stop_to_return_result(action.value)` **before** validation, but `HarnessMetrics.stop_to_return_result` is annotated `content: str | None` and its truncation helper calls `len(s)` (`harness_metrics.py:42,235`). A custom handler returning `TextOnlyResponseAction.return_result(42)` for a `-> int` method — a completely valid use of the documented contract — aborts an otherwise-valid generation with `TypeError: object of type 'int' has no len()`. **I reproduced this live against the PR head.** Sized non-string values also pollute `stop_to_return_result_previews` (typed `list[str]`) without truncation.

The built-in handlers never hit this (`return_text_as_result` passes the string content; the default retries), which is why the suite is green — it only fires for custom handlers, exactly the extension point this PR advertises. Fix: stringify/guard at the call site (`stop_to_return_result(str(action.value) if action.value is not None else None)`, or make the metric accept `Any`), and add the non-string regression test.

### Open question from the earlier review — still live, now scoped

**Bedrock consecutive-assistant messages on the `tool_calls` path.** The `retry` and `return_result` paths are fine (their appended feedback is `Error`, i.e. user-role). But `tool_calls` preserves the `LLMOutput` (assistant) and then adds a synthetic `ToolCallEvent` (also assistant) with no coalescing in any formatter. Anthropic merges consecutive same-role turns; **Bedrock requires strict alternation and can 400.** This is now the only recovery path with that shape — either coalesce the preserved turn + synthetic call for role-strict providers, or verify Bedrock tolerates it.

### Smaller notes

- The empty-response `remove()` is a defensible carve-out from append-only (a truly empty turn carries nothing to preserve and the debug `DebugTrace` records what happened) — worth one sentence in the docs where append-only is claimed, so nobody reads it as unconditional.
- The synthetic `tool_calls` action re-uses `_process_tool_calls` with `preserve_llm_output=True` — the interaction between the synthetic `ToolCallEvent`'s id (`synthetic_{uuid4}`) and any reasoning state on the preserved turn deserves a regression test if reasoning replay lands on this lineage.
- Test gaps worth adding with the blocker fix: non-string `return_result` values, async handler raising, repeated `return_result` validation failure counting toward the `max_consecutive_text_only` backstop.

**Recommendation:** fix the metrics blocker (small, with a regression test), settle the Bedrock role-alternation question for the `tool_calls` action, and this is mergeable — the design is the right shape and the review history shows it responded well to findings.

furgalep:
Bottom-up review follow-up is pushed in f7b46609.

- `TextOnlyResponseAction.return_result(42)` now reaches normal `-> int` validation; the string-preview metric receives only strings and the regression injects real `HarnessMetrics`.
- Non-empty `finish_reason="length"` output is retained for diagnostics but cannot be accepted as a completed result by `return_text_as_result` or another callback.
- The stale whitespace test description now matches the default retry path.
- The Bedrock question was verified rather than papered over: an 8-token live request through the exact NVIDIA `openai/aws/anthropic/bedrock-claude-sonnet-4-5-v1` route accepted the existing consecutive-assistant history, and LiteLLM 1.97.0 coalesces the same history on native Anthropic/Bedrock Invoke and Converse transports. No model-visible recovery event was added.

Validation: 235 focused tests pass; Ruff, production Pyright, and diff check pass. @coderabbitai review

furgalep:
@coderabbitai review\n\nFollow-up fixes make provider truncation/error authoritative over parsed tool calls for both Completion and Responses clients. A truncated batch is now captured for diagnosis but never executed.

furgalep:
@coderabbitai review\n\nThe same truncation precedence now covers CompletionClient XML tool-call fallback paths, with sync and async regressions.

docs/concepts/strategies.md
```diff
@@ -54,6 +54,57 @@ variables, helper definitions persist across cells, and generated code can call
...
+
+class Summarizer(Agent, llm=llm):
+    @strategy(CodeActStrategy(on_text_only=return_text_as_result))
+    async def summarize(self, text: str) -> str:
+        """Summarize the text."""
+        ...
+```
+
+`TextOnlyResponseAction` is the callback's return value; it is not passed to
+`@strategy`. The wiring is:
```
furgalep, line 75: This is an internal note based on an agent conversation, not general documentation. Make sure this is just documentation.

skills/nooa-agent-authoring/SKILL.md
```diff
@@ -195,7 +195,16 @@ async def classify(self, text: str) -> Intent: ...
...
 ```
 
-**Constructors take `config=` only.** `PredictStrategy(max_retries=3)` and `CodeActStrategy(max_iterations=10)` are errors — wrap options in `PredictConfig(...)`/`CodeActConfig(...)`. Useful `CodeActConfig` fields: `max_iterations`, `max_retries`, `cell_timeout`, `max_tokens`, `temperature`, `max_consecutive_text_only`, `restrictions`.
+**Configuration fields go through `config=`.** `PredictStrategy(max_retries=3)` and `CodeActStrategy(max_iterations=10)` are errors — wrap options in `PredictConfig(...)`/`CodeActConfig(...)`. Strategy-level extension points such as `CodeActStrategy(on_text_only=...)` remain direct constructor arguments. Useful `CodeActConfig` fields: `max_iterations`, `max_retries`, `cell_timeout`, `max_tokens`, `temperature`, `max_consecutive_text_only`, `restrictions`.
+
+CodeAct preserves a model response that contains prose but no tool call. By
+default it appends an `Error` asking the model to use `execute_python` or
+`return_result`, then retries. For a method where prose is a valid final value,
+use `CodeActStrategy(on_text_only=return_text_as_result)`. `on_text_only`
+receives a sync or async callback; the callback receives
```
furgalep, line 204: Remove "The action is
the callback result—it is not passed to `@strategy`."

skills/nooa-codeact-advanced/SKILL.md
```diff
@@ -42,15 +42,71 @@ class Notifier(Agent, llm=llm):
...
-`tool_choice` is hardcoded `"auto"`. There is no `allow_text_response` option (older docs mention one) — text handling is entirely the two text-only knobs.
+`tool_choice` is hardcoded `"auto"`.
+
+## Text-only recovery callback
+
+CodeAct expects each model turn to call `execute_python` or `return_result`.
+When a model emits only prose, NOOA preserves that exact assistant turn and
+then invokes the strategy's `on_text_only` callback.
+
+Do not pass `TextOnlyResponseAction` to `@strategy`. The objects have distinct
```
furgalep, line 60: "The action is
the callback result—it is not passed to `@strategy`."

remove.

PR #161, opened 2026-08-18
feat(skills): simplify text skills with ShellTools 🤖🤖🤖

Rexarrior:
## What does this PR do?

- Loads each SKILL.md directory as a regular generated `Skill` whose class documentation comes from the skill file.
- Exposes a deterministic, skill-root-relative `files: list[SkillFile]` manifest.
- Injects a skill-local `ShellTools(cwd=skill_root)` and removes the bespoke `read_file()` / `run_script()` surface.
- Rebuilds documentation, files, and the shell on agent-local registry reload.
- Updates tests, the quickstart, migration guidance, and the changelog.

A skill-local shell keeps every manifest path directly usable from the skill root and preserves the per-agent object isolation introduced by #127.

## Related issues

Fixes #129

## Checklist

- [x] Code follows the project style (`uv run ruff check .` and `uv run ruff format --check .` pass)
- [x] Tests added/updated and passing (6639 passed, 6 skipped, 284 deselected)
- [x] Docs updated for the public API migration
- [x] New source files carry an SPDX license header (no new source files)

Additional check: `uv run pyright src/nooa/skill.py`

🤖🤖🤖

## Summary by CodeRabbit

* **New Features**
  * File-based skills now provide documentation, stable file manifests, and skill-scoped shell tools.
  * Added public access to skill file metadata and updated code execution capabilities.
  * Added composable instrumentation and trace-session hooks.
* **Bug Fixes**
  * Improved skill discovery, reload isolation, UTF-8 handling, and symbolic-link safety.
  * Corrected tracing for generator agent methods.
* **Documentation**
  * Updated guides and examples for shell-based skill operations and durable human-in-the-loop workflows.
* **Refactor**
  * Replaced direct file-reading and script-running helpers with skill-scoped shell operations.
  * Generator methods using the `...` marker now raise `TypeError`.

PR #281, opened 2026-09-05
fix(skills): decode skill bodies as UTF-8 🤖🤖🤖

Sampoornnagpal:
## What does this PR do?

Skill metadata is read as UTF-8, but `_parse_skill_md()` reads the body using the locale encoding. A UTF-8 skill containing `50€ → café 中文 🚀` becomes corrupted instructions on a cp1252 host. Read the body as UTF-8 too.

Add a regression that forces a legacy default even on UTF-8 hosts and checks the loaded skill's full body. The test fails before the fix.

## Related issues

The parser remains in use in #161; this change is limited to its body decoding and does not alter that refactor's APIs or arbitrary-file encoding policy.

## Validation

`uv run pytest -q tests/unit/test_skill.py -k 'not run_script'` — 19 passed, 7 Bash-dependent tests deselected.

Windows/Python 3.12 with external import-only helpers for #84/#85 (`fcntl`/`SIGUSR2`). Helpers are not included; POSIX locks/signals and Bash execution are not validated by this run.

## Checklist

- [x] Ruff lint and formatting pass.
- [x] Regression and skill-loading tests pass.
- [x] Existing UTF-8 skill format retained.
- [x] Existing SPDX headers retained.

## Summary by CodeRabbit

- **Bug Fixes**
  - Skill files are now consistently read as UTF-8, preserving Unicode characters regardless of the platform’s default text encoding.
  - Skill descriptions continue to be parsed correctly when files contain Unicode content.

- **Tests**
  - Added regression coverage for loading path-based skills with Unicode content under differing default encodings.
  - Added explanatory documentation to related test cases.

### DIFF
--- a/src/nooa/skill.py
+++ b/src/nooa/skill.py
@@ -208,7 +208,7 @@ def _parse_skill_md(path: Path) -> tuple[str, str, str]:
 
     props = _read_skill_properties(path)
     description = props.description or path.name
-    _, body = _parse_frontmatter(skill_md.read_text())
+    _, body = _parse_frontmatter(skill_md.read_text(encoding="utf-8"))
     return skill_id, description, body.strip()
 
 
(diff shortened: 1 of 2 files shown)


alessiodevoto (approved):
LGTM! Thanks for this fix @Sampoornnagpal :)

PR #304, opened 2026-09-07
feat(llm): provider identity and capability contracts

furgalep:
## Summary

Add a transport-neutral contracts module that resolves provider routing strings to logical identity, and types the reasoning/capability surfaces providers disagree about. **Purely additive** — nothing consumes these types yet, existing clients/strategies/formatters are untouched, and importing the module pulls in no provider SDK.

### Why

Model identifiers arrive as routing strings — `openai/nvidia/zai-org/glm-5.3`, `openai/azure/openai/gpt-5.6-sol`, `kimi-k3:free` — that mix gateway routes, vendor namespaces, tier suffixes, and provider aliases into one identifier. Inferring semantic identity from those substrings is exactly how a routing prefix like `openai/` gets mistaken for the OpenAI provider. This module resolves a routing string to a logical provider **exactly once, at the adapter edge**, so no other code has to guess.

### What's in it

- **`ProviderIdentity` + `parse_model_string`** — routing/gateway prefixes, namespace prefixes, provider aliases (`claude→anthropic`, `zai→glm`, `moonshot→kimi`), and tier suffixes (`:free`, `:cloud`) stripped once here. Unresolvable strings fail closed with `UnknownProviderIdentityError` rather than guessing. Bedrock-style dotted ids (`anthropic.claude-sonnet-4-5`) normalize to the same identity as the plain form.
- **`ReasoningKind` / `ReasoningRecord` / `ReasoningReplayMode`** + the recursive `JsonValue` alias — retained reasoning artifacts as kind + payload + provenance in NOOA-owned JSON, so they can be stored and compared without holding provider SDK objects.
- **`derive_opaque_replay_key`** — a non-secret SHA-256 compatibility digest over provider + api_style + endpoint/account scope + the declared model compat group. The compat group is the model dimension of the key (raw model ids are not mixed in, so model-id churn never changes the key); normalization happens inside the derivation and is versioned with the key; undeclared models derive no key rather than a speculative one; transport is excluded.
- **`ModelCompatGroup`** — explicitly declared, provider-scoped membership with conservative defaults for the gpt-5 and claude-4-5 families.
- **`ReasoningCapabilities`** with `effort_map` (`None` = unsupported), a default catalog, and an explicit override registry; unknown providers return `None` rather than being promoted to supported.

### Dependency note

`JsonValue` uses a PEP 695 recursive alias as a Pydantic field annotation, which requires `pydantic>=2.11` — the old-style `Union` alias recurses infinitely under Pydantic 2.x schema generation (verified: `RecursionError`). The pydantic floor is raised to `2.11.0` across the workspace pyprojects; the lockfile already resolved to 2.13.4, so no environment changes in practice.

### Tests

Golden JSON fixtures freeze every contract type's schema; 36 tests cover gateway alias resolution, tier/alias normalization, key stability and divergence, same-group key equality, Bedrock dotted-id convergence, fail-closed paths (unknown identity, undeclared compat groups, unknown capabilities), capability override precedence, and a no-provider-SDK-imports guard. The golden fixture also asserts the frozen digest is reproducible from the derivation function.

`uv run pytest tests/unifiedllm/` — **388 passed**. Ruff check + format clean.

🤖🤖🤖

## Summary by CodeRabbit

- **New Features**
  - Added standardized provider identity and model compatibility support.
  - Added model-string normalization for aliases, namespaces, tiers, and vendor prefixes.
  - Added reasoning capability profiles and structured reasoning records.
  - Added deterministic opaque replay-key generation for compatible models.
  - Added registration and lookup APIs for provider capabilities and compatibility groups.
  - Added public access to unified language-model contract types and utilities, with safe handling for unknown providers and models.

### DIFF
--- a/src/nooa/unifiedllm/__init__.py
+++ b/src/nooa/unifiedllm/__init__.py
@@ -1,5 +1,22 @@
 # SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
+from nooa.unifiedllm.contracts import (
+    OPAQUE_REPLAY_KEY_VERSION,
+    ModelCompatGroup,
+    NormalizedModel,
+    ProviderIdentity,
+    ReasoningCapabilities,
+    ReasoningKind,
+    ReasoningRecord,
+    ReasoningReplayMode,
+    UnknownProviderIdentityError,
+    compat_group_for,
+    derive_opaque_replay_key,
+    get_reasoning_capabilities,
+    parse_model_string,
+    register_compat_group,
+    register_reasoning_capabilities,
+)
 from nooa.unifiedllm.fake import FakeLLMClient
 from nooa.unifiedllm.http_config import HttpConfig
 from nooa.unifiedllm.registry import (
@@ -30,6 +47,22 @@ from nooa.unifiedllm.unifiedllm import (
 )
 
 __all__ = [
+    # Provider contracts (identity and capability types)
+    "ModelCompatGroup",
+    "NormalizedModel",
+    "OPAQUE_REPLAY_KEY_VERSION",
+    "ProviderIdentity",
+    "ReasoningCapabilities",
+    "ReasoningKind",
+    "ReasoningRecord",
+    "ReasoningReplayMode",
+    "UnknownProviderIdentityError",
+    "compat_group_for",
+    "derive_opaque_replay_key",
+    "get_reasoning_capabilities",
+    "parse_model_string",
+    "register_compat_group",
+    "register_reasoning_capabilities",
     # Core classes
     "UnifiedLLM",
     "CompletionClient",
--- a/src/nooa/unifiedllm/contracts.py
+++ b/src/nooa/unifiedllm/contracts.py
@@ -242,43 +242,90 @@ _ROUTING_PREFIXES: frozenset[str] = frozenset(
     {"openai", "nvidia", "azure", "aws", "bedrock", "vertex_ai", "nvidia_nim", "huggingface"}
 )
 
-#: Free-tier / capacity tier suffixes stripped before any comparison or hash
-#: (e.g. "kimi-k3:free", "deepseek-v4-flash:cloud").
-_TIER_SUFFIXES: frozenset[str] = frozenset({"free", "cloud"})
+#: Free-tier / capacity / routing tier suffixes stripped before any comparison
+#: or hash (e.g. "kimi-k3:free", "deepseek-v4-flash:cloud", "o3:batch").
+_TIER_SUFFIXES: frozenset[str] = frozenset({"free", "cloud", "batch"})
 
 #: Provider aliases: a leading segment that names a vendor but is not the
 #: canonical logical provider.  Canonical names map to themselves.
 _PROVIDER_ALIASES: dict[str, str] = {
     "claude": "anthropic",
     "anthropic": "anthropic",
+    "z-ai": "glm",
     "zai": "glm",
     "zai-org": "glm",
     "moonshot": "kimi",
     "moonshotai": "kimi",
+    "meta-llama": "meta",
+    "meta": "meta",
+    "mistralai": "mistral",
+    "mistral": "mistral",
+    "x-ai": "xai",
+    "deepseek-ai": "deepseek",
+    "google": "google",
+    "minimaxai": "minimax",
+    "microsoft": "microsoft",
 }
 
 #: Model-family prefixes -> logical provider.  Matched on the final path
 #: segment, at token boundaries only (so "glm" never matches "glmx").
+#: Prefixes are checked in order; the first boundary-respecting match wins,
+#: so "llama" is checked after "nemotron" to keep NVIDIA's llama-based
+#: nemotron models attributed to nvidia.
 _MODEL_FAMILIES: tuple[tuple[str, str], ...] = (
+    ("gpt-oss", "openai"),
+    ("gpt-audio", "openai"),
+    ("gpt-chat", "openai"),
     ("gpt-", "openai"),
+    ("o1", "openai"),
+    ("o3", "openai"),
+    ("o4", "openai"),
     ("chatgpt-", "openai"),
     ("glm", "glm"),
     ("kimi", "kimi"),
     ("deepseek", "deepseek"),
     ("qwen", "qwen"),
-    ("nemotron", "nvidia"),
     ("claude", "anthropic"),
     ("llama", "meta"),
+    ("muse-spark", "meta"),
+    ("muse-glimmer", "meta"),
     ("mistral", "mistral"),
+    ("ministral", "mistral"),
+    ("mixtral", "mistral"),
+    ("codestral", "mistral"),
+    ("devstral", "mistral"),
+    ("voxtral", "mistral"),
     ("gemini", "google"),
+    ("gemma", "google"),
+    ("lyria", "google"),
     ("grok", "xai"),
+    ("phi-", "microsoft"),
+    ("phi4", "microsoft"),
 )
 _FAMILY_BOUNDARY = "-_.0123456789"
+
+#: Distinctive mid-id markers -> logical provider, checked before prefix
+#: matching.  Vendors derive model lines from other families' bases
+#: ("llama-3.1-nemotron-ultra-..." is NVIDIA's post-trained Llama), so no
+#: prefix rule can attribute them; the marker is unambiguous in practice.
+_FAMILY_MARKERS: tuple[tuple[str, str], ...] = (
+    ("nemotron", "nvidia"),
+    ("nemoguard", "nvidia"),
+    ("nemoretriever", "nvidia"),
+    ("nemosmith", "nvidia"),
+    ("nv-embedqa", "nvidia"),
+    ("nv-rerankqa", "nvidia"),
+    ("nv-embed", "nvidia"),
+    ("nv-rerank", "nvidia"),
+)
 
 
 def _resolve_family(segment: str) -> str | None:
     """Return the logical provider for a final model-id segment, if known."""
     lowered = segment.lower()
+    for marker, provider in _FAMILY_MARKERS:
+        if marker in lowered:
+            return provider
     # Bedrock-style ids embed the vendor: "anthropic.claude-3-5-sonnet".
     if "." in lowered:
         head = lowered.split(".", 1)[0]
@@ -303,6 +350,9 @@ def parse_model_string(model: str) -> NormalizedModel:
     callers must fail closed on that.
     """
     raw = model.strip()
+    # Variant markers like OpenRouter's "~anthropic/claude-..." prefix.
+    raw = raw.lstrip("~")
+    raw = raw.strip()
     if not raw:
         return NormalizedModel(provider=None, model="", namespace=None, tier=None)
 
--- a/src/nooa/unifiedllm/contracts.py
+++ b/src/nooa/unifiedllm/contracts.py
@@ -61,11 +61,15 @@ __all__ = [
     "register_reasoning_capabilities",
 ]
 
-#: NOOA-owned recursive JSON value type.  Provider objects are converted to
-#: this shape at the adapter edge; no SDK object ever persists.
-#: PEP 695 recursive alias: the old-style ``JsonValue = list["JsonValue"]``
-#: assignment form recurses infinitely under Pydantic 2.x schema generation,
-#: so this requires pydantic>=2.11.
+#: Recursive JSON value.  Provider objects are converted to this shape at
+#: the adapter edge so no SDK object ever persists or crosses the boundary.
+#:
+#: Defined here because neither Python nor Pydantic ships one: ``json`` is a
+#: (de)serializer, not a type; ``typing.Any`` validates nothing; and the
+#: old-style recursive alias (``JsonValue = list["JsonValue"]``) recurses
+#: infinitely under Pydantic 2.x schema generation.  The PEP 695 ``type``
+#: statement is the one recursive form Pydantic supports as a field
+#: annotation, which is why this module requires pydantic>=2.11.
 type JsonValue = None | bool | int | float | str | list["JsonValue"] | dict[str, "JsonValue"]
 
 #: Version tag mixed into every opaque replay key: when the normalization
@@ -95,12 +99,22 @@ RedactionClass = Literal["opaque", "plain_reasoning"]
 
 
 class ProviderIdentity(BaseModel):
-    """Versioned provider identity, independent of transport routing strings.
+    """Logical provider identity, independent of transport routing strings.
 
-    endpoint_id and account_scope
-    are non-secret fingerprints (never raw credentials).  transport is
-    *not* part of replay compatibility, so it is excluded from
-    derive_opaque_replay_key.
+    Owned by NOOA rather than borrowed from a transport library (litellm or a
+    successor) because the two answer different questions. litellm parses a
+    routing string to pick *where to send the request*; that parsing is
+    transport-specific, mutable across library versions, and confuses gateway
+    routes with providers — an ``openai/`` prefix in a routed id is not
+    evidence the model is served by OpenAI. This identity answers *who
+    produced a stored artifact and who may receive it back*: it is stamped
+    once at the adapter edge, persisted alongside captured reasoning, and
+    compared long after the original request. Making that durable,
+    provider-independent, and testable in isolation is why it lives here.
+
+    endpoint_id and account_scope are non-secret fingerprints (never raw
+    credentials).  transport is *not* part of replay compatibility, so it is
+    excluded from derive_opaque_replay_key.
     """
 
     provider: str  # openai, anthropic, moonshot, zai, ...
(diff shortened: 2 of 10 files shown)


furgalep:
## Review round complete — fresh-eyes subagent review + all CodeRabbit comments addressed

Pushed `cf0fc2a`. Two review passes ran: **(1)** an independent fresh-eyes subagent review (read-only, verified claims against live code, dispatched alongside the CodeRabbit pass), and **(2)** the CodeRabbit findings — 4 actionable + 1 nitpick. All addressed:

### Blocker (fresh-eyes — CodeRabbit missed this one)
**Compat-group members derived different keys.** The key payload mixed raw `model` alongside `compat_group`, so `gpt-5.5` and `gpt-5.6-sol` (same declared group) derived *different* keys — the group was decorative, contradicting D-04 rule 2 ("groups exist to keep model-id churn out of the replay hot path"). **Fix:** the declared compat group is now the model dimension of the key; raw model ids are not in the payload. Added `test_same_compat_group_members_derive_equal_keys`. Golden digest regenerated (nothing consumes the key yet — this was the cheap moment, exactly as the reviewer noted).

### CodeRabbit findings
| Finding | Disposition |
|---|---|
| 🟠 Major: PEP 695 `type JsonValue` vs `pydantic>=2.5.0` floor | **Floor raised to `pydantic>=2.11.0`** in all four pyprojects. I first tried the alternative (old-style `Union` assignment alias) and it **fails under Pydantic 2.x with a RecursionError** during schema generation — so the alias-form fallback is not viable and the floor bump is the correct fix. Fresh-eyes confirmed the version matrix: 2.9 fails, 2.10+ works. |
| 🟡 Canonicalize `provider` in `derive_opaque_replay_key` | **Fixed** — alias map applied inside derivation (`"zai"`/`"claude"` derive the same key scope as `"glm"`/`"anthropic"`), matching the docstring. |
| 🟡 Case normalization read-side only | **Fixed on the write side** — `register_compat_group` lowercases model members; `register_reasoning_capabilities` stores under `provider.lower()`. |
| 🟡 Golden digest never checked against derivation | **Fixed** — the round-trip test now asserts the frozen digest is reproducible via `derive_opaque_replay_key` (this immediately caught the legitimate payload change from the blocker fix). |
| 🧹 Nitpick: restore capability catalog after override test | **Fixed** with try/finally restore + a mixed-case reachability assertion. |

### Additional fresh-eyes fixes
- Bedrock dotted ids (`anthropic.claude-sonnet-4-5`) now strip the vendor head in normalization so both spellings of one identity derive the same key (test added — it was silently deriving `None` before).
- `api_style` normalized (lowercased) in the key payload so `"responses"`/`"Responses"` can't diverge.

### Validation
- `test_contracts.py`: **36 passed** (4 new: same-group key equality, Bedrock dotted form, digest reproducibility, mixed-case registration)
- Full unifiedllm suite: **388 passed**
- `ruff check` + `format --check`: clean

Noted-but-not-changed (documented for later PRs): overlapping compat groups resolve alphabetically; chained tier suffixes strip one level; the `bananaai` registry-restore fixture pattern will become a proper autouse fixture when later PRs share the process.

src/nooa/unifiedllm/__init__.py
```diff
@@ -30,6 +47,22 @@
 )
 
 __all__ = [
+    # Provider contracts (PR 1: identity and capability types, additive)
```
furgalep: Don't leak the PR number into the codebase.

src/nooa/unifiedllm/contracts.py
```diff
@@ -0,0 +1,534 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Transport-neutral provider identity and capability contracts.
+
+This module freezes the PR 1 / Packet A contract surface of the reasoning,
```
furgalep: Don't leak the PR number into the codebase. The comments should be durable comments that document the code and not be about our plan. Code should stand alone at every point.

src/nooa/unifiedllm/contracts.py
```diff
@@ -0,0 +1,534 @@
...
+replay, and observability program:
+
+- llm-reasoning-observability-program.md section 5 (target public contracts),
+- llm-reasoning-capture-replay.md section 7 / 7.1 (compatibility and D-04
+  key derivation rules) and section 12 / 12.1 (capability profile, D-07),
+- llm-program-execution-backlog.md packets C-01 (ProviderIdentity) and
+  C-02 (capability profile / effort map).
```
furgalep: Please go through all the comments and make sure that they are a reflection of the code that's implemented and not about our development plan.

src/nooa/unifiedllm/contracts.py
```diff
@@ -0,0 +1,534 @@
...
+    "register_compat_group",
+    "register_reasoning_capabilities",
+]
+
+#: NOOA-owned recursive JSON value type.  Provider objects are converted to
+#: this shape at the adapter edge; no SDK object ever persists (design section 4).
+#: PEP 695 recursive alias: the old-style ``JsonValue = list["JsonValue"]``
+#: assignment form recurses infinitely under Pydantic 2.x schema generation,
+#: so this requires pydantic>=2.11 (the floor this PR raises).
+type JsonValue = None | bool | int | float | str | list["JsonValue"] | dict[str, "JsonValue"]
```
furgalep, line 73: I'm not sure why we have to define this. Isn't there a built-in type or something else?

src/nooa/unifiedllm/contracts.py
```diff
@@ -0,0 +1,534 @@
...
+    AUTO = "auto"
+    NATIVE_ONLY = "native_only"
+    TEXT_CONTEXT = "text_context"  # force plain text to labeled context
+
+
+#: Redaction class is attached independently of the payload kind (backlog R-01).
+RedactionClass = Literal["opaque", "plain_reasoning"]
+
+
+class ProviderIdentity(BaseModel):
```
furgalep, line 101: Does this need to live in NOOA? Isn't there something in Light LLM or any LLM that takes this responsibility? I can see the one benefit of making this provider-independent, meaning that we could switch from Light LLM to any LLM and back, or pick another library later on, but it's a lot of complexity to onboard, so I just want to make sure it's worth it.

src/nooa/unifiedllm/contracts.py
```diff
@@ -0,0 +1,511 @@
...
+    "derive_opaque_replay_key",
+    "get_reasoning_capabilities",
+    "parse_model_string",
+    "register_compat_group",
+    "register_reasoning_capabilities",
+]
+
+#: NOOA-owned recursive JSON value type.  Provider objects are converted to
+#: this shape at the adapter edge; no SDK object ever persists (design section 4).
+type JsonValue = None | bool | int | float | str | list["JsonValue"] | dict[str, "JsonValue"]
```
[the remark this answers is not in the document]
furgalep: Fixed in 5af5155: raised the pydantic floor to `>=2.11.0` in all four workspace pyprojects. The alternative (old-style `Union` assignment alias) was verified to fail - it recurses infinitely under Pydantic 2.x schema generation (`RecursionError`), so the floor bump is the only viable fix. The lockfile already resolves to 2.13.4, so no environment changes in practice.

src/nooa/unifiedllm/contracts.py
```diff
@@ -0,0 +1,511 @@
...
+        return None
+    parsed = parse_model_string(model)
+    if not parsed.model:
+        return None
+    group = compat_group_for(provider, parsed.model, groups=compat_groups)
+    if group is None:
+        return None
+    payload = {
+        "key_version": OPAQUE_REPLAY_KEY_VERSION,
+        "provider": provider.lower(),
```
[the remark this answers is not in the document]
furgalep: Fixed in 5af5155: `derive_opaque_replay_key` now canonicalizes the caller-supplied provider through the alias map before use, so "zai"/"claude" derive the same key scope as "glm"/"anthropic" - matching the docstring's promise that alias stripping happens inside the derivation.

src/nooa/unifiedllm/contracts.py
```diff
@@ -0,0 +1,511 @@
...
+_REASONING_CAPABILITIES: dict[str, ReasoningCapabilities] = dict(DEFAULT_REASONING_CAPABILITIES)
+
+
+def register_reasoning_capabilities(provider: str, capabilities: ReasoningCapabilities) -> None:
+    """Register (or explicitly override) capabilities for a provider.
+
+    Explicit overrides take precedence over the generated/default catalog
+    (PR 1 test requirement).
+    """
+    _REASONING_CAPABILITIES[provider] = capabilities
```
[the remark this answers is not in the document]
furgalep: Fixed in 5af5155, both sites: `register_reasoning_capabilities` stores under `provider.lower()`, and `register_compat_group` lowercases model members, so mixed-case registrations are reachable by the case-insensitive lookups. A test asserts a mixed-case registration is retrievable.

tests/unifiedllm/test_contracts.py
```diff
@@ -0,0 +1,320 @@
...
+        "NormalizedModel": NormalizedModel.model_json_schema(),
+    }
+
+
+def test_golden_fixture_examples_round_trip() -> None:
+    golden = json.loads(GOLDEN.read_text())
+    examples = golden["examples"]
+    identity = ProviderIdentity.model_validate(examples["provider_identity"])
+    assert identity.provider == "openai"
+    assert identity.opaque_replay_key is not None
```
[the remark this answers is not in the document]
furgalep: Applied in 5af5155: the golden round-trip test now asserts the frozen digest equals `derive_opaque_replay_key(...)` recomputed from the frozen identity fields. This immediately proved its worth - it caught a legitimate derivation change during review, which is exactly the drift it guards.

src/nooa/unifiedllm/contracts.py
```diff
@@ -0,0 +1,537 @@
...
+    provider_key = provider.lower()
+    model_key = model.lower()
+    matches = sorted(
+        (
+            group
+            for group in registry.values()
+            if group.provider.lower() == provider_key and model_key in group.models
+        ),
+        key=lambda group: group.name,
+    )
```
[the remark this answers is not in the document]
furgalep: Fixed in 0dfc14c: compat_group_for now lowercases each injected group's members at lookup time (same handling as register_compat_group's write-time normalization), so a declared mixed-case model resolves through caller-supplied mappings and key derivation. Added a test covering a mixed-case injected group through both compat_group_for and derive_opaque_replay_key.

src/nooa/unifiedllm/__init__.py
furgalep: Done in 5af5155 - the `__init__.py` comment now reads "Provider contracts (identity and capability types)" with no PR reference. Verified with a sweep: no PR numbers, plan references, or process narration remain in any file in this PR.

src/nooa/unifiedllm/contracts.py
furgalep: Done in 5af5155 - all module docstrings and comments were rewritten to document the code on its own terms (what it does and why), and the module docstring now explains the routing-string problem it solves rather than citing design docs. Sweep confirms no plan/PR references remain in any file.

src/nooa/unifiedllm/contracts.py
furgalep: Done - audited every docstring and comment in the PR (contracts.py, test_contracts.py, __init__.py, and the fixture). All now describe the implemented behavior; the only remaining references to version bumps or normalization rules describe what the code itself does. Commit messages were also squashed to a single self-contained message.

src/nooa/unifiedllm/contracts.py
furgalep: Answered in the code in 0dfc14c - the docstring now explains why it's defined here: `json` is a (de)serializer, not a type; `typing.Any` validates nothing; and the old-style recursive alias (`JsonValue = list["JsonValue"]`) recurses infinitely under Pydantic 2.x schema generation (verified: RecursionError). The PEP 695 `type` statement is the one recursive form Pydantic supports as a field annotation - which is also what necessitated the pydantic>=2.11 floor.

src/nooa/unifiedllm/contracts.py
furgalep: Answered in the code in 0dfc14c - the ProviderIdentity docstring now states the ownership rationale directly: litellm parses a routing string to pick *where to send the request*; that parsing is transport-specific and mutable across library versions, and it conflates gateway routes with providers (an `openai/` prefix in a routed id is not evidence the model is served by OpenAI). This identity answers a different question - *who produced a stored artifact and who may receive it back* - which is stamped once at the adapter edge, persisted alongside captured reasoning, and compared long after the original request, potentially under a different transport. That durability requirement, plus the provider-independent switch benefit you named, is why it lives here rather than being borrowed from either library.

src/nooa/unifiedllm/contracts.py
```diff
@@ -0,0 +1,631 @@
...
+    {"openai", "nvidia", "azure", "aws", "bedrock", "vertex_ai", "nvidia_nim", "huggingface"}
+)
+
+#: Free-tier / capacity / routing tier suffixes stripped before any comparison
+#: or hash (e.g. "kimi-k3:free", "deepseek-v4-flash:cloud", "o3:batch").
+_TIER_SUFFIXES: frozenset[str] = frozenset({"free", "cloud", "batch"})
+
+#: Provider aliases: a leading segment that names a vendor but is not the
+#: canonical logical provider.  Canonical names map to themselves.
+_PROVIDER_ALIASES: dict[str, str] = {
```
sklinglernv, line 251: CodeAct take: Could we clarify this boundary? NOOA should own durable logical identity and persisted replay contracts, while LiteLLM can supply model-level capability facts. Please keep parsing as an adapter fallback and make capability lookup model/API-style aware instead of provider-wide.

sklinglernv:
CodeAct take: I think the ownership split is mostly right. NOOA should own persisted reasoning records, replay policy/compatibility, and durable logical identity. LiteLLM cannot fully provide that identity: its get_llm_provider treats the leading openai/ route as the provider for the nested NVIDIA examples here.

Could we make capabilities model/API-style aware, though? LiteLLM already has model-level reasoning and effort metadata, while get_reasoning_capabilities("openai") applies one profile—including an Azure-specific backfill flag—to every OpenAI model and endpoint. I suggest using LiteLLM facts at the adapter edge, then layering small NOOA overrides for missing or incorrect data. Explicit adapter identity would also be preferable, with parse_model_string kept as a legacy fallback.

sklinglernv:
CodeAct take: A couple of small public-API gaps:

- Importing nooa.unifiedllm.contracts still runs the package initializer, which loads LiteLLM/OpenAI and triggers the model-map fetch. The source-text test misses this. Could we add a fresh-process import test and make this path genuinely lazy/lightweight?
- JsonValue is missing from contracts.__all__ and the package exports; RedactionClass is also missing from the package exports. Could we export them consistently, or mark them private if that is intentional?

sklinglernv:
CodeAct take: Could we make ReasoningRecord.payload strict JSON? Pydantic currently accepts NaN/infinity but serializes them as null, so the value changes after persistence. It also coerces bytes to text. Please reject non-finite floats and non-JSON types, with a round-trip test for accepted payloads.

src/nooa/unifiedllm/contracts.py
```diff
@@ -0,0 +1,605 @@
...
+    ("nv-embed", "nvidia"),
+    ("nv-rerank", "nvidia"),
+)
+
+
+def _resolve_family(segment: str) -> str | None:
+    """Return the logical provider for a final model-id segment, if known."""
+    lowered = segment.lower()
+    for marker, provider in _FAMILY_MARKERS:
+        if marker in lowered:
```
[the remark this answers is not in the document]
sklinglernv: CodeAct take: This still looks present on the current head. parse_model_string("some-unknown-vendor/nemotronish-v1") and "xnv-embedder-v1" both resolve to nvidia because line 346 still uses substring matching. Could we require identifier boundaries and add those two regression cases?

src/nooa/unifiedllm/contracts.py
```diff
@@ -0,0 +1,631 @@
...
+    Membership must be declared; a similar-but-undeclared model id fails
+    closed.  Groups are scoped per logical provider.
+    """
+    registry = _COMPAT_GROUPS if groups is None else groups
+    provider_key = provider.lower()
+    model_key = model.lower()
+    # Members are lowercased at comparison time so injected mappings match
+    # regardless of how they were constructed (register_compat_group also
+    # normalizes on write; this covers callers passing their own mapping).
+    matches = sorted(
```
sklinglernv, line 483: CodeAct take: Could we reject overlapping groups or define real override precedence here? If an adapter registers zzz-openai-next with {gpt-5.6-sol, gpt-next}, this lookup still picks openai-gpt-5 for gpt-5.6-sol but zzz-openai-next for gpt-next. The two declared group members then get different replay keys. A regression test with a differently named overlapping group would catch this.

furgalep:
Superseded by the smaller implementation stack: #268 → #312 → #310 → #311. The replacement keeps provider-specific capture/replay inside UnifiedLLM, stores one opaque sidecar on canonical LLMOutput, and avoids importing a provider catalog into core NOOA.

PR #307, opened 2026-09-07
feat(llm): config-driven provider declaration and catalog drift check

furgalep:
## Summary

Let an `llm_config.yaml` alias declare the two things an opaque enterprise/gateway model id cannot prove on its own: its **logical provider** and its **model compat group**.

### Why

Model ids reach the registry as routing strings, and some resolve to no provider at all — enterprise gateways rename models (`acme-internal-model-v2` that is really a wrapped GPT), so without a declaration the id carries no identity and replay compatibility fails closed. The declaration is the config's answer, with no code required:

```yaml
models:
  acme-internal-v2:
    model_name: openai/acme-internal-model-v2
    api_base: https://acme-gateway/v1
    provider: openai          # logical provider the id cannot prove
    compat_group: openai-gpt-5 # opaque replay boundary from the declared group
```

### What's in it

- **`ModelConfig`** gains typed optional `provider` / `compat_group` fields (`extra="allow"` passthrough unchanged); the registry schema documents them.
- **`get_llm_client`** resolves a declaring alias through the identity parser and attaches a `ProviderIdentity` to the client **as metadata only** — request parameters are untouched, and undeclared aliases behave exactly as before (no identity attached; consumers keep failing closed).
- **`nooa.unifiedllm.declaration`** translates the declaration into process-lifetime registrations (compat-group membership, reasoning-capability overrides) that override the built-in catalogs. Fail-closed rules: a declared provider that *contradicts* what the model string itself resolves raises rather than guessing; a group name owned by a different provider is rejected; the declared group is authoritative for the declaring alias's opaque replay key, so other groups claiming the same model cannot silently win; capability errors degrade with a warning (a wrong capability profile degrades handling; a wrong identity misroutes artifacts — that one fails loudly).
- **`scripts/refresh_model_id_corpora.py`** rebuilds the live corpus sections (OpenRouter public catalog; NVIDIA gateway when `NVIDIA_INFERENCE_API_KEY` is set) from their `/v1/models` endpoints and prints the resolution/misattribution summary. **`--check` exits 1 on drift without writing** — a one-command catalog-drift detector. Offline it exits non-zero with a clear message and never truncates the existing fixture; the provenance comment is excluded from drift so its fetch date does not trip the check.

### Notes

- Declared values are verified, never trusted: group extension keeps live-verified existing members (a declaration asserts one more member, not a replacement).
- The azure/vertex_ai/bedrock corpus sections (bundled-catalog provenance, no public endpoint) are preserved verbatim by the refresh.

### Tests

34 new tests in `tests/unifiedllm/test_declaration.py`: opaque alias with a declaration derives the group key; without, no key and no registrations; declared values override the catalog; lazy re-application on changed declarations; identity attached as metadata only (constructor kwargs recorded, no identity keys leak); refresh-script unit tests (offline non-zero exit with fixture untouched, `--check` drift exit 1, `--check` green against the current fixture).

`uv run pytest -q tests/unifiedllm tests/config` — **480 passed**. Live drift check: OpenRouter 430 ids, resolution 0.79, misattributed 0; NVIDIA gateway 90 verifiable ids, resolution 0.91, misattributed 0. Ruff check + format clean.

🤖🤖🤖

## Summary by CodeRabbit

- **New Features**
  - Added configurable provider identities and compatibility groups for supported model aliases.
  - Added reasoning-capability overrides for model configurations.
  - Client metadata now exposes the resolved provider identity without changing request routing.

- **Improvements**
  - Added tooling to refresh and validate model catalogs, with drift detection and safe handling of failed updates.
  - Expanded alias configuration validation and compatibility handling.

- **Tests**
  - Added coverage for provider declarations, metadata wiring, validation, and model-catalog refresh behavior.

### DIFF
--- a/scripts/refresh_model_id_corpora.py
+++ b/scripts/refresh_model_id_corpora.py
@@ -1,0 +1,271 @@
+#!/usr/bin/env python
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Refresh the model-id corpora fixture from the live model catalogs.
+
+Fetches the public OpenRouter catalog (``/v1/models``) and, when
+``NVIDIA_INFERENCE_API_KEY`` is set, the NVIDIA inference gateway catalog,
+rebuilds the ``openrouter`` / ``nvidia_gateway`` sections of
+``tests/unifiedllm/fixtures/model_id_corpora.json``, and prints the
+resolution/misattribution summary for the fresh catalog.
+
+Offline-safe: with no network it exits non-zero with a clear message and
+never truncates the existing fixture (a failed fetch is never written).
+
+Usage::
+
+    uv run python scripts/refresh_model_id_corpora.py           # refresh
+    uv run python scripts/refresh_model_id_corpora.py --check   # report only
+
+``--check`` exits 1 on drift without writing.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import sys
+import urllib.error
+import urllib.request
+from datetime import date
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parent.parent
+FIXTURE = REPO / "tests/unifiedllm/fixtures/model_id_corpora.json"
+
+OPENROUTER_URL = "https://openrouter.ai/api/v1/models"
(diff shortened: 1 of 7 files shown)


furgalep:
Superseded together with #304 by the minimal #268 → #312 → #310 → #311 stack. Registry capability declarations and generated provider catalogs are intentionally deferred; the current stack solves safe capture, persistence, replay, and cross-model text demotion without them.

PR #308, opened 2026-09-07
fix(tools): keep files newline-terminated after replace() at EOF

furgalep:
## Problem

Two related defects in the file-edit path:

1. **Real data bug:** `replace(match, new_text)` whose region reaches end-of-file silently stripped the file's final newline. Re-termination was gated on after-lines existing, so an unterminated replacement at EOF left the file without its trailing newline (git then flags every such file as 'no newline at end of file').
2. **Noisy/incorrect output:** fragment diffs (the `replace()` previews shown as `oo update` activity) emitted git's `\ No newline at end of file` marker whenever the edit snippet stopped before the file's last newline — i.e. almost every edit — even though the file on disk was newline-terminated.

## Fix

- `ShellTools.replace` (Match form) now re-terminates the replacement when the region reaches EOF **and** the original content ended with a newline. A file that genuinely lacks a final newline keeps lacking one — no spurious additions.
- `_edit_diff` gained a `whole_file` flag (default off). The EOF marker is now emitted only for whole-file overwrites (`write_file`), where it is accurate; fragment diffs keep the line newline-terminated (so it can't glue to the next marker) but no longer claim a file state that does not exist.

## Testing

- 2 new regression tests for the EOF cases (keeps newline; preserves genuine lack of one)
- New test that fragment diffs do not emit the marker; marker test updated to the `whole_file=True` path
- 4 existing tests updated where they had pinned the old stripping/marker behavior as intentional
- `tests/tools/test_shell_tools_modern_behavior.py` + `packages/nooa-cli/tests/test_coding_activity.py`: 34 passed; ruff check + format clean

Note: two of the updated tests previously documented the stripping as deliberate ('Nothing follows the region, so no newline is added'). Per discussion this was a bug, not a feature — the intent of the re-termination logic was to preserve the file's termination.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved trailing newlines when replacing file content at the end of a file.
  - Maintained files without a final newline without adding one.
  - Improved diff output to distinguish complete-file changes from partial content changes, including accurate end-of-file markers.

### DIFF
--- a/packages/nooa-cli/src/nooa_cli/coding/activity.py
+++ b/packages/nooa-cli/src/nooa_cli/coding/activity.py
@@ -69,8 +69,15 @@ def _edit_diff(
     old_text: str,
     new_text: str,
     start_line: int | None,
+    *,
+    whole_file: bool = False,
 ) -> tuple[str, bool]:
-    """Return a bounded, line-oriented unified diff and its completeness."""
+    """Return a bounded, line-oriented unified diff and its completeness.
+
+    ``whole_file`` marks both texts as complete file contents, where a missing
+    final newline is a real file state worth reporting. Fragment diffs stop
+    mid-file, so the EOF marker there was noise.
+    """
     if _diff_input_is_too_large(old_text) or _diff_input_is_too_large(new_text):
         return _omitted_diff(path, "file content exceeds the safe diff preview limit")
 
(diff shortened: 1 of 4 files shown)


alessiodevoto (approved):
LGTM!

PR #310, opened 2026-09-08
feat(llm): retain compatibility-scoped OpenAI reasoning state

furgalep:
## Summary

- capture OpenAI encrypted reasoning from native Responses calls and LiteLLM's Chat bridge
- persist one versioned opaque envelope on the canonical `LLMResponse`
- bind Chat and Responses state to the exact public assistant carrier that produced it
- replay only when the effective provider, API style, exact model, and carrier match
- reconstruct native Responses reasoning/function-call ordering without duplicating public tool calls
- keep opaque state out of generic JSON, unrelated providers, logs, and journal diagnostics
- preserve provider-exposed reasoning text as the safe fallback
- reject request overrides that could bypass replay validation

## Why this layer exists

OpenAI reasoning models return provider-owned state that should accompany its matching assistant turn on later requests. Dropping it weakens multi-turn reasoning; sending it beside a changed turn or to a different provider is unsafe.

#312 gives NOOA one durable, provider-independent assistant-turn record: `LLMResponse`. Public text, tool calls, reasoning text, and usage live on that record; provider objects do not. This PR adds the smallest OpenAI-specific adapter at the UnifiedLLM edge. It stores encrypted state as one opaque `llm_state` envelope, then reconstructs provider-native input only for a compatible request.

The public carrier is part of compatibility. Opaque state returned beside a particular assistant message is valid only while that message's text and ordered tool calls (ID, name, and exact arguments) are unchanged. This matters because formatters and middleware expose a public message IR and may legitimately edit it. The provider blob must not silently follow an edited turn.

Compatibility deliberately uses provider, API style, and exact resolved model—not gateway URL or credentials. Transport changes do not by themselves change the provider wire format. Exact-model matching is conservative until a later capability layer can declare verified model-family compatibility.

## Code walkthrough: what changed and why

1. **One durable opaque envelope — `replay_state.py`.**  
   **What:** OpenAI Chat `reasoning_items` and native Responses reasoning items are detached once into `LLMResponse.llm_state`. The same private envelope retains Responses message order and optional phase when flat public text/tool calls cannot reproduce them.  
   **Why:** SQLite/session resume needs the provider state, while the event IR must not learn or expose OpenAI's schema.

2. **Exact Chat carrier binding.**  
   **What:** Chat capture stores one deterministic SHA-256 fingerprint of assistant content plus ordered tool-call ID, name, and exact argument string, not a second copy of those fields. Changed text, calls, order, duplicate IDs, missing/malformed carrier data, and edits to reasoning-only turns all withhold the opaque state.  
   **Why:** a signed/encrypted artifact cannot safely be attached to a semantically different assistant turn.

3. **Exact Responses reconstruction.**  
   **What:** capture uses the SDK's exact text concatenation. Saved function-call slots retain an argument fingerprint rather than duplicating the argument string; message slots retain text needed to reconstruct provider message boundaries. Replay validates the public aggregate and then walks the saved ordering once, restoring interleaved reasoning, messages, calls, and assistant phases. Metadata is retained even without encrypted reasoning when flattening would lose message structure.  
   **Why:** a canonical string cannot express separate commentary/final-answer messages or their position among tool calls. Those provider-required distinctions must survive resume for correct continuation and cache reuse, without adding OpenAI fields to NOOA's public IR. Exact carrier validation prevents replay after an edit. A whole-turn check withholds state if any output item or message block is unsupported, rather than replaying reasoning against a partial turn.

4. **Private formatter carrier.**  
   **What:** ordinary provider messages/items carry `llm_state`, reasoning text, and batch identity only as private attributes on an in-memory dict subclass. JSON sees only the public mapping; middleware borrows immutable event-owned state.  
   **Why:** provider-independent renderers can preserve one logical turn without copying large blobs or leaking control metadata onto the wire.

5. **Small compatibility key.**  
   **What:** LiteLLM resolves provider/model identity, and the envelope binds provider + API style + exact model. Unknown or incompatible routes fail closed.  
   **Why:** model routing strings are not a safe substring-based identity system, and opaque state must never be broadcast speculatively.

6. **Portable fallback.**  
   **What:** compatible OpenAI state is restored exactly; otherwise #312's shared path replays provider-exposed reasoning as ordinary assistant text. Opaque-only incompatible turns do not become empty assistant messages.  
   **Why:** readable reasoning is useful context across models; encrypted protocol state is not.

7. **Bypass and privacy protection.**  
   **What:** direct `messages`/`input`, nested `extra_body`, effective-model overrides, HTTP logs, and journal fallback paths are validated or scrubbed. The shared scrubber only attempts JSON parsing for object/array prefixes, preserves Unicode when reserializing redacted JSON, and leaves clean JSON formatting untouched.  
   **Why:** replay safety must hold at the final serialized request and every diagnostic boundary, not only inside the formatter. OTLP is an observational view, never an agent replay archive: readable reasoning is exported, while opaque state stays in the session archive and is redacted from telemetry.

8. **Honor per-call transport changes.**  
   **What:** CompletionClient omits its constructor-bound OpenAI SDK client when a call changes routing or credentials, letting LiteLLM select the current endpoint/key. ResponsesClient retains its generic HTTP handler: unlike the SDK client, that handler does not bind a URL or API key. Normal calls retain their owned transport.  
   **Why:** a supplied OpenAI SDK client keeps its original URL and credentials even when different values are passed alongside it. Responses handlers instead receive URL/authentication per request, so dropping their transport would unnecessarily discard the configured pool. Twelve real-LiteLLM/MockTransport sync/async Responses regressions verify URL, Authorization, model, and default → override → default behavior. Completion override calls use LiteLLM's default HTTP pool settings.

9. **Remove redundant request and logging copies.**  
   **What:** discard private sidecars and rejected raw state before copying public messages; detach public nested data once in the branch that owns the outgoing request; demote reasoning in that already-owned batch; use the shared HTTP scrubber without a second recursive redaction pass.  
   **Why:** replay must not clone opaque history only to discard the clone, and fallback must not copy the same public message two or three times. The remaining public copy protects caller-owned nested data from request mutations. Fifteen ownership regressions count copies and check that inputs remain unchanged.

10. **Validate shared OpenAI items and version compact bindings.**  
   **What:** Chat and Responses use the same reasoning-item validator: each item must be a reasoning mapping with nonempty encrypted content. Validation-only public content becomes a fingerprint; opaque data is still borrowed during replay. The private envelope is now version 2; version-1 draft envelopes warn and demote available readable reasoning instead of being replayed.  
   **Why:** malformed items such as `[123]` must not reach the adapter, and persisting a large tool argument twice wastes archive space. The version change makes the draft-format incompatibility explicit without adding a second replay implementation. Original archives are not modified.

## Non-goals

- Anthropic signed thinking and Gemini thought signatures (#311)
- model-family capability declarations
- provider-rejection retry ladders
- explicit prompt-cache boundary policy (#313)
- TUI or reasoning-effort controls

## Validation

- scrubber/transport follow-up (2026-09-11): **717 passed**, 5 deselected, across UnifiedLLM and tracing. Twelve Responses transport cases already passed before the production change; seven newly failing scrubber cases pass after it, with two additional formatting/fallback guards. Ruff and targeted Pyright pass. HTTP is mocked; no inference calls.

- fingerprint/item-validation follow-up (2026-09-11): **696 passed**, 5 deselected, across UnifiedLLM and tracing; 30 new cases cover malformed capture/replay, JSON/SQLite round-trips, argument/text storage size, public edits, deterministic fingerprints, and prior draft-version fallback. A synthetic 100 KB tool argument adds **less than 1 KB** of replay metadata on both Chat and Responses. Ruff and targeted Pyright pass; no inference calls.

- copy-reduction round (2026-09-11): **511 passed** across UnifiedLLM and journal/scrubber tests; 15 new ownership cases prove zero opaque-sidecar copies and one public nested copy in the exercised replay paths; Ruff and targeted Pyright pass

- latest regression round: real SDK → JSON/SQLite → serialized HTTP tests cover multi-block text, ordered/phased messages, summary-only reasoning, unsupported whole-turn rejection, and eligible Anthropic cache-marker fallback

- #310 affected replay/formatter/tracing set: **160 passed**; six new sync/async transport-override cases also pass through actual LiteLLM/SDK dispatch
- prior combined-stack full suite (before this follow-up): **7,364 passed**, 10 skipped, 239 deselected, 3 expected xfails
- regressions cover Chat text/tool/order/duplicate/malformed/state-only carrier changes; Responses splits/reorders/edits; effective per-call models; JSON resume; Relay; reserved payload overrides; and redaction
- Ruff, targeted Pyright, DCO trailers, and diff checks pass
- top-stack live NVIDIA-routed GPT-5.6-sol retained encrypted state through a real SQLite close/reopen and fresh client; the stable HTTP request stayed identical and the resumed call reported 6,162 cached input tokens

Built on merged #312.

## Summary by CodeRabbit

* **New Features**
  * Improved preservation and replay of provider reasoning across chat and Responses interactions.
  * Added per-call model overrides with routing, caching, metrics, and tracking based on the selected model.
  * Added support for configured Responses API storage and encrypted-reasoning settings.
  * Improved handling of native response objects and dictionary-based payloads.

* **Security & Privacy**
  * Expanded redaction of private reasoning and encrypted content in logs, traces, and outgoing messages.

* **Breaking Changes**
  * Removed client-level replay scope configuration.

### DIFF
--- a/src/nooa/tracing/_secret_scrubber.py
+++ b/src/nooa/tracing/_secret_scrubber.py
@@ -51,7 +51,7 @@ _SENSITIVE_KEYS = frozenset(
 # Provider-owned replay state is not a user credential, but it has the same
 # telemetry rule: it may go back to its issuer and nowhere else. Provider
 # adapters add their exact wire keys here as support is introduced.
-_OPAQUE_PROVIDER_STATE_KEYS = frozenset({"encrypted_content"})
+_OPAQUE_PROVIDER_STATE_KEYS = frozenset({"encrypted_content", "nooa_llm_state"})
 
 
 def _is_sensitive_key(key: Any) -> bool:
--- a/src/nooa/tracing/_secret_scrubber.py
+++ b/src/nooa/tracing/_secret_scrubber.py
@@ -15,6 +15,7 @@ Usage::
     provider.add_span_processor(SecretScrubSpanProcessor(inner_processor))
 """
 
+import json
 import logging
 import re
 import threading
@@ -47,12 +48,30 @@ _SENSITIVE_KEYS = frozenset(
     }
 )
 
+# Provider-owned replay state is not a user credential, but it has the same
+# telemetry rule: it may go back to its issuer and nowhere else. Provider
+# adapters add their exact wire keys here as support is introduced.
+_OPAQUE_PROVIDER_STATE_KEYS = frozenset({"encrypted_content"})
+
 
 def _is_sensitive_key(key: Any) -> bool:
     normalized = re.sub(r"[^a-z0-9]+", "_", str(key).lower()).strip("_")
     return normalized in _SENSITIVE_KEYS or any(
         normalized.endswith(f"_{sensitive}") for sensitive in _SENSITIVE_KEYS
     )
+
+
+def _is_opaque_provider_state_key(key: Any) -> bool:
+    normalized = re.sub(r"[^a-z0-9]+", "_", str(key).lower()).strip("_")
+    return normalized in _OPAQUE_PROVIDER_STATE_KEYS
+
+
+def _redact_key(key: Any) -> str | None:
+    if _is_sensitive_key(key):
+        return "sensitive_key"
+    if _is_opaque_provider_state_key(key):
+        return "opaque_provider_state"
+    return None
 
 
 # ---------------------------------------------------------------------------
@@ -214,19 +233,29 @@ def scrub_value(value: Any) -> tuple[Any, int]:
     number of secrets redacted within this value.
     """
     if isinstance(value, str):
-        return scrub_string(value)
+        # OpenInference records LLM inputs as JSON string span attributes.
+        # Parse only valid JSON and serialize it again only if the recursive
+        # walk found additional keyed state to redact.
+        try:
+            decoded = json.loads(value)
+        except (json.JSONDecodeError, TypeError):
+            return scrub_string(value)
+        decoded, json_count = scrub_value(decoded)
+        if json_count:
+            return json.dumps(decoded, separators=(",", ":")), json_count
+        return value, 0
     if isinstance(value, dict):
-        scrubbed: dict[Any, Any] = {}
+        scrubbed_mapping: dict[Any, Any] = {}
         count = 0
         for key, item in value.items():
-            if _is_sensitive_key(key):
-                scrubbed[key] = REDACTED
-                stats.record("sensitive_key")
+            if reason := _redact_key(key):
+                scrubbed_mapping[key] = REDACTED
+                stats.record(reason)
                 count += 1
             else:
-                scrubbed[key], n = scrub_value(item)
+                scrubbed_mapping[key], n = scrub_value(item)
                 count += n
-        return scrubbed, count
+        return scrubbed_mapping, count
     if isinstance(value, (list, tuple)):
         new_items = []
         count = 0
@@ -268,9 +297,9 @@ try:
                 redacted_count = 0
 
                 for key, value in span.attributes.items():
-                    if _is_sensitive_key(key):
+                    if reason := _redact_key(key):
                         new_value, n = REDACTED, 1
-                        stats.record("sensitive_key")
+                        stats.record(reason)
                     else:
                         new_value, n = scrub_value(value)
                     scrubbed[key] = new_value
--- a/src/nooa/tracing/_secret_scrubber.py
+++ b/src/nooa/tracing/_secret_scrubber.py
@@ -234,15 +234,17 @@ def scrub_value(value: Any) -> tuple[Any, int]:
     """
     if isinstance(value, str):
         # OpenInference records LLM inputs as JSON string span attributes.
-        # Parse only valid JSON and serialize it again only if the recursive
-        # walk found additional keyed state to redact.
+        # Only parse objects/arrays. Matching the prefix avoids allocating a
+        # stripped copy of a potentially large tool output just to inspect it.
+        if not re.match(r"\s*[\[{]", value):
+            return scrub_string(value)
         try:
             decoded = json.loads(value)
         except (json.JSONDecodeError, TypeError):
             return scrub_string(value)
         decoded, json_count = scrub_value(decoded)
         if json_count:
-            return json.dumps(decoded, separators=(",", ":")), json_count
+            return json.dumps(decoded, separators=(",", ":"), ensure_ascii=False), json_count
         return value, 0
     if isinstance(value, dict):
         scrubbed_mapping: dict[Any, Any] = {}
--- a/src/nooa/unifiedllm/replay_state.py
+++ b/src/nooa/unifiedllm/replay_state.py
@@ -138,13 +138,78 @@ def _matching_payload(state: Any, scope: str | None, state_format: str) -> dict 
     return cast(dict[str, Any], state["payload"])
 
 
+def _chat_public_carrier(message: Any) -> dict[str, Any] | None:
+    """Project the public assistant data to which opaque Chat state is bound."""
+    if _field(message, "role") != "assistant":
+        return None
+    raw_calls = _field(message, "tool_calls")
+    if raw_calls is None:
+        raw_calls = []
+    if not isinstance(raw_calls, list):
+        return None
+
+    calls: list[dict[str, str]] = []
+    for call in raw_calls:
+        function = _field(call, "function")
+        call_id = _field(call, "id")
+        name = _field(function, "name")
+        arguments = _field(function, "arguments")
+        if isinstance(arguments, dict):
+            arguments = json.dumps(arguments)
+        if not all(isinstance(value, str) for value in (call_id, name, arguments)):
+            return None
+        calls.append({"id": call_id, "name": name, "arguments": arguments})
+
+    content = _field(message, "content")
+    if content is not None and not isinstance(content, str):
+        return None
+    content = (content or None) if calls else (content or "")
+    return {"content": content, "tool_calls": calls}
+
+
+def _valid_chat_payload(payload: dict[str, Any]) -> bool:
+    if set(payload) - {"reasoning_items", "carrier", "state_only"}:
+        return False
+    items = payload.get("reasoning_items")
+    carrier = payload.get("carrier")
+    if not isinstance(items, list) or not items or not isinstance(carrier, dict):
+        return False
+    if set(carrier) != {"content", "tool_calls"}:
+        return False
+    if carrier.get("content") is not None and not isinstance(carrier.get("content"), str):
+        return False
+    calls = carrier.get("tool_calls")
+    if not isinstance(calls, list) or not all(
+        isinstance(call, dict)
+        and set(call) == {"id", "name", "arguments"}
+        and all(isinstance(call.get(key), str) for key in ("id", "name", "arguments"))
+        for call in calls
+    ):
+        return False
+    if len({call["id"] for call in calls}) != len(calls):
+        return False
+    if "state_only" in payload and payload["state_only"] is not True:
+        return False
+    return (payload.get("state_only") is True) == (carrier == {"content": "", "tool_calls": []})
+
+
 def capture_chat_state(message: Any, scope: str | None) -> dict | None:
     items = _field(message, "reasoning_items")
     if not isinstance(items, list) or not items:
         return None
-    payload: dict[str, Any] = {"reasoning_items": [opaque_item(item) for item in items]}
+    carrier = _chat_public_carrier(message)
+    if carrier is None:
+        logger.warning("Discarding OpenAI Chat reasoning state with a malformed carrier.")
+        return None
+    payload: dict[str, Any] = {
+        "reasoning_items": [opaque_item(item) for item in items],
+        "carrier": carrier,
+    }
     if not _field(message, "content") and not _field(message, "tool_calls"):
         payload["state_only"] = True
+    if not _valid_chat_payload(payload):
+        logger.warning("Discarding malformed OpenAI Chat reasoning state.")
+        return None
     return _envelope(scope, _CHAT_FORMAT, payload)
 
 
@@ -278,12 +343,24 @@ def prepare_chat_messages(messages: list[dict[str, Any]], scope: str | None) -> 
         message.pop(LLM_STATE_KEY, None)
         message.pop("reasoning_items", None)
         payload = _matching_payload(state, scope, _CHAT_FORMAT)
-        if payload and isinstance(payload.get("reasoning_items"), list):
+        restored = False
+        if payload is not None and not _valid_chat_payload(payload):
+            logger.warning(
+                "Opaque Chat reasoning state is malformed; portable reasoning text "
+                "will be replayed instead."
+            )
+        elif payload is not None and _chat_public_carrier(message) != payload["carrier"]:
+            logger.warning(
+                "Opaque Chat reasoning state was not replayed because its public assistant "
+                "carrier changed; portable reasoning text will be replayed instead."
+            )
+        elif payload is not None:
             message["reasoning_items"] = payload["reasoning_items"]
-        else:
+            restored = True
+        if not restored:
             demote_reasoning_text(message, reasoning)
         if (
-            payload is None
+            not restored
             and state is not None
             and not reasoning
             and message.get("role") == "assistant"
--- a/src/nooa/unifiedllm/replay_state.py
+++ b/src/nooa/unifiedllm/replay_state.py
@@ -118,7 +118,7 @@ def _envelope(scope: str | None, state_format: str, payload: dict[str, Any]) -> 
         "version": _STATE_VERSION,
         "scope": scope,
         "format": state_format,
-        "payload": copy.deepcopy(payload),
+        "payload": payload,
     }
 
 
@@ -184,7 +184,7 @@ def prepare_chat_messages(messages: list[dict[str, Any]], scope: str | None) -> 
     """Strip private/raw state and restore only a matching Chat payload."""
     prepared: list[dict[str, Any]] = []
     for original in messages:
-        state = copy.deepcopy(carried_state(original))
+        state = carried_state(original)
         reasoning = carried_reasoning(original)
         message = copy.deepcopy(dict(original))
         message.pop(LLM_STATE_KEY, None)
@@ -237,7 +237,7 @@ def prepare_responses_batch(
     if not isinstance(items, list) or not isinstance(order, list):
         return demote_responses_batch(clean, state, reasoning)
     if payload.get("state_only") is True:
-        return [copy.deepcopy(item) for item in items if isinstance(item, dict)]
+        return [item for item in items if isinstance(item, dict)]
 
     calls = {
         item.get("call_id"): item
@@ -259,14 +259,14 @@ def prepare_responses_batch(
             if isinstance(index, int) and 0 <= index < len(items):
                 item = items[index]
                 if isinstance(item, dict):
-                    pending.append(copy.deepcopy(item))
+                    pending.append(item)
             continue
         if slot.get("type") == "function_call":
             call_id = slot.get("call_id")
             carrier = calls.get(call_id) if isinstance(call_id, str) else None
             if carrier is not None and isinstance(call_id, str):
                 replay.extend(pending)
-                replay.append(copy.deepcopy(carrier))
+                replay.append(carrier)
                 emitted_calls.add(call_id)
                 last_carrier_emitted = True
             else:
@@ -276,7 +276,7 @@ def prepare_responses_batch(
         if slot.get("type") == "message":
             if message is not None and not emitted_message:
                 replay.extend(pending)
-                replay.append(copy.deepcopy(message))
+                replay.append(message)
                 emitted_message = True
                 last_carrier_emitted = True
             else:
@@ -286,7 +286,7 @@ def prepare_responses_batch(
     if last_carrier_emitted:
         replay.extend(pending)
     replay.extend(
-        copy.deepcopy(item)
+        item
         for item in clean
         if not (
             item is message
--- a/src/nooa/unifiedllm/replay_state.py
+++ b/src/nooa/unifiedllm/replay_state.py
@@ -132,7 +132,9 @@ def _matching_payload(state: Any, scope: str | None, state_format: str) -> dict 
         or not isinstance(state.get("payload"), dict)
     ):
         return None
-    return cast(dict[str, Any], copy.deepcopy(state["payload"]))
+    # Event history owns this payload. Provider adapters may read and serialize
+    # it, but must not mutate caller input or require a per-request history copy.
+    return cast(dict[str, Any], state["payload"])
 
 
 def _is_state_only(state: Any, state_format: str) -> bool:
--- a/src/nooa/unifiedllm/replay_state.py
+++ b/src/nooa/unifiedllm/replay_state.py
@@ -67,17 +67,16 @@ def _normalized_endpoint(value: Any) -> str:
     return f"{parsed.scheme.lower()}://{parsed.netloc.lower()}{path}{query}"
 
 
-def _effective_endpoint(provider: str | None, configured: Any, resolved: Any) -> str:
-    endpoint = configured or getattr(litellm, "api_base", None)
-    if not endpoint and provider == "openai":
-        endpoint = (
-            os.getenv("OPENAI_BASE_URL")
-            or os.getenv("OPENAI_API_BASE")
-            or "https://api.openai.com/v1"
-        )
-    elif not endpoint and provider == "azure":
-        endpoint = os.getenv("AZURE_API_BASE")
-    return _normalized_endpoint(endpoint or resolved)
+def _uses_native_openai_endpoint(api_params: dict[str, Any]) -> bool:
+    endpoint = (
+        api_params.get("api_base")
+        or api_params.get("base_url")
+        or getattr(litellm, "api_base", None)
+        or os.getenv("OPENAI_BASE_URL")
+        or os.getenv("OPENAI_API_BASE")
+        or "https://api.openai.com/v1"
+    )
+    return _normalized_endpoint(endpoint) == "https://api.openai.com/v1"
 
 
 def replay_scope(
@@ -310,11 +309,9 @@ def add_encrypted_reasoning_include(api_params: dict[str, Any], scope: str | Non
         return
     if scope and scope.startswith("responses:azure:"):
         include.append(_ENCRYPTED_REASONING_INCLUDE)
-    elif scope and scope.startswith("responses:openai:"):
-        endpoint = _effective_endpoint(
-            "openai", api_params.get("api_base") or api_params.get("base_url"), None
-        )
-        if endpoint == "https://api.openai.com/v1":
-            include.append(_ENCRYPTED_REASONING_INCLUDE)
+    elif (
+        scope and scope.startswith("responses:openai:") and _uses_native_openai_endpoint(api_params)
+    ):
+        include.append(_ENCRYPTED_REASONING_INCLUDE)
     if include:
         api_params["include"] = include
--- a/src/nooa/unifiedllm/replay_state.py
+++ b/src/nooa/unifiedllm/replay_state.py
@@ -408,9 +408,10 @@ def prepare_chat_messages(messages: list[dict[str, Any]], scope: str | None) -> 
     for original in messages:
         state = carried_state(original)
         reasoning = carried_reasoning(original)
-        message = copy.deepcopy(dict(original))
+        message = dict(original)
         message.pop(LLM_STATE_KEY, None)
         message.pop("reasoning_items", None)
+        message = copy.deepcopy(message)
         payload = _matching_payload(state, scope, _CHAT_FORMAT)
         restored = False
         if payload is not None and not _valid_chat_payload(payload):
@@ -448,10 +449,12 @@ def _clean_responses_batch(batch: Any) -> list[dict[str, Any]]:
     for original in batch:
         if not isinstance(original, dict) or response_item_type(original) == "reasoning":
             continue
-        item = copy.deepcopy(original)
+        # Strip private sidecars and rejected wire state before detaching public
+        # content. Opaque state is borrowed separately after compatibility checks.
+        item = dict(original)
         item.pop(LLM_STATE_KEY, None)
         item.pop("reasoning_items", None)
-        clean.append(item)
+        clean.append(copy.deepcopy(item))
     return clean
 
 
--- a/src/nooa/unifiedllm/replay_state.py
+++ b/src/nooa/unifiedllm/replay_state.py
@@ -213,12 +213,24 @@ def capture_chat_state(message: Any, scope: str | None) -> dict | None:
     return _envelope(scope, _CHAT_FORMAT, payload)
 
 
-def _responses_message_text(item: Any) -> str:
+def responses_message_text(item: Any) -> str:
     content = _field(item, "content", [])
     if not isinstance(content, list):
         return ""
-    texts = [text for block in content if isinstance((text := _field(block, "text")), str)]
-    return "\n".join(texts)
+    texts = [
+        text
+        for block in content
+        if _field(block, "type") == "output_text"
+        and isinstance((text := _field(block, "text")), str)
+    ]
+    return "".join(texts)
+
+
+def responses_output_text(output: list[Any]) -> str:
+    """Match the SDK's separator-free aggregation of assistant text blocks."""
+    return "".join(
+        responses_message_text(item) for item in output if response_item_type(item) == "message"
+    )
 
 
 def _responses_call_slot(item: Any) -> dict[str, Any] | None:
@@ -243,7 +255,7 @@ def _valid_responses_payload(payload: dict[str, Any]) -> bool:
         return False
     items = payload.get("items")
     order = payload.get("order")
-    if not isinstance(items, list) or not items or not isinstance(order, list):
+    if not isinstance(items, list) or not isinstance(order, list) or not order:
         return False
     if any(
         not isinstance(item, dict)
@@ -274,7 +286,11 @@ def _valid_responses_payload(payload: dict[str, Any]) -> bool:
                 return False
             carriers.append(slot)
         elif slot_type == "message":
-            if set(slot) != {"type", "content"} or not isinstance(slot.get("content"), str):
+            if (
+                set(slot) - {"type", "content", "phase"}
+                or not isinstance(slot.get("content"), str)
+                or ("phase" in slot and slot["phase"] not in ("commentary", "final_answer"))
+            ):
                 return False
             carriers.append(slot)
         else:
@@ -284,7 +300,8 @@ def _valid_responses_payload(payload: dict[str, Any]) -> bool:
     if state_only not in (None, True):
         return False
     return (
-        sorted(indexes) == list(range(len(items)))
+        bool(items or carriers != _flatten_responses_carriers(carriers))
+        and sorted(indexes) == list(range(len(items)))
         and len({slot["call_id"] for slot in carriers if slot["type"] == "function_call"})
         == sum(slot["type"] == "function_call" for slot in carriers)
         and (state_only is True) == (not carriers)
@@ -305,13 +322,35 @@ def _public_responses_carriers(items: list[dict[str, Any]]) -> list[dict[str, An
     return carriers
 
 
+def _flatten_responses_carriers(carriers: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    """Project provider message boundaries into the canonical LLMResponse view."""
+    calls = [slot for slot in carriers if slot["type"] == "function_call"]
+    messages = [slot for slot in carriers if slot["type"] == "message"]
+    text = "".join(slot["content"] for slot in messages)
+    return (
+        [{"type": "message", "content": text}] if text or (messages and not calls) else []
+    ) + calls
+
+
 def capture_responses_state(output: list[Any], scope: str | None) -> dict | None:
     items: list[Any] = []
     order: list[dict[str, Any]] = []
     has_public_carrier = False
+    reasoning_items = [item for item in output if response_item_type(item) == "reasoning"]
+    summary_only = any(_field(item, "encrypted_content") is None for item in reasoning_items)
+    if summary_only and any(
+        _field(item, "encrypted_content") is not None for item in reasoning_items
+    ):
+        logger.warning(
+            "Responses reasoning contains items without encrypted content; replaying "
+            "portable summaries instead of an incomplete opaque reasoning sequence."
+        )
     for item in output:
         item_type = response_item_type(item)
         if item_type == "reasoning":
+            # A summary-only reasoning item has no opaque data to retain.
+            if summary_only:
+                continue
             order.append({"type": "reasoning", "index": len(items)})
             items.append(opaque_item(item))
         elif item_type == "function_call":
@@ -320,9 +359,16 @@ def capture_responses_state(output: list[Any], scope: str | None) -> dict | None
                 order.append(slot)
                 has_public_carrier = True
         elif item_type == "message":
-            order.append({"type": "message", "content": _responses_message_text(item)})
+            slot = {"type": "message", "content": responses_message_text(item)}
+            phase = _field(item, "phase")
+            if phase is not None:
+                slot["phase"] = phase
+            order.append(slot)
             has_public_carrier = True
-    if not items:
+    carriers = [slot for slot in order if slot["type"] != "reasoning"]
+    # Keep provider message boundaries/phase when the canonical flat text and
+    # calls alone cannot reproduce them, even without encrypted reasoning.
+    if not items and carriers == _flatten_responses_carriers(carriers):
         return None
     payload: dict[str, Any] = {"items": items, "order": order}
     if not has_public_carrier:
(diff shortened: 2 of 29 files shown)


src/nooa/unifiedllm/unifiedllm.py
```diff
@@ -1564,18 +1569,147 @@ def _extract_reasoning_and_usage(raw_response: Any) -> tuple[str | None, dict[st
...
+
+
+def _llm_state_scope(
+    model: str,
+    api_style: Literal["chat", "responses"],
+    params: dict[str, Any],
+    override: str | None,
+) -> str | None:
+    """Identify the exact model route that may receive captured state.
+
```
furgalep: Is it true that encrypted reasoning may only go back to the exact model?

furgalep: No — not in this PR. The gate here is MODEL-EXACT by default, which is stricter than the provider-level core we discussed in the evidence document. `_llm_state_scope` digests (litellm-resolved model, normalized endpoint), so encrypted reasoning goes back only to the exact same model on the exact same endpoint. A registry `replay_scope: <name>` declaration is the explicit escape hatch that deliberately groups routes (digest of the declared name + endpoint), and unknown routes fail closed at capture (state is dropped, not stored). Two caveats from my review worth knowing for this decision: (1) the survey evidence (Hermes #32716, Pi effort-provenance 400s) supports exactly this strictness for OpenAI specifically — its encrypted content is account/issuer-bound, so model+endpoint-exact is the right default for OpenAI; (2) the strictness is also why `replay_scope` exists — model-exact means model-id churn (gpt-5.6 -> gpt-5.6-sol) silently orphans stored state, and the declared group is how you opt into verified compatibility. The provider-level boundary we discussed would instead be the default and grouping would be implicit; this PR chooses exact-by-default with explicit grouping, which is a defensible and safer reading for OpenAI-only capture.

furgalep:
## Full review: retain issuer-scoped reasoning state

**Verdict: strong design, safe core — two silent-failure footguns should be fixed before merge, plus one ordering decision.** No blockers.

### What this PR gets right (verified by execution, not just reading)

- **The gate is genuinely airtight where it fires.** Chat-captured state cannot reach a Responses client (format + scope-prefix double gate, `unifiedllm.py:1648-1663`); cross-model and cross-endpoint replay is stripped; unknown routes fail closed at *capture* (`_make_llm_state` returns None for scope=None — state is dropped, never stored unscoped).
- **No bypass path.** Both replay-side transforms (`_prepare_completion_messages` at chat, `_transform_messages` for Responses) strip raw `reasoning_items`, native `type:"reasoning"` items, and gate `_batch` reasoning slots through the envelope only. No formatter emits `thinking_blocks` or `provider_specific_fields`, so provider state cannot sneak past on another field.
- **Leakage-checked.** `llm_state` is `repr=False` on events (encrypted content never reaches LLM-visible context or `pformat`), and the journal sideband skips it. SQLite round-trips preserve it (tested, both backends); legacy events with the old `reasoning_items` field load cleanly (Pydantic drops the unknown field).
- **Identity without substring matching.** `_llm_state_scope` resolves through `litellm.get_llm_provider` rather than string parsing — the right call, and the same conclusion the catalog-survey evidence reached independently.
- **Thread-safety clean:** `_replay_scope` set once in `__init__`; module constants immutable.

**Verification I ran:** 1,490 tests (unifiedllm + context_blocks + strategies) and 1,222 (runtime + round-trip), all green; ruff check + format clean; DCO green.

### Should-fix before merge — both are silent no-ops/leaks in plausible configs

**1. `replay_scope` silently disables the `include` request** (`unifiedllm.py:1777-1791`). `_include_encrypted_reasoning` only fires for scopes starting `responses:openai:` / `responses:azure:`; a `replay_scope` override produces `responses:declared:sha256:…`, so `include: ["reasoning.encrypted_content"]` is never added. **I reproduced this live:** `ResponsesClient(model="openai/gpt-5.6", replay_scope="my-group")` → captured `include` param is `None`. On a real OpenAI/Azure route that means no encrypted content is ever requested → nothing captured → the whole feature silently no-ops, while capture/replay machinery runs. Fix: also match `responses:declared:`, or document that `replay_scope` requires explicit `include`. Add the regression test (the existing declared-scope test only passes because it sets `include` by hand).

**2. Env-selected endpoints share one scope digest** (`unifiedllm.py:1602,1622`). When the base URL comes from `OPENAI_BASE_URL`, `litellm.get_llm_provider` returns `api_base=None`, so **two different gateways produce byte-identical scopes** — I reproduced this with two `OPENAI_BASE_URL` values. Harmless within one process, but events persist: resuming a stored session against a different gateway replays issuer-A encrypted state to issuer-B — exactly the leak this PR exists to prevent. Fix: fold the effective env base into `_normalized_endpoint` when `resolved_api_base` is None.

### Decision needed: multi-tool-call replay ordering

The direct `_batch` path preserves reasoning/call interleaving correctly, but the **event path** (CodeAct attaches state at `tool_call_index==0`, `codeact.py:1250`) captures ALL reasoning into one bag and replays it before the *first* function-call carrier: a turn that produced `rs_1, fc_1, rs_2, fc_2` replays as `rs_1, rs_2, fc_1, fc_2` (`unifiedllm.py:1746-1756, 2877-2882`). Whether OpenAI tolerates reordering is unverified; the wire order differs from the producing turn. Either interleave per-carrier or record the decision (provider accepts reordered encrypted items) with evidence.

### Nits (won't block)

- `_batch` replay-count mismatch is silently lossy (more items than slots → extras dropped, no log).
- `_normalized_endpoint` drops query strings — right call for secret-avoidance since the digest is stored in events, but worth a one-line comment on the trade-off.
- Four copies of the `**({LLM_STATE_KEY: copy.deepcopy(...)} if ... else {})` idiom — one helper would fold them.
- Double deep-copy on the Anthropic Responses path (measured 0.5 ms/200 messages — immaterial, noted for tidiness).

### Test gaps worth adding alongside the fixes

Declared-`replay_scope` include loss; env-var scope stability; `_batch` count mismatch; interleaved ordering; `ReasoningCompletionClient` passthrough; async chat-path capture (only sync chat + async responses are covered); one end-to-end session-resume test through actor → event store → formatter → client (current tests are per-layer).

**Recommendation:** fix issues 1 and 2, decide the ordering question, then this is mergeable. The envelope/scope design is sound and considerably simpler than the alternatives — `b5724d19` is a good shape for this layer.

src/nooa/unifiedllm/unifiedllm.py
```diff
@@ -2396,29 +2797,24 @@ def _make_call():
             _update_token_calibration(self.model, messages, usage, tools=api_params.get("tools"))
 
         output: list[Any] = raw_response.output  # type: ignore[assignment]
-        raw_tool_calls = [item for item in output if item.type == "function_call"]
+        llm_state = _responses_llm_state(output, state_scope)
+        raw_tool_calls = [item for item in output if _response_item_type(item) == "function_call"]
 
         if raw_tool_calls:
             tool_calls = [
                 ToolCall(id=tc.call_id or "", name=tc.name or "", arguments=tc.arguments or "")
```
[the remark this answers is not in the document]
furgalep: Addressed in e2995e19: Responses function calls now use one shared dict/object accessor in both sync and async paths, with regressions for dictionary-shaped output.

tests/test_event_backend_roundtrip.py
```diff
@@ -242,13 +242,18 @@ def test_event_roundtrip_via_all_events(backend, tag, event, expected_type, expe
...
+                "scope": "responses:openai:default",
+                "format": "openai-responses",
+                "payload": {"items": [{"encrypted_content": "opaque"}]},
+            },
         ),
         ToolCallEvent(
             tool_call_id="tc-state",
             name="execute_python",
             arguments={"code": "print(1)"},
             llm_output_id="assistant-turn-id",
```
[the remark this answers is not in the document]
furgalep: Addressed in e2995e19: the persistence fixture now stores a real LLMOutput and linked ToolCallEvent together and verifies the restored relationship.

furgalep:
Also addressed the outside-diff CodeRabbit test note in 7f1fee91: the multi-call projection now asserts exact call cardinality, order, and the source LLMOutput id for every execution.

furgalep:
@coderabbitai review

furgalep:
@coderabbitai review

furgalep:
@coderabbitai review

furgalep:
@coderabbitai review

src/nooa/unifiedllm/replay_state.py
```diff
@@ -0,0 +1,366 @@
...
+
+def _credential_fingerprint(provider: str, configured: Any, resolved: Any) -> str | None:
+    credential = configured or resolved
+    if not credential:
+        try:
+            credential = litellm.get_api_key(provider, resolved)
+        except Exception as exc:  # noqa: BLE001 - missing identity disables replay
+            logger.debug("Could not resolve opaque-state credential for %s: %s", provider, exc)
+            credential = None
+    if not credential and provider == "azure":
```
furgalep: This seems really random.

src/nooa/unifiedllm/replay_state.py
```diff
@@ -0,0 +1,366 @@
...
+        credential = (
+            os.getenv("AZURE_API_KEY")
+            or os.getenv("AZURE_OPENAI_API_KEY")
+            or os.getenv("AZURE_AD_TOKEN")
+        )
+    reveal = getattr(credential, "get_secret_value", None)
+    if callable(reveal):
+        credential = reveal()
+    if not isinstance(credential, str) or not credential:
+        return None
```
furgalep: Is this an error path? What happens when this returns None?

Does it really make sense to hash the credential?

furgalep: Why does credential have to be part of the fingerprint? Can we skip it?

src/nooa/unifiedllm/replay_state.py
```diff
@@ -0,0 +1,366 @@
...
+        params.get("organization")
+        or params.get("openai_organization")
+        or getattr(litellm, "organization", None)
+        or os.getenv("OPENAI_ORGANIZATION")
+    )
+    project = (
+        params.get("project")
+        or params.get("openai_project")
+        or getattr(litellm, "project", None)
+        or os.getenv("OPENAI_PROJECT")
```
furgalep: what are openai project and organization? Do these HAVE to be here? 

I expected we would only have to match the model, not all the other stuff.

furgalep:
@coderabbitai review

src/nooa/unifiedllm/replay_state.py
furgalep: This is LiteLLM's canonical credential-resolution fallback, but I agree the call is not self-explanatory. `get_llm_provider()` often returns the resolved key directly; if it does not, `get_api_key(provider, resolved_key)` checks LiteLLM's configured/global/environment key sources. If neither path yields a key, `replay_scope()` returns `None`: the call still runs, but opaque state is neither captured nor replayed and portable reasoning uses the #312 text fallback. No error is raised.

src/nooa/unifiedllm/replay_state.py
furgalep: The credential is included only as a non-reversible tenant proxy: the raw key is never stored, and its hash is itself folded into the final scope digest. The reason is OpenAI's observed cross-account binding for encrypted reasoning; model equality alone cannot distinguish two tenants using the same model and endpoint. The tradeoff is deliberate fail-closed behavior: key rotation or an unresolvable key can orphan opaque replay, but the turn still resumes through plain-text reasoning. If we decide that cross-account rejection is acceptable, we can remove this dimension, but that would be a policy relaxation rather than just cleanup.

src/nooa/unifiedllm/replay_state.py
furgalep: They are optional OpenAI request selectors, not requirements: when absent, `account` is `{}`. A legacy/user key plus `OpenAI-Organization` or `OpenAI-Project` can route requests to a different tenant even though the model, endpoint, and credential string are unchanged, so including an explicitly selected value prevents cross-tenant replay. The model is only the protocol-compatibility dimension; provider/API style and endpoint identify the wire/issuer, and credential plus optional org/project identify the account. I have left this thread open because choosing model-only would materially relax the gate and should be an explicit review decision.

src/nooa/unifiedllm/replay_state.py
furgalep: Removed in 0f0f2d6a. Replay no longer resolves or hashes credentials at all; provider/model identity comes from LiteLLM and endpoint identity comes from the effective base URL.

src/nooa/unifiedllm/replay_state.py
furgalep: Agreed and fixed in 0f0f2d6a. Credential presence, absence, and rotation now produce the same replay key. The regression also proves organization/project changes are ignored. Only provider, API style, exact resolved model, and endpoint participate.

src/nooa/unifiedllm/replay_state.py
furgalep: They do not have to be here, and 0f0f2d6a removes both. The key is now the small conservative boundary we discussed: provider + API style + exact resolved model + normalized endpoint. Exact model can later become a declared compatibility family in the capability layer; there is no manual `replay_scope` escape hatch in this PR.

src/nooa/unifiedllm/unifiedllm.py
furgalep: Confirmed against the current OpenAI docs: no, exact model is stricter than the provider contract, which permits reuse within a model family and omits incompatible reasoning. This PR intentionally keeps exact-model equality as the simple interim fail-closed rule; the capability layer can later replace it with a declared family. Commit 0f0f2d6a removes the unrelated credential/account and manual-group machinery.

furgalep:
@coderabbitai review

src/nooa/unifiedllm/replay_state.py
```diff
@@ -0,0 +1,317 @@
...
+    if not has_public_carrier:
+        payload["state_only"] = True
+    return _envelope(scope, _RESPONSES_FORMAT, payload)
+
+
+def prepare_chat_messages(messages: list[dict[str, Any]], scope: str | None) -> list[dict]:
+    """Strip private/raw state and restore only a matching Chat payload."""
+    prepared: list[dict[str, Any]] = []
+    for original in messages:
+        state = copy.deepcopy(carried_state(original))
```
furgalep: What's the deep copy rationale? Are we creating a lot of allocations?

furgalep: Good catch. The original rationale was mutation isolation: never let LiteLLM/provider request transforms mutate the persisted event. But the implementation copied at every intermediate hop, including several copies of an already-detached Responses batch. That was unnecessary and would multiply allocations as retained history grows.

Fixed in f588b7b7 with a narrower ownership rule that matches the append-only event direction: capture detaches provider-owned data once; RenderedMessage, the private carrier, and middleware borrow the immutable event state; the replay adapter makes one mutable deep copy at the final provider boundary. Public wire messages are still detached before we strip/add fields. A regression deliberately mutates the outgoing provider payload and verifies the stored LLMResponse state is unchanged. The focused formatter/middleware/replay set (132 tests), Ruff, and Pyright are green.

furgalep: Follow-up correction in c67a91b2: the final provider-edge deep copy was also unnecessary. The event owns immutable state and UnifiedLLM now passes that same object through to the LiteLLM call boundary; provider adapters have a read-only-input contract and own serialization. Tests pin object identity for both Responses items and Chat reasoning_items. I also found the pre-existing cache marker injector deep-copied the entire message history, which would have cloned attached state despite this rule. It is now copy-on-write: only messages that actually receive a cache marker are shallow-copied, with only the modified final content block copied when required. Unmarked history is borrowed. Focused replay/cache tests pass.

furgalep:
## Full review: retain issuer-scoped reasoning state (re-review of head `c67a91b2`)

**Verdict: approve-equivalent — the design held up under a fresh end-to-end read of all 12 commits since the 2026-09-08 review, and the post-review refactors made the code materially better. Three minor nits below, none blocking.**

### Re-review scope

The previous full review covered `7f1fee91`. Since then the branch gained 12 commits (identity simplification, gateway-route decoupling, ownership-edge copies, dispatch borrow, integration privacy fixes). I re-read the full diff at `c67a91b2` (24 files, +1284/−123) and re-ran the affected suites locally.

### What I verified this round (by execution, not just reading)

- **Fail-closed gate is still airtight.** The scope key is now exactly `api_style:provider:sha256(model)`; gateway/credential/endpoint changes provably don't disable replay (tests `test_scope_is_stable_across_explicit_endpoints`, `..._environment_selected_endpoints`, `..._auth_rotation_and_account_metadata` all pass). Hand-crafted envelopes and direct `reasoning_items` wire dicts cannot bypass the gate (`test_direct_reasoning_items_cannot_bypass_envelope_gate`, `test_non_openai_chat_provider_cannot_receive_reasoning_state`).
- **The borrowing refactor is sound.** `ReplayCarryingMessage.llm_state` now aliases immutable event state; the adapters copy only at ownership edges (`prepare_chat_messages`, `_clean_responses_batch`, `_transform_messages` all `deepcopy` before mutating). The relay no-op path preserves dict subclasses by identity (`intercepted_msgs != ctx.messages`), and `test_noop_relay_roundtrip_preserves_private_message_sidecars` pins it.
- **Privacy boundary verified in tests.** `encrypted_content` is recursively redacted in the scrubber (including the new JSON-string re-encode path for OpenInference attributes), the journal (`_safe_msg_to_dict`), and HTTP debug logs. `FakeLLMClient` now strips private state so no non-provider consumer can observe it.
- **Copy-on-write cache injection is correct.** The `copied`-set guard makes repeated copies of the same index idempotent, and `copy_last_content_block` re-lists content before the Anthropic block mutation, so the original message's blocks are never aliased. Batch-marker tests (`f40c3671`) cover the interaction with replay batches.
- **`include: reasoning.encrypted_content` is requested only where supported** (native OpenAI endpoint or Azure), and custom gateways never get it — matching the `store`/`include` registry plumbing added in `registry.py`.
- **Local runs:** 19/19 `test_reasoning_state_replay.py` and 210/210 across the affected unifiedllm/tracing/relay/context_blocks suites (`uv run pytest`). CI at head is fully green (lint, test, 3.13-compat, macOS sandbox, builds, secret-scan, DCO).

### Nits (non-blocking)

1. **Dead code:** `_is_state_only` (`replay_state.py:140`) has no callers anywhere in src or tests. `state_only` is checked directly via `payload.get("state_only")` in `prepare_responses_batch`. Delete it or use it in `prepare_chat_messages` for symmetry.
2. **Dead code:** `demote_chat_reasoning` (`_llm_state.py:99`) is now fully superseded by `replay_state.prepare_chat_messages`; its drop-empty-assistant logic was re-implemented inline there. No imports remain — remove it (keep `demote_responses_batch`, which is the live fallback path).
3. **Pathological edge in `prepare_responses_batch`:** if a single captured response contained two `function_call` items with the same `call_id` (last-wins in the `calls` map), the order loop would emit that carrier once per matching slot, duplicating a public call. OpenAI never does this in practice, so it's fine to leave — but a one-line comment or a `call_id in emitted_calls` guard at line ~271 would make it provably impossible.

### Ordering reconstruction

Re-checked the interleave algorithm: reasoning `pending` is flushed before each matching carrier, trailing reasoning only if a carrier was last emitted, and unmatched public items are appended in original order without duplication. Multi-call order is pinned by `test_responses_multi_call_state_preserves_provider_order`. This remains the trickiest code in the PR and it's now the clearest version of it.

**Conclusion:** this is ready to merge from my side. The three nits are cleanup-grade and can land in a follow-up or a trivial squash-fix commit.

alessiodevoto:
I would request changes on this PR. Preserving reasoning state is the right direction, but several guarantees claimed by the implementation do not hold. I reviewed the current head, `c67a91b2070dc6ef0aff2920590653864212a0e9`.

**All four code findings below were reproduced locally**, across six failing regression checks against that commit. These checks exercise the actual NOOA code, with provider calls mocked at the LiteLLM boundary: they establish what NOOA dispatches, not what a live provider accepts or rejects. The separate cross-account/gateway concern is an unverified assumption, not a reproduced provider failure.

The problem this solves is real: an assistant turn can contain public text, tool calls, and opaque provider-owned reasoning state. Retaining only the text and calls loses information that can help the model continue working. OpenAI recommends returning reasoning items alongside the associated function calls and results. [OpenAI reasoning documentation](https://developers.openai.com/api/docs/guides/reasoning#preserve-reasoning-across-calls)

This PR stores that state inside `LLMResponse`, carries it through rendering without exposing it in ordinary message JSON, and restores it immediately before provider dispatch. It matches provider, API style, and model; when they differ, it withholds the opaque state and uses available plain-text reasoning instead. That separation is sensible.

I found three concrete correctness issues, plus a validation gap:

1. **The compatibility check uses the wrong model when a call overrides it.**

   The scope is calculated from `self.model`, but the outgoing request subsequently applies `**kwargs`, which can replace `model`. I reproduced a `ResponsesClient` configured for `openai/gpt-5.6` retaining OpenAI reasoning while dispatching with `model="anthropic/claude-sonnet-4-5"`. The captured LiteLLM arguments contained the overridden model and the original reasoning item together. Responses captured during an overridden call also receive the wrong scope, by the same code path.

   The same construction appears in both client implementations and their synchronous/asynchronous paths. Build the effective request configuration first, then derive replay identity from the model actually being dispatched—or explicitly reject model overrides. [Responses code](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/c67a91b2070dc6ef0aff2920590653864212a0e9/src/nooa/unifiedllm/unifiedllm.py#L2275-L2294), [Completion code](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/c67a91b2070dc6ef0aff2920590653864212a0e9/src/nooa/unifiedllm/unifiedllm.py#L1769-L1786)

2. **The mutation guard does not detect ordinary message edits.**

   It checks that adjacent items share a batch ID and batch size. It does not check their contents or individual positions. Three local reproductions showed that:

   - Editing a rendered assistant carrier's `content` in place still replays the original reasoning alongside the replacement text.
   - Swapping the second and third carriers in a three-call batch silently restores the old call order.
   - Adding public text to a reasoning-only carrier discards that text entirely.

   This contradicts the promise that modified public messages survive while stale state is dropped. The carrier needs a way to verify the original public content and ordering, with safe fallback when either changes. Merely keeping the same Python objects cannot establish that. [Batch check](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/c67a91b2070dc6ef0aff2920590653864212a0e9/src/nooa/unifiedllm/unifiedllm.py#L2547-L2564), [reasoning-only handling](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/c67a91b2070dc6ef0aff2920590653864212a0e9/src/nooa/unifiedllm/replay_state.py#L237-L242)

3. **The cache optimization can discard portable reasoning.**

   Copying a marked message with `dict(message)` removes its private reasoning attributes. In the Anthropic Responses path, cache marking happens before reasoning is converted to text. I rendered `LLMResponse(content="public answer", reasoning="portable reasoning text")` and called an Anthropic `ResponsesClient` with `cache_control_injection_points=[{"role": "assistant", "position": "last"}]`. The outgoing input retained the public answer and cache marker, but lost the reasoning entirely.

   Preserve the metadata when copying a carrier, or apply cache markers after replay/fallback has been resolved. [Copy operation](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/c67a91b2070dc6ef0aff2920590653864212a0e9/src/nooa/unifiedllm/unifiedllm.py#L1251-L1265), [operation order](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/c67a91b2070dc6ef0aff2920590653864212a0e9/src/nooa/unifiedllm/unifiedllm.py#L2277-L2286)

4. **“Malformed state fails closed” is only partially implemented.**

   The code validates the envelope header and some container types, but accepts malformed items inside a matching payload. In the reproduction, I retained a valid captured envelope and ordering, replaced `payload.items` with `[{"type": "reasoning"}]`, and supplied portable reasoning text. The invalid item was forwarded and the fallback was suppressed.

   This needs structural validation of the recognized payload and ordering references before replay. The encrypted contents themselves should remain opaque. [Relevant code](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/c67a91b2070dc6ef0aff2920590653864212a0e9/src/nooa/unifiedllm/replay_state.py#L233-L264)

There are also assumptions that should be explicit:

| Assumption | Assessment |
|---|---|
| Exact-model matching is required | It is a conservative implementation choice. OpenAI documents reuse within a model family, so this intentionally sacrifices some valid reuse. |
| Changing gateways or accounts preserves compatibility | Not established by the tests. They mock provider calls and prove that NOOA forwards the state, not that the destination accepts it. The key trusts configured names and does not verify backend identity. |
| A rejected replay automatically falls back | It does not. Local mismatches get fallback; rejection by a provider is a separate failure path, explicitly outside this PR's scope. |
| Shared state is immutable | This is a convention, not enforced immutability. However, a test through the real pinned LiteLLM path, with only HTTP dispatch mocked, confirmed that its input copying protects the stored state during ID normalization. I am not reporting that as a production mutation bug. |

The cited OpenAI documentation supports model-family compatibility; it does not establish the broader cross-account/gateway guarantee. I would not mechanically reintroduce credential hashes, either. The implementation should distinguish a verified compatibility agreement from a matching configuration string.

I would keep the overall architecture: one durable assistant-turn record, opaque state owned by the provider adapter, public formatting, and conservative fallback. Fix the boundaries above and narrow the documented guarantees to what is actually verified. This does not require a wholesale redesign.

Validation performed locally:

- **883 existing focused tests passed**: 846 across UnifiedLLM, context blocks, tracing, middleware integration, and CodeAct text-only replies; another 37 for Relay.
- **Six additional checks failed**, covering the four findings above: in-place text edits, text added to a reasoning-only carrier, model override, reordered calls, assistant cache marking, and malformed matching payloads.
- Two additional checks passed: canonical `LLMResponse` JSON persistence/resume retained compatible reasoning, and the real LiteLLM normalization path preserved stored state.
- No live provider calls were made. Repository source files were not changed.

furgalep:
Thank you — I reproduced and fixed all four findings in 3babc65f.

1. **Effective request model:** all four dispatch paths (Completion/Responses, sync/async) now derive replay scope, cache behavior, schema compatibility, diagnostics, and token calibration from the model that the request actually sends. An OpenAI-configured client overridden to Anthropic no longer forwards or captures the OpenAI envelope.
2. **Public-carrier integrity:** the durable Responses envelope now records the minimal public projection for every ordering anchor (message text or call id/name/arguments). Replay compares the current ordered projection before restoring state. Text edits, call reordering, and edits to a reasoning-only placeholder now preserve the edited public data and drop the stale blob. This also works after persistence/resume; it does not rely only on Python object identity.
3. **Cache copy:** copy-on-write cache marking now uses a shallow subtype-preserving copy, then detaches only the nested content it mutates. Portable reasoning sidecars survive Anthropic assistant cache marking.
4. **Malformed state:** matching v1 Responses payloads now validate the exact NOOA-owned schema, complete/unique reasoning indexes and call ids, state-only consistency, reasoning item type, and non-empty encrypted content before forwarding anything. #310 warns and safely demotes malformed state; #311 deliberately upgrades malformed current-version state to a terminal error under the stack's warning/error policy.

I also narrowed the PR description: gateway/account stability is NOOA's declared compatibility policy, not evidence that every destination accepts an issuer-owned blob; provider-rejection recovery remains a non-goal.

Validation at this commit:
- full suite: **7,174 passed**, 6 skipped, 298 deselected, 3 expected xfails
- affected UnifiedLLM/formatter/middleware set: **750 passed**, 5 deselected
- Ruff and targeted Pyright clean

furgalep:
@coderabbitai review

furgalep:
Follow-up from the cross-stack boundary audit: commits 5bfb64b1 and 45695d05 close two additional configured-vs-effective paths.

- `input=` cannot replace a Responses payload after replay gating, and constructor-level `messages=` cannot replace Chat history.
- OpenAI SDK `extra_body` is also checked: nested `input`/`messages` or `model` now fail with a direct configuration error, because the SDK applies those values after top-level parameters.
- The prepared payload is assigned after config merging as defense in depth.
- Responses no longer re-applies constructor `reasoning` after kwargs merging, so a per-call override is the actual value sent in both sync and async paths.

The expanded affected set is green: **759 passed, 5 deselected**; Ruff and targeted Pyright remain clean.

furgalep:
@coderabbitai full review

src/nooa/tracing/_secret_scrubber.py
```diff
@@ -47,6 +48,11 @@
     }
 )
 
+# Provider-owned replay state is not a user credential, but it has the same
+# telemetry rule: it may go back to its issuer and nowhere else. Provider
+# adapters add their exact wire keys here as support is introduced.
+_OPAQUE_PROVIDER_STATE_KEYS = frozenset({"encrypted_content"})
```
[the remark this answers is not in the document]
furgalep: Fixed in baffb0fb. The shared scrubber now redacts the normalized private envelope key, including unknown future payload fields. Mapping and JSON-string journal regressions pass.

src/nooa/unifiedllm/unifiedllm.py
```diff
@@ -1784,21 +1796,27 @@ def call(
         If retry_config.retry_on_empty_content is True, will retry when the model
         returns empty content but has reasoning_content (common with some reasoning models).
         """
-        messages = demote_chat_reasoning(messages)
+        call_config = {**self.config, **kwargs}
+        self._validate_request_config("messages", call_config)
+        effective_model = self._effective_model(call_config)
```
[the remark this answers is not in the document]
furgalep: Confirmed against installed LiteLLM and fixed in baffb0fb. Per-call routing or credentials now omit the constructor-bound SDK client so LiteLLM resolves the current values. Unchanged calls retain the owned client. Six sync/async tests exercise real LiteLLM/SDK dispatch with HTTP mocked, including default to override to default URL and Authorization checks.

alessiodevoto:
Follow-up review of `aca3c9ae937b56b50b5643967a988f0f59677636`, focusing on unnecessary complexity and duplication.

**The original fixes work: all eight checks from my previous review now pass, along with 939 focused tests.** Thank you for addressing those cases.

The design is reasonable, but there is avoidable duplication. I would ask for a small cleanup pass. Most added lines are tests; preserving message order, phases, and mutation checks is necessary complexity.

The main opportunities are:

1. **Replay metadata duplicates potentially large public content.**

   Chat stores assistant text and full tool arguments again in `payload.carrier`; Responses repeats call arguments in its ordering records. These values already exist on `LLMResponse`.

   I measured a synthetic turn with a 100,000-character tool argument: its serialized size grew from **100,377 to 200,750 bytes** with replay state.

   For data needed only to detect edits, a fingerprint of the normalized public turn would be sufficient. Retain actual content where reconstructing separate Responses messages requires it. This would also remove some carrier-schema validation. This is a design improvement to consider, rather than a prerequisite for fixing the remaining correctness gap below. [Chat capture](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/aca3c9ae937b56b50b5643967a988f0f59677636/src/nooa/unifiedllm/replay_state.py#L211-L228), [Responses call slots](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/aca3c9ae937b56b50b5643967a988f0f59677636/src/nooa/unifiedllm/replay_state.py#L251-L264)

2. **The borrowing optimization still makes an unnecessary deep copy.**

   `_clean_responses_batch()` calls `copy.deepcopy(original)`. Because `original` is the special dict subclass, that copies its private state attributes too. I instrumented the operation and confirmed traversal of the stored envelope, payload, and reasoning item. This rebuilds mutable containers; it does not imply that Python duplicates the bytes of immutable string values.

   Copying `dict(original)` instead would detach only the public message. Some fallback paths then copy those public messages again, so ownership could be clearer: prepare one mutable public copy, then operate on it. [Copy operation](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/aca3c9ae937b56b50b5643967a988f0f59677636/src/nooa/unifiedllm/replay_state.py#L444-L455), [fallback copy](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/aca3c9ae937b56b50b5643967a988f0f59677636/src/nooa/_llm_state.py#L99-L107)

3. **The separate validators have already diverged.**

   Responses validates individual reasoning items; Chat mostly validates the carrier and checks that `reasoning_items` is a nonempty list.

   I reproduced a remaining gap locally: capture a valid Chat envelope, replace `state["payload"]["reasoning_items"]` with `[123]`, and pass it through `prepare_chat_messages()` with the unchanged public carrier and matching scope. The result is:

   ```python
   [{"role": "assistant", "content": "done", "reasoning_items": [123]}]
   ```

   The malformed item passes preparation and portable fallback is suppressed. This reproduction uses the actual local preparation function; no live provider call is involved.

   A small shared reasoning-item validator, with format-specific carrier checks, would reduce duplication and fix this inconsistency. **This is the remaining correctness issue I would address before approval.** [Chat validator](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/aca3c9ae937b56b50b5643967a988f0f59677636/src/nooa/unifiedllm/replay_state.py#L185-L208), [Responses item validation](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/aca3c9ae937b56b50b5643967a988f0f59677636/src/nooa/unifiedllm/replay_state.py#L267-L282)

4. **HTTP logging performs redundant redaction.**

   `_redact_body()` runs the shared `scrub_value()`, then recursively walks the result again with `_redact_opaque_state()`. The shared scrubber now handles both opaque keys already.

   The second traversal and its separate key list can be removed. This is straightforward cleanup with less maintenance risk. [HTTP redaction](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/aca3c9ae937b56b50b5643967a988f0f59677636/src/nooa/unifiedllm/http_logging.py#L117-L125), [shared opaque-key list](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/aca3c9ae937b56b50b5643967a988f0f59677636/src/nooa/tracing/_secret_scrubber.py#L51-L54)

I tested an **in-memory cleanup prototype** that removed the extra HTTP traversal and copied only public message mappings: **140 relevant tests passed**, with the direct test of the removed helper excluded. This validates those two narrow changes against the selected tests; the fingerprint proposal and removal of subsequent fallback copies were not implemented or tested. Repository source files remain unchanged.

I would keep the versioned envelope, ordering reconstruction, and regression coverage. Fix the Chat validator and remove the redundant copying/redaction now; treat compacting the stored carrier as a separate, worthwhile design improvement. A broader abstraction or framework rewrite would add more complexity than this needs.

furgalep:
Review of `e73f1c0b`. Two fixes requested before merge; everything else in this PR reads well and the targeted suites pass on the current head (1935 passed, ruff clean). Note that #312 has already landed on `main`, so please retarget this PR's base to `main`.

### 1. `ResponsesClient` ignores per-call transport overrides that `CompletionClient` now honours

`CompletionClient._completion_http_client` (`src/nooa/unifiedllm/unifiedllm.py:1784`) correctly drops the constructor-bound httpx client when a call passes a different `model`, `api_base`, `base_url`, `api_key`, or `custom_llm_provider`, because LiteLLM uses a supplied SDK client's bound URL and key and silently ignores the matching call parameters.

`ResponsesClient.call` and `ResponsesClient.acall` did not get the same treatment. Both still do

```python
http_client = self._http
assert http_client is not None
if http_client.sync_client is not None:   # / async_client
    api_params.setdefault("client", ...)
```

at `unifiedllm.py:2365` and `unifiedllm.py:2498`, so a per-call `api_base` or `api_key` override on the Responses path is accepted (this PR even computes `effective_model` and the replay scope from it) but the request still goes to the constructor's endpoint with the constructor's key.

**Fix:** move `_completion_http_client` up to `UnifiedLLM` (rename to `_request_http_client` or similar) and use it from all four call sites. Add the Responses equivalents of the cases in `tests/unifiedllm/test_completion_transport_overrides.py`: override `api_base`, override `api_key`, override `model`, and no override, asserting whether `client` is present in the `litellm.responses` / `litellm.aresponses` kwargs.

### 2. `scrub_value` re-serialises JSON strings with `ensure_ascii=True` and parses every string

`src/nooa/tracing/_secret_scrubber.py:240-245` now tries `json.loads` on every string attribute and, when the recursive walk redacts something, writes it back with `json.dumps(decoded, separators=(",", ":"))`.

Two problems:

- `json.dumps` defaults to `ensure_ascii=True`, so any non-ASCII text in a redacted span attribute (user prompts, tool output, file contents) comes back as `\uXXXX` escapes. That changes the recorded payload for reasons unrelated to redaction.
- Every string value, including large tool outputs and code blocks, is now run through `json.loads`. Most fail on the first byte, but strings starting with `{` or `[` that are not valid JSON get scanned to the failure point before falling back to the regex path.

**Fix:**

- Pass `ensure_ascii=False` to that `json.dumps`.
- Only attempt the parse when the stripped string starts with `{` or `[`; everything else should go straight to `scrub_string` as before.
- Add a test in `tests/tracing/test_secret_scrubber.py` that redacts a key inside a JSON string containing non-ASCII text (e.g. `{"api_key": "sk-…", "text": "héllo 世界"}`) and asserts the non-ASCII characters survive verbatim, plus one asserting a non-JSON string is returned unchanged and never parsed.

### Not blocking

The borrowed-state design (no deep copy of `llm_state`, archive `reasoning_items` placed directly on the wire message) is fine with LiteLLM 1.97: the Chat→Responses bridge builds fresh dicts from each reasoning item, and the Responses input path only walks the items to pop `cache_control`, which archive items never carry. `tests/integration/test_cache_resume_live.py` already asserts the archive is unchanged after a request, which is the right guard if LiteLLM changes that.

furgalep:
Withdrawing item 1 (Responses transport overrides). Verified against LiteLLM 1.97: `response_api_handler` uses a supplied `client` only as a bare `HTTPHandler`/`AsyncHTTPHandler` and computes `url` and `headers` per request from the call's `api_base` and `api_key` (`llms/custom_httpx/llm_http_handler.py`, the `sync_httpx_client.post(url=api_base, headers=headers, …)` call). Unlike the Chat path's bound OpenAI SDK client, per-call overrides are honoured while keeping the configured pool. No change needed there.

Item 2 (scrubber `ensure_ascii` and parse guard) still stands.

furgalep:
Follow-up to [the review of e73f1c0b](https://github.com/NVIDIA-NeMo/labs-OO-Agents/pull/310#issuecomment-5631744395), now addressed in bdf1f4f1.

**1. Responses transport overrides: verified working; regression coverage added.**

The distinction is the wrapper type. CompletionClient's OpenAI path supplies an OpenAI SDK client bound to its original URL/key. ResponsesClient instead uses [generic HTTPHandler/AsyncHTTPHandler wrappers](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/bdf1f4f1/src/nooa/unifiedllm/unifiedllm.py#L275), which receive the URL and authorization per request. Reusing these handlers retains the configured connection pool without overriding request routing.

Added [12 sync/async regression cases](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/bdf1f4f1/tests/unifiedllm/test_responses_transport_overrides.py) for no override, api_base, base_url, api_key, model, and combined endpoint/key overrides. Each runs default → override → default through real NOOA/LiteLLM serialization and dispatch, with only HTTP mocked. They assert the actual request URL, Authorization header, and model, and prohibit an unintended fallback to a network transport. All 12 passed before the scrubber fix; Responses dispatch is unchanged. The shared Completion helper is therefore not needed on this path.

**2. Scrubber: fixed.**

JSON parsing is attempted only when the first non-whitespace character is an opening object/array delimiter; the prefix check does not allocate a stripped copy of large text. Redacted JSON uses ensure_ascii=False. Ordinary text still gets regex secret scrubbing, invalid JSON falls back to it, and clean JSON retains its original formatting.

Nine new scrubber cases cover literal Unicode in object/array attributes, ordinary strings never reaching json.loads, continued secret redaction, invalid JSON, and untouched clean formatting. Seven reproduced failures before the fix; all now pass.

Validation: **717 passed** on #310's UnifiedLLM/tracing set; **1,056 passed** on the propagated stack's UnifiedLLM/context/tracing/Relay set. Ruff and targeted Pyright pass. No live inference calls.

The PR already targets main. Its walkthrough now explicitly explains the different Chat and Responses transport behavior.

furgalep:
Superseded by #318, which retains OpenAI reasoning in ordered response parts passed directly back to UnifiedLLM. This replaces the envelope/carrier reconstruction design. Review #318 first, then #319 for cache policy. Closing this PR to focus human review on that replacement; its branch and review history are retained.

PR #312, opened 2026-09-08
refactor(llm): make LLMResponse the canonical assistant turn

furgalep:
## Summary

- make the LLMResponse returned by UnifiedLLM the single durable assistant-turn event
- store public text, ordered tool calls, plain reasoning, opaque replay state, normalized usage, and correlation metadata once
- keep raw SDK responses and arbitrary typed return values live-only
- reconstruct provider-independent messages from the canonical turn, including reasoning-only persisted turns
- keep execute_python receipts immutable across recursive generation so provider prompt prefixes stay cacheable
- read-migrate old SQLite LLMOutput rows and fail loudly when code subscribes to or queries removed event names

## Why this layer exists

A provider response has two jobs in NOOA: it is the value strategy code consumes now, and it is the assistant turn that must be persisted and reconstructed later. Separate LLMResponse, LLMOutput, and LLMComplete objects copied the same facts across three schemas. Those copies could drift or drop text, tool ordering, reasoning state, or usage.

This PR makes one object cross that boundary. UnifiedLLM creates LLMResponse; the runtime enriches and stores that exact object. The durable object remains provider-independent, so the event list is a public IR rather than a LiteLLM/OpenAI/Anthropic data structure.

That IR still has to become provider messages on the next generation. Context formatting reconstructs the public assistant turn and carries persisted reasoning/state privately to UnifiedLLM. At the final LLM boundary, plain reasoning is safely demoted to ordinary assistant text by default, while opaque state is withheld. This means a reasoning-only structured response survives archive/resume even when its live-only parsed Python value is gone. Provider-specific follow-ups (#310 and #311) can replace the fallback with exact native replay only after their compatibility gate succeeds. Keeping that decision at the final boundary avoids both data loss and the false claim that reasoning was originally provider `content`.

This reconstruction is also what keeps caching sound: the historical prefix is derived from one immutable turn instead of three drifting event copies, and provider-owned replay material lives in one place rather than being duplicated onto tool events.

## Code walkthrough: what changed and why

1. **Canonical contract — `src/nooa/llm_types.py`.**  
   **What:** defines provider-independent ToolCall, normalized LLMUsage, and the one durable LLMResponse. Every public field has a schema description; finish_reason is normalized to stop, tool_calls, length, or error.  
   **Why:** callers, storage, renderers, and telemetry need one stable vocabulary rather than provider-specific finish strings and three partially overlapping response types.

2. **One object from provider to event log — UnifiedLLM and RuntimeActor.**  
   **What:** every provider adapter builds LLMResponse; the runtime adds generation/model/dynamic-context metadata and records that same object instead of creating LLMOutput and LLMComplete copies.  
   **Why:** object identity removes synchronization bugs. Text, tool calls, reasoning, usage, and replay state cannot be lost while copying the client result into session history.

3. **Provider-independent conversation reconstruction — formatter and strategy paths.**  
   **What:** ToolCallEvent links execution/result data to `llm_response_id`; formatters join those events to the ordered calls declared on the canonical response. Reasoning and opaque state travel only as private in-memory attributes, outside the public provider-message mapping. CodeAct renders one stable `status: accepted` receipt while lifecycle status changes internally and the actual outcome is appended as PythonOutput.  
   **Why:** a durable assistant turn must be reconstructable into the provider message sequence needed for replay and caching, while tool execution remains a separate public runtime event. Rewriting an earlier receipt after recursive generation would invalidate the cached nested trajectory.

4. **Safe default replay policy at the LLM edge.**  
   **What:** Completion and Responses clients turn portable reasoning into assistant text when no provider-specific replay layer recognizes the private state; opaque-only turns are withheld instead of becoming empty assistant messages.  
   **Why:** text reasoning is useful across model switches, but an opaque provider blob is not generic conversation data. #312 must be independently safe before the exact OpenAI/Anthropic/Gemini gates land above it.

5. **Explicit durable/live boundary.**  
   **What:** normalized content, calls, reasoning, opaque llm_state, usage, and correlation fields serialize. `raw_response` and `parsed` do not.  
   **Why:** provider SDK objects are version-specific and frequently non-serializable; arbitrary parsed Python objects make archives depend on application classes. Structured-output source JSON remains durable in content or provider-exposed reasoning.

6. **Safe integrations — SQLite, ATIF, ACP, Relay, memory, tracing, and skills.**  
   **What:** consumers read the canonical event; usage/cost lives on the same turn; Relay projects only public fields and emits no synthetic message for a state-only response; opaque state is excluded from search and generic exports.  
   **Why:** making the event public is safe only if every projection has an explicit public/private boundary.

7. **Intentional alpha migration.**  
   **What:** SQLite maps persisted `event_type=LLMOutput` rows to LLMResponse on read, including archived rows. LLMComplete needs no stored migration because it was never persisted. Subscribing to or querying LLMOutput/LLMComplete raises a specific error naming LLMResponse and the archive behavior.  
   **Why:** existing sessions keep their assistant history, while source code cannot silently subscribe to an event that will never fire. We preserve data, not dead alpha APIs.

8. **Regression coverage.**  
   **What:** tests pin object identity, persistence, active/archived legacy reads, helpful legacy subscription/query failures, failed/state-only turns, text-plus-tools, malformed batches, recursive-generation prefix stability, Relay projection, usage, and a serialize/reload replay of structured JSON emitted only as reasoning.  
   **Why:** these are the boundaries where the previous three-object design lost data or where a canonical public IR could accidentally leak private state.

## Validation

- final combined-tip full suite (through #313): **7,203 passed**, 6 skipped, 298 deselected, 3 expected xfails
- #312 full suite after the portable-reasoning move: **7,129 passed**, 6 skipped, 298 deselected, 3 expected xfails
- Ruff format, Ruff lint, targeted Pyright, and git diff checks pass
- live provider matrix from the stack verifies JSON resume, retained-state replay, incompatible-state stripping, text-reasoning demotion, trailing dynamic context, and cache reuse; the final ownership move was additionally covered by deterministic persistence/replay tests

#310 targets this branch and adds gated OpenAI opaque-state replay.

## Summary by CodeRabbit

* **New Features**
  * Added unified LLM response handling with structured content, parsed results, usage metrics, reasoning, and tool calls.
  * Added support for multiple ordered tool calls, linked execution replay, and preserved assistant text alongside tool calls.
  * Added a running execution status for tool results.
  * Structured outputs now expose parsed values separately from raw response content.
* **Bug Fixes**
  * Improved replay reliability by omitting incomplete tool-call batches and preserving reasoning across supported providers.
  * Legacy stored responses remain readable through automatic conversion.
* **Documentation**
  * Updated guidance and examples to reflect the unified response format.


furgalep:
@coderabbitai review\n\nBottom-up review fixes are now pushed: production adapters preserve mixed assistant text/tool turns, incomplete or malformed batches fail closed at replay, linked state round-trips through persistence, and obsolete code-comment mutation is removed.

furgalep:
@coderabbitai review

furgalep:
@coderabbitai review

sklinglernv:
This is a big change, did you run capability A/B tests? are we introducing any issues with some model providers?

sklinglernv:
Also, would it be possible to test for a GPT and an Anthropic model if caching still works on a real endpoint?

furgalep:
@coderabbitai review

furgalep:
Provider and caching validation is now complete on the current stack tip after rebasing onto #315:

- Full repository suite: **7,132 passed**, 7 skipped, 3 expected xfails.
- Focused provider, IR, and context slice: **849 passed**.
- NVIDIA-routed gpt-5.6-sol using Responses: encrypted reasoning was captured, persisted through JSON resume, replayed exactly on the next request, and accepted. With a trailing dynamic context block, the second request reported **3,766 / 3,821 cached input tokens** and the recorded wire body contained explicit mode plus exactly one breakpoint.
- NVIDIA-routed Claude Haiku: signed thinking plus text reasoning persisted exactly, replayed through a tool-result turn, and the next call read **5,819 / 5,925 cached tokens** with 64 newly written.
- NVIDIA-routed Gemini Flash Lite: opaque state plus text reasoning persisted and replayed; the repeated stable prefix reported **5,429 cached tokens** with a changed trailing dynamic block.
- Cross-provider regressions explicitly prove that incompatible opaque state is stripped, while plain reasoning is replayed as ordinary assistant text.

This is not a benchmark-quality capability-score A/B, and I do not want to overclaim one. #312 does not change prompting or reasoning effort; it removes the parallel response and event copies while pinning provider projection with formatter and serialized-wire tests. The live matrix checks the failure modes this IR change could introduce: provider request validity, resume, exact state replay, dynamic suffix placement, and cache reuse.

The available credentials are for NVIDIA inference-hub routes, not direct native OpenAI, Anthropic, or Google accounts. Native wire shapes are additionally covered by deterministic serialized-request tests; the live numbers above are specifically the enterprise gateway paths we deploy.

src/nooa/llm_types.py
```diff
@@ -0,0 +1,187 @@
...
+    consumers read its model and usage metadata.
+    """
+
+    _role: ClassVar[Role] = Role.ASSISTANT
+
+    raw_response: Any = Field(
+        default=None,
+        exclude=True,
+        repr=False,
+        description="Non-serialized provider response retained on the live in-memory object",
```
furgalep: Why isn't this serialized?

src/nooa/llm_types.py
```diff
@@ -0,0 +1,187 @@
...
+    content: Annotated[
+        str,
+        spec(max_string=None),
+        Field(description="Exact normalized assistant text used for replay"),
+    ] = ""
+    parsed: Any = Field(
+        default=None,
+        exclude=True,
+        repr=False,
+        description="Non-serialized typed result retained on the live in-memory object",
```
furgalep: Why not serialized?

src/nooa/llm_types.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+    tool_calls: list[ToolCall] = Field(
+        default_factory=list,
+        repr=False,
+        description="Ordered public tool calls emitted on this assistant turn",
+    )
+    finish_reason: Literal["stop", "tool_calls", "length", "error"] = Field(
```
furgalep, line 135: Is this the full list of stop reasons that LLMs can return?

src/nooa/llm_types.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+    generation_id: str = Field(
+        default="",
+        repr=False,
+        description="Generation turn that produced this response",
+    )
+    dynamic_context: str = Field(
+        default="",
+        repr=False,
+        description="Trailing dynamic context rendered for this request",
```
furgalep: I don't get what goes in this variable.

src/nooa/llm_types.py
```diff
@@ -0,0 +1,187 @@
...
+
+from nooa.agentdoc import spec
+from nooa.context_blocks.events import EventBase
+from nooa.context_blocks.roles import Role
+
+
+class ToolCall(BaseModel):
+    """Provider-independent tool call exactly as emitted by the model."""
+
+    id: str
```
furgalep: Should we annotate these with descriptions?

src/nooa/llm_types.py
```diff
@@ -0,0 +1,187 @@
...
+    name: str
+    arguments: str
+
+
+class LLMUsage(BaseModel):
+    """Normalized usage reported for one successful LLM response."""
+
+    input_tokens: int = 0
```
furgalep: Should we annotate these with descriptions?

src/nooa/llm_types.py
furgalep: Yes. I added Pydantic field descriptions for the provider-assigned call id, tool name, and exact JSON argument string. They now appear in the generated schema rather than requiring readers to infer their roles.

src/nooa/llm_types.py
furgalep: Yes. Every normalized usage counter now has a schema description, including the distinction between cache reads, cache writes, reasoning output, totals, and estimated USD cost.

src/nooa/llm_types.py
furgalep: dynamic_context is the exact trailing live-state block that was added to the request which produced this response (for example, current progress/state). It is persisted for session export and debugging; I expanded the field description to say that explicitly.

src/nooa/llm_types.py
furgalep: No—this is intentionally NOOA's complete normalized list, not every provider's vocabulary. Provider values such as content_filter, Responses incomplete status, and provider failures are collapsed to error/length; downstream strategies therefore need only handle four portable outcomes. I clarified that in the schema description.

src/nooa/llm_types.py
furgalep: parsed can be any Python object, so persisting it would make the event archive depend on arbitrary application classes and serializers. For structured output, the source JSON remains durably in content; parsed is the convenient typed live view returned to strategy code. I expanded the description with that reason.

src/nooa/llm_types.py
furgalep: raw_response is a provider SDK object: it may not be serializable, is provider/version-specific, and duplicates the normalized durable fields. The exact replay material we actually need is extracted into the serializable llm_state; the raw object remains available during the live call only. I made that boundary explicit in the field description.

furgalep:
@coderabbitai review

furgalep:
@coderabbitai review

src/nooa/atif/exporter.py
```diff
@@ -139,8 +138,7 @@ class _DispatchStep(NamedTuple):
     "AfterAgentCall": "on_after_agent_call",
     "BeforeTurn": "on_before_turn",
     "SystemPrompt": "on_system_prompt",
-    "LLMComplete": "on_llm_complete",
-    "LLMOutput": "on_llm_output",
```
furgalep, line 143: When would LLMOutput fire? When would LLMComplete fire?

furgalep:
@coderabbitai review

src/nooa/unifiedllm/unifiedllm.py
```diff
@@ -1951,22 +1902,23 @@ def _make_call():
 
             return LLMResponse(
                 raw_response=raw_response,
-                content=parsed_content,
+                content=text_content,
```
sklinglernv, line 1915: [Codex Take] This fallback parses structured output from `reasoning` when `content` is empty, but constructs the durable event with `content=text_content` (still empty) while `parsed` is excluded from serialization. After persistence and reload, both the typed value and its source JSON are gone, so neutral replay omits this successful assistant turn. This is reachable for models such as Nemotron that return structured JSON in reasoning. Please retain `parseable_content` (or the exact extracted JSON) in `content` when reasoning supplied the payload, and make the same change in the async branch.

src/nooa/context_blocks/formatter.py
```diff
@@ -274,6 +297,88 @@ def _event_block_to_messages(
...
+                logger.warning(
+                    "LLMResponse %s tool call %s has no execution event — emitting "
+                    "placeholder tool_result to preserve protocol ordering.",
+                    event.id,
+                    call.id,
+                )
+                messages.append(
+                    RenderedMessage(
+                        role=Role.TOOL,
+                        content="(tool call was not executed)",
```
sklinglernv: [Codex Take] The immediate protocol receipt plus later system-message result is the right design for nested NOOA calls. The gap is the timing for a multi-call batch: `_process_tool_calls` creates each `ToolCallEvent` only immediately before executing that call. If the model emits calls A and B and A triggers a nested generation, A already has the stable `status: accepted` receipt, but B has no execution event yet, so this formatter emits `(tool call was not executed)`. After A returns and B starts, the same history changes to `status: accepted`. Please create stable accepted/pending receipts for the entire batch before executing A, then append the actual outputs as system events as intended. Filtered projections should likewise avoid turning a missing linked event into an assertion of nonexecution.

src/nooa/atif/exporter.py
furgalep: Before #312, both fired after each successful provider call, in this order: `LLMComplete` first (`record=False`) carried model, usage/cost, reasoning, dynamic-context, and tool-call telemetry; `LLMOutput` second (`record=True`) carried only the assistant content used by conversation history. That split existed because telemetry was originally non-conversational, but it forced the runtime and consumers to correlate two copies of one provider turn. Now the exact `LLMResponse` returned by UnifiedLLM is recorded once. `on("LLMResponse", ...)` fires when that response is added, after the provider returns and before the strategy executes its requested tools. `EventManager._validate_event_type_name` rejects subscriptions to either removed name with a migration message: use `LLMResponse`; stored `LLMOutput` rows auto-migrate, while `LLMComplete` was never persisted.

src/nooa/unifiedllm/unifiedllm.py
furgalep: I traced this through serialization and do not think the proposed promotion is safe. The source JSON is not lost: `reasoning` is a serialized `LLMResponse` field; only the arbitrary live `parsed` object and provider SDK `raw_response` are excluded. Promoting reasoning into `content` would make the durable event claim the provider emitted assistant text when it emitted none, and later replay could send the same material as both reasoning and ordinary assistant content. Commit 0ec74617 adds a round-trip assertion proving the reasoning JSON survives serialization with `content` still exact. Reasoning projection/replay belongs to the provider-aware follow-up layer, which can preserve or demote it deliberately without corrupting this canonical turn.

src/nooa/context_blocks/formatter.py
furgalep: Confirmed and fixed in 0ec74617. The formatter no longer synthesizes `(tool call was not executed)` or reconstructs a linked execution as a standalone assistant turn. It emits a canonical tool-call turn only when every call has a visible linked execution with a recorded result; otherwise the entire assistant tool-call batch is omitted. A linked execution whose source response was filtered out is omitted as well. Only genuinely unlinked legacy/synthetic tool events retain standalone rendering. Focused coverage includes both missing-result and filtered-source cases.

furgalep:
Capability A/B is complete using `nooa eval` through NVIDIA Inference Hub.

### Setup

- Baseline: `main` at `ee61c901`
- PR measurement: `#312` at `2aaa9fd5`
- Suite: all 40 entries in `tests/capability/config.yaml`, one data row per capability, `runs=1` = 160 evaluations per branch
- Identical comparable-tier routes on both branches:
  - OpenAI `openai/azure/openai/gpt-5.6-sol` via Responses, medium reasoning
  - Anthropic `openai/azure/anthropic/claude-sonnet-4-6`
  - Google `openai/gcp/google/gemini-3.1-pro-preview`
  - NVIDIA `openai/nvidia/nvidia/nemotron-3-super-v3`

### Initial A/B

| Model | main | #312 |
|---|---:|---:|
| GPT-5.6 Sol | 40/40 | 40/40 |
| Claude Sonnet 4.6 | 39/40 | 39/40 |
| Gemini 3.1 Pro | 40/40 | 39/40 |
| Nemotron 3 Super | 39/40 | 40/40 |
| **Total** | **158/160 (98.8%)** | **158/160 (98.8%)** |

The two differing cells swapped: Nemotron `context_notes` timed out on main and passed on #312; Gemini `employee_lookup` passed on main and selected the wrong salary on #312. There were no provider/transport failures.

Per the agreed rule, I reran both differing capabilities three times on both branches for Gemini and Nemotron:

| Capability/model | main rerun | #312 rerun |
|---|---:|---:|
| `context_notes` / Gemini | 3/3 | 3/3 |
| `context_notes` / Nemotron | 3/3 | 3/3 |
| `employee_lookup` / Gemini | 0/3 | 2/3 |
| `employee_lookup` / Nemotron | 2/3 | 3/3 |

Including the initial observation, the apparent differences do not reproduce as a #312 regression: Nemotron dynamic context is 3/4 main vs 4/4 #312; Gemini employee lookup is 1/4 main vs 2/4 #312. This is consistent with model variability, with no negative signal attributable to the IR change.

Viewer runs: [main initial](http://pmgb300ws-0375.ipp9c1.colossus.nvidia.com:5001/eval/experiment/capability_20260910_064740_400605_p12), [#312 initial](http://pmgb300ws-0375.ipp9c1.colossus.nvidia.com:5001/eval/experiment/capability_20260910_065341_447989_p12), [main rerun](http://pmgb300ws-0375.ipp9c1.colossus.nvidia.com:5001/eval/experiment/capability_20260910_070029_785280_r3_p6), [#312 rerun](http://pmgb300ws-0375.ipp9c1.colossus.nvidia.com:5001/eval/experiment/capability_20260910_070029_538148_r3_p6).

The eval records report about 1.88M total input+output tokens including probes and follow-ups, so I stopped live calls there. The subsequent review-only fix at `0ec74617` changes fail-closed projection of incomplete/filtered tool batches and adds serialization coverage; it was validated locally with **773 passed** plus ruff and diff checks.

sklinglernv (commented):
[Codex Take] The core design is right: `LLMResponse` should be the canonical provider-produced assistant turn, with provider messages and tool executions derived from it. Separating requested tool calls from linked execution events removes duplicated, competing representations.

Three replay boundaries still need completion:
- Reasoning-backed structured output must retain a durable replayable answer.
- A missing execution in a projection cannot be treated as proof that the call was not executed; batch lifecycle records should exist before nested generation can occur.
- `llm_state` is captured but not consumed during replay, so exact provider replay remains incomplete pending the compatibility gate.

I would keep this architecture and fix these lifecycle/replay invariants before treating the refactor as complete.

furgalep:
@sklinglernv Thanks — your structured-output comment identified a real resume boundary, even though I don't think copying `reasoning` into `content` is the safe fix.

#312 now states the durable contract accurately: the source JSON survives in exact provider `content` **or** provider-exposed `reasoning` (`6063395c`). Keeping those fields distinct avoids claiming the provider emitted ordinary assistant text and avoids replaying the same bytes twice when opaque/native reasoning replay is available.

The projection belongs in stacked #311: UnifiedLLM carries the serialized reasoning through the provider edge and demotes it to ordinary assistant text when native replay is unavailable. I added a regression there (`045e432b`) that serializes and reloads a reasoning-only structured response, confirms the live `parsed` object is absent, and verifies the exact JSON is still sent as assistant context.

I reopened the thread so you can confirm that split addresses the underlying concern.

The other requests are also covered: the capability A/B is posted above, live NVIDIA-routed GPT and Anthropic cache reuse was verified with trailing dynamic context, and the incomplete linked tool-batch issue was fixed in `0ec74617`.

src/nooa/unifiedllm/unifiedllm.py
furgalep: Reopened: the proposed reasoning-to-content copy would break exact provider fidelity, but the resume concern is valid. #312 now documents that source JSON persists in content or reasoning (6063395c), and stacked #311 now has a serialize/reload-to-replay regression (045e432b) proving reasoning-only structured JSON is demoted to assistant text at the provider edge. Details: https://github.com/NVIDIA-NeMo/labs-OO-Agents/pull/312#issuecomment-5614724212

sklinglernv (approved):
Thanks for doing this! LGTM now with the upcoming changes.

src/nooa/unifiedllm/unifiedllm.py
furgalep: Fixed in ccca8770 and simplified in 8de1b22d. The canonical response still preserves the provider truth (`content` remains empty and the source JSON remains in serialized `reasoning`), but #312 now carries that reasoning privately through context formatting and demotes it to ordinary assistant text at the final UnifiedLLM boundary. The regression serializes/reloads a reasoning-only structured response with no live `parsed` value, then asserts the JSON reaches the next Completion request. Opaque-only state remains withheld unless a stacked provider gate recognizes it. This makes #312 independently resume-safe without duplicating reasoning into `content` or depending on #311.

furgalep:
@coderabbitai review

PR #315, opened 2026-09-09
feat(strategy): backport per-method LLM aliases to main

furgalep:
## Summary

- backport the reviewed per-method LLM alias implementation from #251 onto current `main`
- resolve `@strategy(llm="alias")` and call-site string aliases lazily, with per-agent caching
- retain precedence of call-site override, method strategy override, then agent default
- preserve the validation and isolation hardening from the original review round

## Context

#251 was accidentally merged into `dev/tui` instead of `main`. This PR replays its three commits directly on current `main`, giving downstream consumers such as NeMo Gym a focused revision to pin. After this merges, `dev/tui` should be rebased onto `main` so the duplicate patch is dropped.

Original commits: `75e1071`, `cb8c7c9`, `74a2cdd`.

## Validation

- `pytest tests/test_method_llm_callable.py -q` — 43 passed
- NeMo Gym NOOA adapter suite against this branch — 106 passed
- `git diff --check origin/main..HEAD`

## Summary by CodeRabbit

* **New Features**
  * Strategy-level and call-level LLM overrides now support registry aliases and LiteLLM model strings, in addition to configured clients and supported callables.
  * LLM selections are resolved when needed, improving startup behavior, and reused appropriately for subsequent calls.
  * Standalone strategies can use aliases and model strings, while callable overrides remain available only for agent-bound strategies.
* **Documentation**
  * Updated guidance explains supported override formats, resolution timing, caching, and standalone strategy restrictions.

### DIFF
--- a/docs/concepts/strategies.md
+++ b/docs/concepts/strategies.md
@@ -118,6 +118,10 @@ class ResearchAgent(Agent, llm=fast_llm):
         ...
 ```
 
+The `llm=` override also accepts a registry alias or model string
+(``@strategy(PredictStrategy(), llm="gpt-5-mini")``) — resolved lazily on the
+first call per instance, so declaring it constructs no client at import time.
+
 LLM overrides may also be supplied per instance or per call. Keep model routing
 out of the public method contract unless the application truly needs callers to
 choose it.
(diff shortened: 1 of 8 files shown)


sklinglernv (approved):
LGTM

PR #318, opened 2026-09-11
feat(llm): store ordered turns and replay compatible reasoning

furgalep:
Published head: `46d27810`. Latest full offline suite: **7,633 passed**. Wren independently accepted the three new replay fixes; prior review records below identify their own revisions.

## Summary

UnifiedLLM returns an `LLMResponse`; callers pass that object back in the next message list. UnifiedLLM handles compatible native reasoning and portable text reasoning. To edit history, replace the response with a plain dictionary; that replacement has no native state.

This replaces #310/#311 and builds on merged #312. **There is no separate turns lookup, message ID key, cache marker, or new cache policy in this PR.** Cache boundaries belong to #319.

```python
reply = await client.acall(history)
history.append(reply)
# Append matching results if reply requests tool calls.
next_reply = await client.acall(history)

# Editing deliberately gives up native replay for this turn.
history[index] = {**dict(history[index]), "content": "edited answer"}
```

## Why

Flattening a provider response into independent text and tool-call lists loses the order and grouping needed to replay it. The superseded stack reconstructed that information using ledgers, batch identities, fingerprints, and private dictionary carriers.

Ordered parts retain it. The response travels back to UnifiedLLM unchanged, so no runtime lookup or per-request content comparison is needed. Consumers read its public mapping view, not provider fields. The JSON-only relay is the one boundary that compares returned public messages to preserve unchanged response objects.

This simplifies the superseded reasoning stack; it is not a claim of fewer lines than main without reasoning support.

1. **`llm_types.py`, `_immutable_json.py`** store ordered immutable text, reasoning, and tool-call parts. Derived public views avoid maintaining another authoritative message. Public mapping reads and archive serialization are distinct: one omits native state, the other persists it. Large immutable strings are shared.
2. **`unifiedllm/chat_parts.py`, `response_parts.py`, `replay_state.py`** capture provider output and project it back at dispatch. Native data stays attached to its owning part and is replayed only for a compatible API/provider/model. Readable reasoning remains portable. Capture/projection docstrings explain ownership, ordering, compatibility, and why these operations belong inside UnifiedLLM.
3. **UnifiedLLM clients** accept response objects alongside ordinary dictionaries. Chat and Responses retain separate adapters because their wire messages, tool schemas, and parsing differ, while sharing the durable caller interface. Raw Chat `reasoning_content` stays caller-specified; automatic text demotion applies to captured responses, not arbitrary dictionaries.
4. **EventBase and formatter** separate responsibilities. EventBase supplies identity, lifecycle, emptiness, and searchable fields. The formatter recognizes the public LLMResponse type with one small TypeGuard; assistant replay hooks no longer live on every event. Unchanged rendering preserves the response object; truncation or omitted calls produces a portable dictionary. Display and budgeting still use public text/tool views.
5. **Runtime and middleware** carry these messages without interpreting provider state. Read-only response access prevents accidental edits; replacing a list element is explicit. Only leading system messages become Responses instructions; later system messages stay in conversation order.
6. **NeMo Relay** projects public JSON and reconciles returned entries by index and public equality. Unchanged entries regain the original objects; edited/shifted entries remain dictionaries. Native state never enters relay JSON.
7. **Storage and observability** persist parts in SQLite/snapshots while excluding opaque state from public journal/search/OTLP export. Readable reasoning remains available to OTLP. Unsupported journal objects use type-only placeholders, not str/repr. Excessively deep or cyclic values are redacted in full rather than bypassing the scrubber.
8. **`unifiedllm/errors.py`** collects the two framework exceptions: malformed replay data is terminal, while empty model output can follow the existing retry policy. Malformed tool-call dictionaries and missing Responses fields are validated before incidental AttributeError/KeyError/model-validation failures can enter generic recovery.
9. **Usage and tests** keep normalized input/output/cache-read/cache-write/reasoning tokens and estimated cost on the response; no UI is added. Contract tests cover wire order, compatibility, persistence, collapse, truncation, relay identity, malformed inputs, tracing, and ownership. Old reconstruction-mechanism tests disappear with those mechanisms.

## Migration and limits

- `dict(response)` is portable public data; `response.model_dump()` is an archive. Arbitrary `json.dumps(history)` needs an explicit public projection.
- Public edits discard native state. Request containers are detached where downstream mutation requires it; canonical native string payloads are not deep-copied each round trip.
- Unknown capture routes warn and retain readable text. Malformed supported native data fails explicitly. Empty public call IDs are accepted unless native state needs a binding; duplicate nonempty IDs fail.
- Scope is API style + resolved provider + exact model, **not endpoint/account isolation**.
- Legacy flat archives load readable text/reasoning/calls but lose their old opaque sidecar once.
- `collapse()` archives the requested range. ToolResult is nested in ToolCallEvent, but the original LLMResponse is separate; the formatter omits incomplete batches rather than expanding collapse ranges atomically.
- No TUI, effort-level controls, or retention toggle changes.

## Validation

Reviewed revision `6d13a313`:

- Full root + memory + benchmark offline suite: **7,726 passed, 21 skipped, 235 deselected** (255 seconds).
- Focused UnifiedLLM/context/tracing suite: **1,191 passed**.
- Wren accepted the fixes after independent review and **2,419 scoped tests passed**.
- Ruff formatting/lint and whitespace checks pass.
- All seven human comments and five CodeRabbit findings have replies. No additional inference credits were used for this cleanup.

Prior live evidence on the object-in-list stack is recorded in #319: OpenAI, Claude Sonnet, and Gemini each retained state across SQLite close/reopen and reused the stable prefix with changed trailing context. Those live runs predate these review fixes; they are not presented as fresh runs of this revision.

See [the published design and migration walkthrough](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/46d2781056a7b00ce9ae2ecfaae45a7e54089667/docs/assistant-turn-replay.md). This revision also updates its EventBase and raw-dictionary explanation.

## Latest review follow-up

At `e43727e1`:

- The formatter omits duplicate-ID tool-call batches with the existing warning. One result must not satisfy two calls; archives remain readable and successful gateway responses with missing IDs can still be captured.
- The OTLP journal stores readable reasoning in content-addressed blocks and the viewer resolves those blocks. This avoids publishing the same large text repeatedly while keeping plain reasoning visible; opaque provider state is still excluded from OTLP.
- Removed overwritten test expectations. New behavior was tested failing before implementation. The affected suite passes **2,530 tests**, and Wren accepted this exact head after independent review.

No additional live inference was used for these fixes.

## Second-pass review fixes

At `492e7ded`, nested formatting honors visibility rules and LLMResponse supplies readable values through the existing formatting protocol. This prevents provider state from appearing in nested context, traces or search while restoring the visible answer; archives still keep the state for replay. Edited responses own their metadata dictionary. Malformed Responses inputs fail before transport, and signature scrubbing covers the reported shapes. The transport and replay tests now assert the behavior their names promise; the existing display regression moved into CI's test directory. The authoring skill's registry instructions match the loader.

Validation: **7,624 passed, 6 skipped, 298 deselected, 3 expected failures** in the full offline suite. Behavioral regressions were observed failing before fixes; the transport test also fails under removal of owned-client injection. No live-provider rerun or inference spend for this revision. Wren independently accepted `492e7ded`: all reported leak probes clean, display restored, mutation test catches removed transport reuse, and 3,562 scoped tests passed.

## Follow-up comment fixes

At `f877def5`, nested formatting checks instance-level hiding so a hidden field stays hidden inside lists and dictionaries. Search formatting keeps scalar strings unquoted so exact regex searches still match. Responses accepts leading system text blocks by joining their text, and rejects unsupported blocks with a clear error instead of failing during string joining.

All **11 regression cases failed before implementation** and pass after the fixes. Full offline suite: **7,635 passed, 6 skipped, 298 deselected, 3 expected failures**. Ruff checks pass. Wren accepted this head after independent leak probes and **3,269 tests**. The three new CodeRabbit findings have replies and are resolved. No live inference was used for this cleanup.

## Severin review fixes — what and why

At `0da40b5a`:

- Preserve Chat's `reasoning_content` source field on compatible replay; carrying the same text in answer content is not the same wire contract. One marker on the existing part avoids a second text copy.
- Keep usable Responses answers/refusals when a valid built-in item is outside our replay format. Warn and drop native state for the whole turn, rather than discard a successful answer or replay an incomplete native turn.
- Preserve the validated structured result when think-tag cleanup only separates reasoning from the answer.
- Count extracted Responses instructions in calibration: the provider bills them even though they are outside the `input` list.
- Serialize nested benchmark trajectory values as public JSON objects, not repr strings, without exposing provider state.

TDD regressions cover each behavior. Full offline suite: **7,654 passed, 6 skipped, 302 deselected, 3 expected failures**. Four bounded live Hub tool loops passed after SQLite reopen and checked exact source-field replay. The opt-in omitted-field A/B tests also completed; acceptance alone is not evidence of retained reasoning. Release-only wiring is [#324](https://github.com/NVIDIA-NeMo/labs-OO-Agents/pull/324) on #319, using the existing pre-draft release gate.

## Inline signatures, public input blocks, and reasoning-only replies

At `46d27810`, all three colleague findings were reproduced and fixed:

- Capture strips the reserved `__thought__` encoding independently of route identity, before deciding whether native replay is permitted. Public IDs and incompatible requests cannot retain the signature. Raw dictionaries containing this encoding now raise a clear error; pass captured `LLMResponse` objects for protected replay. This also changes legacy raw-dictionary histories containing signed IDs.
- Responses translates public assistant content blocks and tool calls directly. A valid list of text blocks is no longer incorrectly validated as a normalized Chat provider response; block content is retained without mutating the caller's data.
- Compatible Chat replay retains `reasoning_content` even when there is no answer or tool call, including a length-stopped reasoning-only reply.

TDD: the initial nine regressions produced **7 failures and 2 passes** before implementation. The final UnifiedLLM suite passed **831 tests**; the full offline run passed **7,633**, with **7 skipped, 302 deselected and 3 expected failures**. HTTP-mocked sync/async tests include SQLite close/reopen and incompatible-model replay. Wren independently accepted the commit after reproductions and **206 tests** in the four touched modules. No live inference was used for this increment.

Limits: LiteLLM strips inline signatures on destination aliases it does not recognize as Gemini; the documented Hub Gemini aliases retain them. Direct providers outside the supported scope set still replay captured readable reasoning as ordinary text instead of the original `reasoning_content` field; this is a separate follow-up, not fixed here. The expanded offline wire matrix is in #325.

## Summary by CodeRabbit

- **New Features**
  - Added structured, read-only assistant response history that preserves unedited responses across calls.
  - Improved reasoning and tool-call replay across supported providers, with portable fallbacks when native state cannot be reused.
  - Added configurable Responses API storage and inclusion settings.
  - Added clear errors for invalid or incomplete reasoning responses.

- **Bug Fixes**
  - Improved persistence, caching, relay handling, and history collapsing.
  - Prevented private provider state and sensitive reasoning data from appearing in searches, traces, logs, or serialized messages.

- **Documentation**
  - Added guidance for response replay, editing, portability, and cached context behavior.


src/nooa/context_blocks/events.py
```diff
@@ -91,6 +91,33 @@ class EventBase(BaseModel):
...
+        """Optional ready-to-replay message; generic events use normal formatting."""
+        return None
+
+    @property
+    def is_replay_turn(self) -> bool:
+        """Whether the event supplies a complete model turn for replay."""
+        return False
+
+    @property
+    def replay_tool_calls(self) -> tuple:
```
furgalep: This for sure feels like abstraction leak. Are the others also? What's the mental model of EventBase?

src/nooa/unifiedllm/_message_utils.py
```diff
@@ -0,0 +1,24 @@
...
+"""Provider-neutral dictionary helpers used only at the UnifiedLLM boundary."""
+
+from __future__ import annotations
+
+from typing import Any
+
+LLM_STATE_KEY = "_nooa_llm_state"
+
+
+def demote_reasoning_text(message: dict[str, Any], reasoning: str | None) -> None:
```
furgalep: I don't get the point of this.

src/nooa/unifiedllm/chat_parts.py
```diff
@@ -0,0 +1,202 @@
...
+    _scope_provider,
+    _tool_call_state,
+    _warn_unknown_fields,
+    logger,
+    opaque_item,
+)
+from .response_parts import _capture_summary, _require_encrypted_reasoning, _restore_summary
+
+
+def capture_chat_parts(message: Any, scope: str | None) -> tuple[AssistantPart, ...]:
```
furgalep, line 27: This should have a doc string that explains what it does and why.

src/nooa/unifiedllm/chat_parts.py
```diff
@@ -0,0 +1,202 @@
...
+        )
+        return tuple(part.model_copy(update={"native": None}) for part in parts)
+    if any(isinstance(part, ToolCall) and not part.id for part in parts) and any(
+        part.native for part in parts if not isinstance(part, AssistantText)
+    ):
+        raise ReasoningReplayError("Native reasoning requires nonempty tool call ids.")
+    return tuple(parts)
+
+
+def project_chat_turn(turn: LLMResponse, scope: str | None) -> tuple[dict, dict[str, str]]:
```
furgalep, line 183: This should have a docstring that explains what it does and why.

src/nooa/unifiedllm/replay_state.py
```diff
@@ -0,0 +1,321 @@
...
+_INLINE_THOUGHT_SIGNATURE_SEPARATOR = "__thought__"
+_SUPPORTED_PROVIDERS = {
+    "openai",
+    "azure",
+    "anthropic",
+    "gemini",
+}
+
+
+class ReasoningReplayError(RuntimeError):
```
furgalep: Is this the only error in Unified LLM? If there are others, should we put them together in an errors.py?

src/nooa/unifiedllm/unifiedllm.py
```diff
@@ -1715,6 +1749,45 @@ def _patched_remove(self, model, messages, tools=None):
 
 
 class CompletionClient(UnifiedLLM):
+    @staticmethod
```
furgalep, line 1759: Do we still need different clients for completion and responses?

src/nooa/unifiedllm/unifiedllm.py
```diff
@@ -2443,204 +2349,126 @@ async def _make_call():
...
+            ):
+                # Public direct-call input has no native replay authority.
+                from .chat_parts import capture_chat_parts
 
-        instructions = "\n\n".join(instructions_parts) if instructions_parts else None
-        return transformed, instructions
+                turn = LLMResponse(parts=capture_chat_parts(msg, None))
+                transformed.extend(response_parts.project_turn(turn, None))
+            else:
+                item = copy.deepcopy(msg)
```
furgalep, line 2529: Are these deep copies necessary?

src/nooa/context_blocks/events.py
furgalep: Agreed—the assistant-specific hooks were an abstraction leak. EventBase now describes a durable record: identity/type, lifecycle, emptiness, and searchable fields. I removed its replay hooks and restored the formatter's small `_is_llm_response` TypeGuard. The formatter owns event-to-message semantics; only LLMResponse owns assistant replay operations. Generic events have a regression asserting they do not expose those assistant APIs.

Published in `6d13a313` and accepted by Wren (2,419 independent scoped tests passed). Full offline suite: 7,726 passed, 21 skipped.

src/nooa/unifiedllm/chat_parts.py
furgalep: Added architectural docstrings to both Chat and Responses capture functions. They explain why capture stores ordered public parts with native data on the owning part, why mutable provider containers are detached once, what information the Chat transport has already normalized, and when unknown routes keep only readable text. Malformed supported data is an explicit contract error, not silent state loss.

Published in `6d13a313` and accepted by Wren (2,419 independent scoped tests passed). Full offline suite: 7,726 passed, 21 skipped.

src/nooa/unifiedllm/_message_utils.py
furgalep: Removed `_message_utils.py`. It was folding a raw dictionary's `reasoning_content` into `content`, but a caller-written Chat dictionary should retain that explicitly supplied wire field. Automatic portable reasoning belongs to projection of a captured LLMResponse, not a rewrite of arbitrary dictionaries. The new regression checks both behaviors together: raw reasoning_content survives unchanged, while a captured turn still shares readable reasoning as portable text.

Published in `6d13a313` and accepted by Wren (2,419 independent scoped tests passed). Full offline suite: 7,726 passed, 21 skipped.

src/nooa/unifiedllm/chat_parts.py
furgalep: Added architectural docstrings to both projection functions. They explain how UnifiedLLM reconstructs provider messages from stored parts, sends native state only for compatible turns, preserves readable reasoning otherwise, and allocates request containers while sharing immutable strings. Responses projection also explains why preserving order/boundaries enables prefix reuse without itself guaranteeing a cache hit or selecting cache policy.

Published in `6d13a313` and accepted by Wren (2,419 independent scoped tests passed). Full offline suite: 7,726 passed, 21 skipped.

src/nooa/unifiedllm/replay_state.py
furgalep: There are two framework-defined exceptions in UnifiedLLM: ReasoningReplayError and EmptyContentError. Both now live in `unifiedllm/errors.py` and are exported from `nooa.unifiedllm`. The distinction remains intentional: malformed replay data is terminal; an empty response can follow the existing retry policy. Provider/SDK exceptions are not duplicated into our own hierarchy.

Published in `6d13a313` and accepted by Wren (2,419 independent scoped tests passed). Full offline suite: 7,726 passed, 21 skipped.

src/nooa/unifiedllm/unifiedllm.py
furgalep: I would keep CompletionClient and ResponsesClient separate here. Both already exist on main; their wire messages, tool schemas, response parsing, and capture/projection paths differ. Merging them now would replace class dispatch with API-style conditionals at each stage. The durable caller interface is shared: both accept history containing LLMResponse and return LLMResponse. Consolidating the smaller ReasoningCompletionClient variant can be considered separately without coupling it to this replay change.

No client consolidation is included in this PR.

src/nooa/tracing/_litellm_journal.py
```diff
@@ -69,6 +70,12 @@ def _msg_to_dict(msg: Any) -> dict:
         return {"raw": str(msg)}
 
 
+def _safe_msg_to_dict(msg: Any) -> dict:
+    """Normalize one provider message and remove provider-only opaque state."""
+    scrubbed, _ = scrub_value(_msg_to_dict(msg))
```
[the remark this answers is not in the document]
furgalep: Verified and fixed. Unsupported message objects now produce only an `unsupported_type` placeholder; neither str nor repr is used. Chat and Responses output extraction share that normalization. The regression includes an object with encrypted_content and no model_dump, and str/repr methods that raise if called, plus a SimpleNamespace case.

Published in `6d13a313` and accepted by Wren (2,419 independent scoped tests passed). Full offline suite: 7,726 passed, 21 skipped.

src/nooa/unifiedllm/unifiedllm.py
furgalep: I checked both copies and kept the existing ownership contract. This branch handles caller-owned raw dictionaries, whose nested content blocks are rewritten into Responses input blocks and may be modified by the downstream SDK. A shallow dictionary copy would leave those nested containers shared; cache_control is also caller-owned mutable metadata. `test_responses_passthrough_still_detaches_nested_public_content` verifies isolation for user, assistant, and tool messages.

This is not copying retained opaque payloads each round trip: canonical LLMResponse objects take the parts-projection path, which shares immutable string leaves. Python deepcopy also shares ordinary strings, including image strings. I added those reasons beside the copies; no new copies were introduced.

Published in `6d13a313` and accepted by Wren (2,419 independent scoped tests passed). Full offline suite: 7,726 passed, 21 skipped.

src/nooa/tracing/_secret_scrubber.py
```diff
@@ -214,19 +248,38 @@ def scrub_value(value: Any) -> tuple[Any, int]:
...
+        # OpenInference records LLM inputs as JSON string span attributes.
+        # Only parse objects/arrays. Matching the prefix avoids allocating a
+        # stripped copy of a potentially large tool output just to inspect it.
+        if not re.match(r"\s*[\[{]", value):
+            return scrub_string(value)
+        try:
+            decoded = json.loads(value)
+        except (json.JSONDecodeError, TypeError):
+            return scrub_string(value)
+        decoded, json_count = scrub_value(decoded)
```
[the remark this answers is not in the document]
furgalep: Verified and fixed. The public scrub_value entry catches RecursionError around the entire recursive operation, including JSON decoding and recursive traversal, and replaces the complete value with REDACTED. The recursive worker does not catch at a nested leaf. Regressions cover deep JSON, deep mappings, cycles, and a failure after decoding.

Published in `6d13a313` and accepted by Wren (2,419 independent scoped tests passed). Full offline suite: 7,726 passed, 21 skipped.

src/nooa/unifiedllm/response_parts.py
```diff
@@ -0,0 +1,158 @@
...
+
+def _capture_text(blocks: list[dict], separator: str) -> str:
+    """Move text into the public part; retain only block metadata and lengths."""
+    texts = []
+    for block in blocks:
+        text = block.pop("text")
+        if not isinstance(text, str):
+            raise ReasoningReplayError("Responses text must be a string.")
+        texts.append(text)
+        block["_text_length"] = len(text)
```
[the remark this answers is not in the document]
furgalep: Verified and fixed. Message content, text, function name, and arguments now receive explicit type checks with defaulted reads before part construction. Malformed content is also rejected in the output-shape scan, before it can raise TypeError while iterating. The regression covers missing/null/wrong-type fields as raw dictionaries and Pydantic SDK-shaped objects serialized with exclude_none=True.

Published in `6d13a313` and accepted by Wren (2,419 independent scoped tests passed). Full offline suite: 7,726 passed, 21 skipped.

src/nooa/unifiedllm/replay_state.py
```diff
@@ -0,0 +1,321 @@
...
+def reject_native_message(message: dict[str, Any], scope: str | None) -> None:
+    """Raw dictionaries are portable input, not a provider-state replay API."""
+    private_keys = {LLM_STATE_KEY, "reasoning_items", "thinking_blocks", "provider_specific_fields"}
+    nodes = [message]
+    content = message.get("content")
+    if isinstance(content, list):
+        nodes.extend(block for block in content if isinstance(block, dict))
+    for call in message.get("tool_calls") or []:
+        nodes.append(call)
+        nodes.append(call.get("function", {}))
```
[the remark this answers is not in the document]
furgalep: Verified and fixed. Raw tool_calls must be a list, each entry a dictionary, and each function a dictionary before any dereference. Invalid values raise terminal ReasoningReplayError. Parameterized tests cover seven malformed shapes on both clients, sync and async, and assert that none of the four provider transport functions is called.

Published in `6d13a313` and accepted by Wren (2,419 independent scoped tests passed). Full offline suite: 7,726 passed, 21 skipped.

tests/context_blocks/test_cached_renderer.py
```diff
@@ -220,6 +222,58 @@ def find_user_event_msg(out: list[dict]) -> dict:
...
+                    metadata=BlockMetadata(tag="2"),
+                    event=turn,
+                ),
+                _dynamic_block("live_state", "version two"),
+            ],
+            block_formatter=CachedBlockFormatter(),
+            provider_formatter=OpenAIProviderFormatter(),
+        ).output
+
+        assert first[:-2] == second[: len(first) - 2]
```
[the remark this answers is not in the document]
furgalep: Verified and fixed in #318: the assertion now compares first[:-1] against the matching prefix of the second render, covering the historical user event too. In rebased #319, the separate metadata boundary is also present, so that test explicitly verifies the boundary and excludes only boundary + live suffix; all history remains included.

Published in `6d13a313` and accepted by Wren (2,419 independent scoped tests passed). Full offline suite: 7,726 passed, 21 skipped.

furgalep:
Published the reviewed fixes at `6d13a31387c8f5a1402bed115828cae34257e061` through the GitHub API. Commit IDs and full file-tree hashes match the locally tested revisions exactly; no code changes were made during publication. Wren has accepted this revision. The publication-pending notices are removed and the walkthrough links point to the published head.

@coderabbitai Please recheck the existing findings against this published revision; the earlier replies describe the fixes and regression coverage.

furgalep:
A capped live Gemini effort probe exposed an unsigned thinking block that the capture path incorrectly treated as malformed signed thinking. Fixed separately in ecb1f6c0 + b8f0c27f: missing/null signatures yield portable reasoning text, while signed neighboring text/tool-call parts retain their own state. Empty or non-string signatures still fail. Regression cases include archive round trips and compatible/unknown scopes. Wren reviewed the correction; 64 focused tests pass (2,370 affected tests passed before the null-field follow-up). A real Hub call on the corrected core now completes: 15 input / 210 output tokens, 208 reasoning tokens, readable reasoning retained and one native-bearing part. The reasoning-level feature remains a separate follow-up PR.

src/nooa/context_blocks/formatter.py
```diff
@@ -341,11 +331,11 @@ def _event_blocks_to_messages(
     messages: list[RenderedMessage] = []
     for block in blocks:
         event = block.event
-        if _is_llm_response(event) and _is_replayable_tool_call_turn(event):
+        if _is_llm_response(event) and event.replay_tool_calls:
```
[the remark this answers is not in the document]
furgalep: Fixed in e43727e1. The formatter now treats duplicate call IDs (including two empty IDs) as an invalid visible execution batch and uses the existing warn-and-omit path, so one execution cannot be replayed twice. The captured/archived response remains intact. I deliberately did not reject every empty ID at capture: some compatible gateways omit IDs, and capturing a successful response should not itself fail.

TDD: reproduced four failures first (current and legacy archives, repeated nonempty and empty IDs), then added the guard. All four regressions and 2,530 affected tests pass. Wren independently reviewed and accepted this fix.

furgalep:
Review audit complete at e43727e1, including human threads, CodeRabbit inline findings, and findings in review bodies.

Outside-diff finding: **empty tool-call IDs**. I did not adopt blanket rejection at capture, because successful gateway responses can omit IDs. The underlying replay bug is fixed at the existing formatter batch check instead: duplicate IDs, including repeated blanks, warn and omit the ambiguous batch. Original responses and legacy archives stay readable. Four regression cases were observed failing before the fix and now pass. See the inline duplicate-ID reply.

Earlier review-body nits are also addressed:
- **Repeated reasoning in the journal:** input and output reasoning are stored as content-addressed blocks, deduplicated across repeated calls, and resolved back to text by the viewer. Plain-text reasoning still reaches OTLP; opaque state remains excluded. The regression failed before the fix, then passed; it covers repeated input/output reasoning, exact text recovery, old inline records, and missing blocks.
- **Dead expected values:** removed the four overwritten per-shape assignments; kept the actual wire-order assertion.

The seven earlier human discussions were checked against current code and their existing answers, then marked resolved. No further code changes were needed for those or the five previously resolved CodeRabbit findings.

Validation: 2,530 affected tests passed (6 skipped, 69 deselected); lint/format/whitespace checks pass. Wren accepted this exact head with an independent 2,585-test run. No live inference credits used for this cleanup.

furgalep:
Second-pass review update: please hold merging while I finish verification. Wren reproduced two issues in the current head: nested formatting/search can expose provider state, and pformat(LLMResponse) loses the readable answer. Both now have failing-then-passing regressions and a local fix at 492e7ded; full offline validation and independent re-review are in progress. The same change also addresses metadata aliasing, Responses input errors, scrubber signature keys, and strengthens the transport/render/replay tests. I will post the validated head and complete dispositions before requesting re-review.

furgalep:
Second-pass fixes are now published at **492e7ded**; #319 and #320 are rebased onto it. The earlier readiness statement applied to the preceding revision. Wren has been asked to recheck these new heads; I am not claiming renewed independent approval yet.

What changed and why:

- **B1/B2 — nested privacy and visible output:** nested pformat now honors repr/exclude/hidden fields, just as top-level formatting does. LLMResponse supplies its readable content, reasoning and tool calls through the existing instance-values protocol; property metadata preserves content's no-truncation rule. Search retains live nested objects until public formatting instead of serializing their archive representation. This closes the reported list/dict/PythonOutput/context/OTLP/search paths without provider-aware consumer branches. Archive storage still preserves native state.
- **S1 — edited metadata:** public edits copy the small metadata dictionary so updating the replacement cannot mutate the original event's top-level metadata. This does not copy history or provider blobs.
- **S2 — Responses input errors:** invalid message elements/content blocks and missing or non-string tool_call_id now raise actionable TypeError/ValueError before transport.
- **S3/S4:** scrubber covers inline_thought_signature and signature; Responses call/acall docstrings describe the object-in-list interface and leading-system-only instruction handling.
- **T1:** the completion transport test now distinguishes the owned pool from the fallback pool. Temporarily deleting the two client-injection lines makes both sync and async regressions fail; restoring them passes.
- **T2/T4:** removed vacuous assertions, assert response identity plus safe public projection, actually render dynamic expressions, require a non-null replay scope, reject every incompatible reasoning item (not just one exact dict), and check the entire resumed tool-result sequence including call IDs.
- Moved the previously undiscovered max-string event test into tests/agentdoc. Corrected the skill registry paragraph against the actual loader; its existing compatibility field is supported by NOOA's loader. Clarified the response alias, formatter subclass, and copy rationale. #320 now loads its example registry explicitly before using its alias.

Scope dispositions: T3's pre-existing relay-header test/behavior and unrelated dead helpers stay separate. The harmless calls-only public empty-string versus Chat wire null distinction is unchanged. Further stylistic cleanup of existing exporter helpers and terminology is not part of this correctness patch.

TDD evidence: B1/B2 produced **12 failing tests** before the fix; S1–S3 produced **23 failing cases** before their fixes. The combined 51 targeted tests pass. Full #318 offline suite: **7,624 passed, 6 skipped, 298 deselected, 3 expected failures**. Rebased #319 affected suite: **3,341 passed**. Rebased #320 final affected suite: **3,357 passed** (including its additional mutation-validation fix). Ruff/format and whitespace checks pass. No paid provider calls were made.

furgalep:
Update: Wren has now independently accepted all three published heads: #318 `492e7ded`, #319 `d34b06c8`, #320 `bd7ea3e7`. They verified all reported leak paths are clean, visible output and the content-length rule are restored, and removing owned transport reuse now fails the tests. Independent scoped results: 3,562 tests for #318, 3,234 for #319, and 2,758 for final #320. My final runs also passed (full #318: 7,624; affected #319: 3,341; affected #320: 3,357). The hold requested above is cleared for human re-review. No unresolved GitHub review threads at the final check.

src/nooa/agentdoc/_pformat.py
```diff
@@ -1350,9 +1357,24 @@ def _format_nested_instance(
...
+    from nooa.agentdoc._visibility import is_hidden_field
+
+    excluded = {f.name for f in type_info.fields if not f.repr}
+    excluded.update(
+        name for name, field in getattr(obj_type, "model_fields", {}).items() if field.exclude
+    )
+    field_names = [
+        f.name
+        for f in type_info.fields
+        if f.name in values and f.name not in excluded and not is_hidden_field(obj_type, f.name)
```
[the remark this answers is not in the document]
furgalep: Fixed in f877def5. Both nested visibility filters now check the instance, so per-instance hidden fields cannot bypass the declared-field or extra-value path. Regression tests cover list/dict nesting, both field paths, and an untouched instance. Tests failed before the fix; the full offline suite now passes 7,635 tests. Wren independently rechecked the leak probes.

src/nooa/runtime/event_manager.py
```diff
@@ -459,13 +460,9 @@ def _get_searchable_text(self, event: EventBase) -> str:
...
 
-        # Opaque replay state is durable but not a conversational/search field.
-        for value in event.model_dump(exclude={"llm_state"}).values():
+        for value in event.searchable_fields().values():
             if value is not None:
-                if isinstance(value, list):
-                    parts.append(" ".join(str(item) for item in value))
-                else:
-                    parts.append(str(value))
+                parts.append(pformat(value))
```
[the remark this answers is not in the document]
furgalep: Fixed in f877def5 using pformat(value, unquote_strings=True). Scalar search text stays literal, while nested values still honor visibility rules. Three regressions exercise anchored regex searches for ordinary text, multiline text, and literal quotes; all failed before the fix and now pass.

src/nooa/unifiedllm/unifiedllm.py
```diff
@@ -2443,204 +2350,141 @@ async def _make_call():
...
-            # Legacy OpenAI format: tool result messages
-            if msg.get("role") == "tool":
-                # Extract output text: content may be a string OR a list of blocks
-                # (position-based cache injection converts strings to block arrays)
+            msg = dict(original)
+            replay_state.reject_native_message(msg, state_scope)
+            if isinstance(msg.get("content"), list) and any(
+                not isinstance(block, dict) for block in msg["content"]
+            ):
+                raise TypeError("Message content blocks must be dictionaries.")
```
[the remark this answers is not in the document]
furgalep: Fixed in f877def5. Leading system text/input_text blocks are joined into text before instructions are assembled; unsupported blocks or non-string block text raise a clear ValueError. Tests cover both supported block types, subsequent system-message ordering, unchanged caller inputs, and four invalid shapes. All six regression cases failed before the fix and now pass.

src/nooa/unifiedllm/chat_parts.py
```diff
@@ -0,0 +1,235 @@
...
+                    raise ReasoningReplayError(
+                        "Encrypted Chat reasoning is unsupported for this provider."
+                    )
+                _require_encrypted_reasoning(block)
+                _restore_summary(block, part.text)
+            elif field != "thinking_blocks":
+                raise ReasoningReplayError("Unknown native Chat reasoning field.")
+            message.setdefault(field, []).append(block)
+        elif part.text:
+            text.append(part.text)
```
sklinglernv, line 243: Preserve native `reasoning_content` for compatible DeepSeek replay. A plain `reasoning_content` captured at lines 98-107 reaches this branch without `native`, so it is folded into ordinary assistant `content` even when the response is sent back to the same route. DeepSeek thinking-mode tool calls treat `reasoning_content` as a distinct history field and require prior values to be passed back; LiteLLM can mask the omission by injecting a blank field, but then the actual reasoning chain is absent from its native channel. The nonce experiment confirms that the text remains visible after being moved into answer content, not that the wire contract is preserved. Please retain enough source/capability information to restore `reasoning_content` for compatible destinations and reserve content demotion for incompatible routes. See https://api-docs.deepseek.com/guides/thinking_mode/.

src/nooa/unifiedllm/response_parts.py
```diff
@@ -0,0 +1,196 @@
...
+    opaque_item detaches mutable provider containers; frozen native data then
+    survives storage and repeated requests without copying its string payloads.
+    Unrecognized routes retain portable parts only. A summary-only reasoning
+    item makes the entire turn portable, since mixing a rewritten public turn
+    with retained native state would violate replay consistency. Missing or
+    malformed supported fields raise the terminal ReasoningReplayError.
+    """
+    unsupported = unsupported_responses_parts(output)
+    if unsupported:
+        raise ReasoningReplayError("Unsupported Responses output parts: " + ", ".join(unsupported))
```
sklinglernv: Do not make an otherwise successful Responses result unusable solely because it includes a valid built-in output item or refusal block. For example, a completed `web_search_call` followed by a usable text message reaches this check, raises `ReasoningReplayError`, and discards the final text that the previous client path returned. Responses output can legitimately contain server-side tool actions and `message.refusal` content. Please represent/pass through the supported item when replay is required, or preserve the usable outcome while explicitly marking the turn non-replayable, instead of aborting capture of the entire provider response.

src/nooa/unifiedllm/unifiedllm.py
```diff
@@ -2149,14 +2098,19 @@ def call(
                     else think_reasoning
                 )
 
-                response.content = cleaned_content
-                response.reasoning = combined_reasoning
+                response = response.replace_parts(
```
sklinglernv, line 2108: This cleanup loses an already validated structured result. `super().call(..., output_model=...)` sets `response.parsed`, but `replace_parts()` delegates to `model_copy()`, which explicitly resets `parsed` to `None`. Thus a direct `ReasoningCompletionClient` response such as `<think>...</think>{"answer":42}` is parsed successfully and then returned with `parsed is None`; the async branch has the same behavior. Please preserve the validated object when this transformation only separates reasoning from text, or re-run output-model validation after cleaning.

src/nooa/unifiedllm/unifiedllm.py
```diff
@@ -2323,68 +2278,24 @@ def _make_call():
...
-                raw_response=raw_response,
-                content=text_content,
-                parsed=parsed_content,
-                tool_calls=[],
-                finish_reason=_map_responses_finish_reason(raw_response),
-                reasoning=None,
-                usage=usage,
-                llm_state=_responses_llm_state(output),
+            _update_token_calibration(
+                effective_model, input_messages, usage, tools=api_params.get("tools")
```
sklinglernv: Include the extracted `instructions` in token calibration. The provider's reported input usage covers both `instructions` and `input`, but this estimate sees only `input_messages` and tools. A large leading system prompt plus a short user message therefore produces a severely inflated calibration ratio, affecting later `count_tokens()` estimates for this model. The async call at line 2363 has the same issue. Please pass instructions into the estimator, for example as a synthetic system message, or extend the calibration interface to account for them explicitly.

src/nooa/context_blocks/events.py
```diff
@@ -91,6 +95,20 @@ class EventBase(BaseModel):
...
+
+        Events are meaningful by default, including events without text fields.
+        Specialized durable records may distinguish observability from replay.
+        """
+        return False
+
+    def searchable_fields(self) -> dict[str, Any]:
+        """Public fields for search/debug export; consumers need not know their layout."""
+        # Keep nested objects intact so their display rules still hide private fields.
+        return self.__instance_values__()
```
sklinglernv, line 110: Keep this display-oriented projection from changing JSON export types. `nooa_bench.runner._write_trajectory()` expands `searchable_fields()` into its JSON payload and uses `json.dumps(..., default=str)`. Returning nested Pydantic objects here changes `ToolCallEvent.result` from a JSON object such as `{"tool_call_id":"c1","content":"actual result","result_status":"complete"}` into the repr string `tool_call_id='c1' content='actual result' ...`; consumers reading `event["result"]["content"]` then fail. I reproduced this against the current head. Please separate search/display values from JSON-safe public export, or recursively serialize public nested models at the trajectory export boundary while continuing to exclude hidden/private fields.

sklinglernv (commented):
Reviewed the current head (`f877def`). The ordered-parts model is a sound direction, but the inline findings below cover remaining compatibility and regression cases: native DeepSeek reasoning replay, valid Responses output shapes, structured-output preservation, token calibration, and benchmark trajectory serialization.

I revalidated after the latest updates: 132 focused tests pass, and direct probes confirm that unsigned thinking capture, caller-supplied `reasoning_content`, and `pformat(LLMResponse)` are now fixed.

furgalep:
Bounded open-model replay results (local correction 4a43e16a, pending independent review): DeepSeek V4 Pro, Kimi K3, GLM 5.3, and Qwen 3.5 397B all passed two-turn tool exchanges through NVIDIA Inference Hub. Each returned nonempty reasoning_content and a tool call; after SQLite close/reopen, a fresh client sent exactly the original reasoning text in reasoning_content on the next HTTP request, and the continuation completed. A trailing live-context message was included. Eight calls total: 2,759 input tokens including cache reads, 797 output tokens; no retries. This establishes field preservation and successful continuation, not whether each route rejects missing reasoning. No negative live calls were spent. The fix stores only a source-field marker beside the single readable text value; incompatible scopes still demote to ordinary text. Five regression cases failed before the fix. Sync/async mocked HTTP, resume, cross-provider, and privacy tests now pass: 1,040 UnifiedLLM/tracing tests. Severin’s other four findings remain open and are not included in these results.

src/nooa/unifiedllm/chat_parts.py
furgalep: Fixed in 4a43e16a (included in published head 0da40b5a). Capture now stores a source-field marker on the existing reasoning part, without duplicating the text. Compatible replay writes that text back to reasoning_content; edited or incompatible turns still demote it to ordinary text. Added direct DeepSeek scope support and mocked HTTP + SQLite regressions. Real Hub tool loops passed for DeepSeek V4 Pro, Kimi K3, GLM 5.3 and Qwen 3.5 397B, checking the actual outbound field after session reopen. An opt-in A/B case removes only that field after SDK serialization; success alone is not treated as proof of retention.

src/nooa/unifiedllm/response_parts.py
furgalep: Fixed in 0da40b5a. Unsupported Responses shapes now warn and preserve usable answer/refusal text while dropping native state for the entire turn. We do not pretend an incomplete provider turn can replay exactly. Malformed recognized fields still raise; unsupported-only output without a readable outcome also raises. Added sync/async tests for built-in output, refusal alone, and both together, including archive round-trip and absent native state. These reproduced the failure before implementation.

src/nooa/unifiedllm/unifiedllm.py
furgalep: Fixed in 0da40b5a in both sync and async paths. After separating think tags, ReasoningCompletionClient restores the already-validated parsed object on the replacement response. The replacement still drops native state; ordinary public edits still invalidate parsed results. Updated the regression to use an Answer Pydantic model and JSON answer, pass output_model, and assert that the validated object is preserved by identity. Both tests failed before the fix.

src/nooa/unifiedllm/unifiedllm.py
furgalep: Fixed in 0da40b5a. The calibration helper accepts instructions and prepends a synthetic system message only for estimation; both Responses call paths pass the effective instructions sent to the provider. Existing history dictionaries and strings are borrowed, not deep-copied. New sync/async regressions reproduce the long-system/short-user skew and assert both the complete estimator input and a 1.0 ratio for matching usage.

src/nooa/context_blocks/events.py
furgalep: Fixed in 0da40b5a at the benchmark JSON-export boundary. The JSON default handler recursively projects each nested Pydantic object through public live values, honoring repr/exclude/visibility rules instead of stringifying models or dumping nested archive data. ToolCallEvent.result remains a JSON object; nested LLMResponse values expose readable content/reasoning but no native state. The regression reproduced the string-index TypeError before the fix and now checks tool-result fields and nested privacy together. Full offline suite: 7,654 passed, 6 skipped, 3 expected failures.

alessiodevoto (approved):

furgalep:
All three new colleague findings were valid and are fixed in 46d27810: route-independent inline-signature extraction and raw-dictionary rejection; direct translation of public Responses text-block lists alongside calls; and retention of reasoning_content-only replies. Initial regressions: 7 failures before implementation. Final full offline run: 7,633 passed, 7 skipped, 302 deselected, 3 expected failures. Sync/async HTTP-body tests cover SQLite reopen and incompatible replay. Wren independently reproduced and accepted the fixes (206 touched-module tests). The stack is rebased; #325 now has 75 wire-contract cases including inline-only Gemini gateway replies. The description and linked docs explain the deliberate raw signed-dictionary compatibility change and the remaining LiteLLM alias/direct-provider limitations. No live inference was used for this increment.

PR #319, opened 2026-09-11
feat(llm): use one stable-prefix cache boundary policy

furgalep:
Published head: `b926359e`. Rebased on #318’s five Severin fixes; cache-policy changes are unchanged. The seven-model live gate passed on #324, including all three closed-provider caching/resume cases.

## Summary

Depends on #318; supersedes the closed #313.

The renderer places a `CacheBoundary()` before live context. Like `LLMResponse`, it belongs to the UnifiedLLM interface and travels through the message list unchanged:

```python
from nooa.unifiedllm import CacheBoundary

messages = [*history, CacheBoundary(), {"role": "user", "content": live_state}]
response = await client.acall(messages)
```

UnifiedLLM consumes the object after projecting assistant turns and selects the last eligible stable cache block. The formatter does not translate the boundary or understand caching: **the formatter and RenderedMessage have no changes relative to #318**.

## Why

Exact replay preserves a reusable prefix, but does not select where a provider writes its cache. Changing trailing context needs a breakpoint before it. Positional rules such as “all system messages” or “last tool result” can select changing content.

This PR removes those positional methods and their configuration. One boundary policy replaces them, without a second compatibility path.

1. **UnifiedLLM input type (`llm_types.py`, exported by `nooa.unifiedllm`):** `CacheBoundary` is a small frozen Pydantic model with a read-only public mapping, like `LLMResponse`. This keeps SDK utilities compatible without consumer-specific workarounds. It carries no provider fields. Ordinary readers can still use `message["role"]` and `message.get(...)`.
2. **Cached renderer:** place that object before live context through the existing `replay_message` path. The formatter passes it through like an assistant response. No adjacent-message flag, dictionary encoding, or cache-specific formatter branch remains. The standalone Anthropic exporter continues omitting framework metadata.
3. **UnifiedLLM clients and `cache_policy.py`:** preserve the object during provider projection, then consume it and mark the last eligible stable block. This matters because one assistant response can expand into several wire items. Only changed target containers are copied; unrelated messages and large string leaves are shared. Duplicate boundaries and raw dictionary markers raise actionable errors before transport. The fake client consumes boundaries too, keeping its recorded messages provider-shaped.
4. **NeMo Relay:** extend the existing response projection/reconciliation to boundaries. Only this JSON seam converts the object to public metadata; an unchanged round trip restores the original object. An edited marker is not silently reinterpreted as cache policy.
5. **Configuration and registry:** replace `cache_control_injection_points` with constructor-only `cache_breakpoint`. Removed settings name the replacement and `CacheBoundary()` in their migration error. Per-call/extra-body misuse fails before dispatch. Explicit mappings cannot silently follow a model switch.
6. **Tests and documentation:** replace positional scenarios with boundary placement, ownership, native-turn expansion, relay identity and changing-context contracts. Mocked HTTP tests verify the actual OpenAI/Anthropic request fields. The opt-in live suite closes/reopens SQLite and uses a fresh client between sends, comparing durable state and stable wire prefixes.

## Configuration

- `cache_breakpoint="auto"` (CompletionClient default): recognized Anthropic routes receive a native breakpoint; other routes retain provider-default caching.
- `"anthropic"`: explicit Anthropic Chat mapping for a declared route.
- `"openai"`: explicit OpenAI Responses mapping, opt-in on supporting routes.
- `None` (ResponsesClient default): no NOOA-generated provider markers; the boundary is still consumed. ResponsesClient accepts only `None` or `"openai"`.

Without a boundary, direct callers default to leading system/developer instructions. To cache completed history, put `CacheBoundary()` after it. Raw `nooa_cache_boundary` dictionaries are not a second supported input representation.

OpenAI explicit mode with no eligible stable block warns and creates no cache writes; it does not silently revert to implicit writes on dynamic content. No cache hit is guaranteed: edits, expiry, routing, request settings and provider thresholds still matter. Gemini remains on implicit caching; this PR does not manage cached-content resources, add effort controls, or change the TUI.

## Validation

Current revision: `286e1c9e4f6f1d3a9114b764e0a4cdb158adac41`.

SDK-compatibility follow-up: **2,410 passed, 6 skipped, 67 deselected** across UnifiedLLM, context blocks, runtime, tracing, Relay and nested history. The original safety-net test is restored unchanged. A new regression compares LiteLLM token counting and JSON/Mapping access against the equivalent public dictionary. Repository-wide lint and formatting pass. Wren independently accepted this exact revision after **2,470 affected-suite tests passed**. They verified SDK token counting, public dict/JSON views, relay identity, read-only assignment, and the restored safety-net test.

Previous revision `93b57d7d` full root + memory + benchmark offline suite (not rerun for this three-file compatibility follow-up): **7,767 passed, 21 skipped, 248 deselected** (252 seconds).

- 137 initial focused tests passed; after matching fake dispatch to real clients, all 52 regression/adjacent checks passed, including sync/async fake dispatch.
- Repository-wide Ruff lint and formatting pass.
- Real NeMo Relay Rust JSON round trip and request intercept preserve boundary and neighboring response identity.
- Wren accepted the final typed-boundary revision `286e1c9e`; no outstanding findings from that review.
- No live inference credits spent on this internal representation change. Existing mocked HTTP assertions remain unchanged.

Earlier live evidence, **not rerun at this revision**: NVIDIA Inference Hub on production revision `e527f9ce`; passing OpenAI/Sonnet tests at `fef83178`, Gemini at `808ddbfe`.

| Model | Resumed input tokens | Cached input tokens |
|---|---:|---:|
| GPT-5.6 Sol | 6,120 | 6,096 |
| Claude Sonnet 5 | 10,858 | 10,828 |
| Gemini 3.1 Pro Preview | 24,667 | 20,350 |

All three live checks close/reopen SQLite, construct a fresh client, change trailing context, and verify durable native state and stable HTTP prefixes. That round used 14 requests, 140,307 input / 5,856 output tokens, including two test-assumption corrections; retries were disabled.

See [the policy, migration, provider references and evidence walkthrough](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/286e1c9e4f6f1d3a9114b764e0a4cdb158adac41/docs/stable-prefix-caching.md).

## Latest review follow-up

At `5d915386`, rebased onto #318 `e43727e1`: the authoring skill now explains the `CacheBoundary` object, its `role=metadata` public view, and why UnifiedLLM consumes it before sending provider messages. This makes the framework control record distinguishable from conversation content without teaching consumers provider policy. The skill frontmatter is valid YAML. **2,590 affected tests passed**, and Wren accepted this head. No live tests were rerun for this documentation/rebase change.

## Second-pass review rebase

Head `d34b06c8` inherits #318's nested-formatting privacy and display corrections. The cache policy and renderer behavior are unchanged; only conflict context in documentation/imports needed resolution. The inherited authoring-skill registry explanation is corrected. **3,341 affected tests passed**, with 1 skipped and 69 deselected. No live-provider rerun for this rebase. Wren independently accepted `d34b06c8` after 3,234 scoped tests.

## Follow-up comment-fix rebase

Head `d26bbe6c` inherits #318's instance-hiding, scalar search, and leading-system text fixes. The cache feature commits are unchanged by the rebase (range-diff verified). **3,427 affected tests passed, 1 skipped, 69 deselected**. Wren confirmed the cache code is unchanged and accepted the updated core. Existing live evidence remains from the documented earlier run; no new inference was used for this rebase.

## Summary by CodeRabbit

- **New Features**
  - Added stable-prefix caching with explicit cache boundaries.
  - Added configurable cache breakpoint behavior for Anthropic and OpenAI Responses clients, with Gemini-compatible pass-through behavior.
  - Added public `CacheBoundary` support across messages, rendering, middleware, replay, and fake clients.
  - Added validation for unsupported providers, malformed boundaries, and conflicting cache settings.

- **Documentation**
  - Documented configuration, provider behavior, migration guidance, limitations, and validation.

- **Bug Fixes**
  - Improved cache-boundary preservation during persistence, provider switching, and resumed conversations.
  - Preserved readable reasoning across supported replay scenarios.


src/nooa/unifiedllm/cache_policy.py
```diff
@@ -0,0 +1,161 @@
...
+    if not responses:
+        raise ValueError("The OpenAI explicit cache mapping requires ResponsesClient")
+    marked = _mark_responses_cache_breakpoint(clean, boundary)
+    if not marked and instructions:
+        content, _ = _mark_responses_text(instructions)
+        clean.insert(0, {"role": "system", "content": content})
+        instructions = None
+    # No eligible stable text: explicit mode deliberately avoids caching a
+    # changing suffix. Never invent an empty text block just to host a marker.
+    return clean, instructions, True
```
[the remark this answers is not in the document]
furgalep: Verified the behavior against the current [OpenAI prompt-caching guide](https://developers.openai.com/api/docs/guides/prompt-caching). Explicit mode without a mark indeed performs no caching. We are intentionally keeping that behavior for an explicitly selected NOOA boundary: falling back to implicit mode could write beyond that boundary into changing context.

This is not the default ResponsesClient policy (that is None). The opt-in path now warns when no eligible stable block exists, and the client docstring and policy document explain the no-cache result. This also covers stable but unmarkable output-only history, not just wholly dynamic input. Existing regressions assert the warning and explicit-mode request. Wren reviewed and accepted this policy choice; I am not applying the suggested automatic fallback.

tests/unifiedllm/test_reasoning_state_replay.py
```diff
@@ -289,7 +289,7 @@ def test_responses_model_override_uses_effective_replay_scope() -> None:
             client.call(
                 _render_responses(first),
                 model="anthropic/claude-sonnet-4-5",
-                cache_control_injection_points=[],
+                cache_breakpoint=None,
```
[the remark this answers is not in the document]
furgalep: Verified: the misplaced keyword could reach the provider mock. Moved cache_breakpoint=None to the two ResponsesClient constructors. I also added a configuration error for cache_breakpoint in call/acall or extra_body, so a framework-only setting cannot leak into a provider request. Eight sync/async, Chat/Responses, top-level/extra_body cases assert rejection before any transport call.

The default-auto detail is outdated (ResponsesClient now defaults to None), but the argument-leak finding was still valid.

Published in `f52378ee`, rebased on the #318 review fixes. Wren accepted this revision. Full offline suite: 7,757 passed, 21 skipped.

furgalep:
Published the reviewed fixes at `f52378eea5d07fe7e1bd22bc6198258f2a865910` through the GitHub API. Commit IDs and full file-tree hashes match the locally tested revisions exactly; no code changes were made during publication. Wren has accepted this revision. The publication-pending notices are removed and the walkthrough links point to the published head.

@coderabbitai Please recheck the existing findings against this published revision; the earlier replies describe the fixes and regression coverage.

skills/nooa-agent-authoring/SKILL.md
```diff
@@ -56,6 +56,16 @@ llm = CompletionClient(model="m", base_url="https://.../v1", api_key="...")  # a
 
 Keys come from `.env` (library use) or `~/.config/nooa/secrets.yaml`.
 
+**Caching:** the cached renderer places a boundary before live context; UnifiedLLM
+maps it at dispatch. `cache_breakpoint="auto"` (CompletionClient default) marks recognized Anthropic
```
furgalep: This doesn't explain the role=Metadata that is used below

furgalep:
Implemented the standalone-block design in `cfc1ba6f`:

- CachedBlockFormatter inserts `CacheBoundary()` between stable history and live context.
- The provider formatter translates that block to `{"role":"metadata","nooa_cache_boundary":true}`.
- Removed `cache_boundary_before` from ordinary RenderedMessage objects. No following message is required to carry the boundary.
- UnifiedLLM policy and wire representation are unchanged; adjacent response objects keep their identity and native state.

Updated the code walkthrough and documentation. Full offline suite: 7,760 passed, 21 skipped; repository-wide Ruff checks pass. Wren accepted the delta with 2,462 independent scoped tests passing. No live inference spending.

furgalep:
Reply to [the metadata-role skill comment](https://github.com/NVIDIA-NeMo/labs-OO-Agents/pull/319#discussion_r3992985422):

Fixed in 5d915386. The skill now explains that the cached renderer inserts CacheBoundary() before live context, the formatter passes that UnifiedLLM object through unchanged, and its public JSON view has role=metadata. That role denotes a framework control record, not a model conversation message; UnifiedLLM consumes it before provider dispatch and applies the provider cache policy. Direct callers can insert the same object.

Also quoted the existing compatibility frontmatter value so its colon is valid YAML. NOOA skill loading/frontmatter checks and the 2,590-test affected suite pass; Wren accepted the update.

GitHub rejected an inline reply because this account already has a pending review. I have left that review untouched and posted the answer here instead. The earlier CodeRabbit cache-policy and per-call-override findings were rechecked and remain addressed.

@coderabbitai review

src/nooa/unifiedllm/cache_policy.py
```diff
@@ -0,0 +1,164 @@
...
+            {
+                "type": "input_text",
+                "text": content,
+                "prompt_cache_breakpoint": marker,
+            }
+        ], True
+    if isinstance(content, list):
+        for index in range(len(content) - 1, -1, -1):
+            block = content[index]
+            if isinstance(block, dict) and block.get("type") == "input_text":
```
sklinglernv: **P2 — Cache boundaries stop before stable multimodal blocks**

The OpenAI path searches only for `input_text`, while `_mark_anthropic` recognizes native `image` even though the actual Completion path still contains LiteLLM universal `image_url` blocks at this point. When the last stable message contains text followed by an image and then `CacheBoundary()`, both mappings put the marker on the preceding text block. The image therefore remains after the cache breakpoint even though it is before the requested boundary.

I reproduced this through the Anthropic HTTP serialization path: `cache_control` landed on the text block, followed by the converted image and then the live suffix. This causes stable, potentially expensive image/file inputs to be reprocessed on every request. OpenAI Responses explicitly supports `prompt_cache_breakpoint` on [`input_image`](https://developers.openai.com/api/reference/cli/__sdk_schema?declaration=%28resource%29+beta.responses+%3E+%28model%29+beta_response_input_image+%3E+%28schema%29&selected=%28resource%29+beta.responses), and [Anthropic supports caching images and documents](https://platform.claude.com/docs/en/build-with-claude/prompt-caching).

Please recognize the eligible projected multimodal types (`input_image`, `input_file`, and pre-LiteLLM `image_url`/document shapes) and place the marker on the latest eligible block.

Validation: 1,125 affected tests passed (5 deselected), and Ruff passed. The live provider suite was not run.

sklinglernv (approved):
Minor comment. otherwise LGTM

src/nooa/unifiedllm/cache_policy.py
furgalep: Fixed in efa521ea7133479da206f66215795f5418dd1e77. Both mappings now mark the latest eligible stable block, including images and files, rather than stopping at the preceding text. Only the changed containers are copied; preceding text and the original input remain untouched.

TDD: the new regressions reproduced the failure before the fix. Added 20 cases, including 8 checks of actual serialized HTTP bodies. Cache suites: 102 passed; broader affected suites: 2,239 passed, 7 skipped. Wren independently reviewed and accepted this exact commit.

Wren also confirmed image markers live through the Hub: OpenAI cached 3,032 of 3,035 input tokens on the second call; Anthropic read all 5,266 cache-written tokens. The existing SQLite-resume live test now includes a trailing stable image for both providers without increasing its call count; that expanded resume test has not yet been rerun live. The release gate inherits it when rebased.

PR #320, opened 2026-09-12
feat(llm): declare reasoning levels per configured route

furgalep:
## What and why

Let callers inspect and select reasoning levels without knowing each provider's request fields. Registry YAML maps each label to complete request settings. Selecting a label replaces those top-level settings before dispatch; there are no provider-name rules, nested merge rules or new dependencies.

This PR targets main, which includes #318 and #319. It does not change retained reasoning, replay, archives, cache policy or the TUI.

- **`unifiedllm/reasoning.py`** validates declarations and applies selections for both clients. Unknown support, unsupported selection and invalid labels have distinct errors. Reserved fields prevent a level from changing the route or history. Selection repeats declaration validation because nested configuration dictionaries can be edited after construction.
- **`unifiedllm.py`** exposes levels and their documented default, with persistent or per-call selection. No selection preserves raw settings; explicit `None` bypasses a persistent selection. The documented default is metadata, not an instruction to spend more tokens.
- **`registry.py` and `config/model_config.py`** carry declarations from YAML. Changing an alias's route or client type clears inherited levels, default and selection; callers must declare choices for the replacement route. The fake client validates selections too.
- **Settings precedence**: a selected level replaces constructor defaults, including values inside `extra_body`, without mutating the original configuration. Conflicting per-call raw settings fail. A supplied SDK client is rejected during managed selection because it can change the endpoint.
- **Examples and documentation** use placeholder endpoints and illustrative parameter shapes. They are not a supported-model catalog. Private routes, credentials guidance and deployment measurements belong in private repositories.
- **`skills/nooa-agent-authoring/SKILL.md`** teaches agents how to configure and select levels. A test executes its actual YAML/Python example through registry loading and mocked dispatch.
- **Tests** cover configuration ownership, validation, route changes, sync/async dispatch and outgoing HTTP bodies. Optional live probes resolve aliases supplied through `NOOA_REASONING_TEST_MODELS`, using normal registry credentials rather than hard-coded infrastructure.

## Validation

Review follow-up `5a40b67c` addresses the three route/precedence findings and removes internal deployment details from this PR's examples, documentation and live test.

- 11 regressions were observed failing before the fixes.
- Reasoning and HTTP tests: **75 passed**, including the registry-driven live-probe function exercised with mocked HTTP.
- Affected UnifiedLLM, context and runtime suites: **2,314 passed, 7 skipped, 67 deselected**.
- Ruff, formatting and whitespace checks pass. No paid inference was used for these changes.

These tests prove request construction, not that every provider honors every setting. The caller's route needs its own reviewed declaration and validation.

[Architecture and usage](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/feat/llm-reasoning-levels/docs/reasoning-levels.md).


src/nooa/unifiedllm/reasoning.py
```diff
@@ -0,0 +1,92 @@
...
+    levels: dict[str, dict[str, Any]] | None = None
+    default: str | None = None
+
+    @model_validator(mode="after")
+    def validate_declaration(self):
+        if self.default is not None and self.default not in (self.levels or {}):
+            raise ValueError("reasoning_default must name a declared reasoning level")
+        for level, settings in (self.levels or {}).items():
+            if not level.strip() or not settings:
+                raise ValueError("reasoning_levels must have non-empty names and request settings")
```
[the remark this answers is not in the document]
furgalep: Fixed in e6168697. ReasoningConfig now rejects routing, input, declaration, and extra_body keys in level patches at client construction, naming the reserved fields. This prevents a selected level from changing the route or smuggling UnifiedLLM controls; arbitrary provider request parameters remain allowed, so this is not a provider-field allowlist.

TDD: the 11 reserved-key cases failed before the fix and pass now; a separate test preserves support for new provider fields. All 13 example HTTP mappings pass. The full offline suite is green: 7,638 passed, 6 skipped, 301 deselected, 3 expected failures. Wren independently accepted the change.

src/nooa/unifiedllm/reasoning.py
```diff
@@ -0,0 +1,110 @@
...
+    """Map public level names to request settings for one configured route.
+
+    None means unknown support; an empty mapping means unsupported. The default
+    documents the route's default, not a request to send it on every call.
+    Declarations live in registry YAML, not a model-name table in this module.
+    """
+
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    levels: dict[str, dict[str, Any]] | None = None
```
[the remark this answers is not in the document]
furgalep: Fixed in bd7ea3e7. I reproduced four post-construction mutations (model, api_base, extra_body, reasoning_level) that bypassed the reserved-field check. settings() now calls the existing validate_declaration() before copying/applying a selected level. This reuses the same rules as construction instead of adding another validator or immutable wrapper.

TDD: four failures before the change; all 58 effort/HTTP-mapping tests pass afterward. Final affected suite on this head: 3,357 passed, 1 skipped, 69 deselected. No live inference used.

furgalep:
Updated the agent-facing nooa-agent-authoring skill in debe768a. Discovery text now includes model registry/reasoning configuration. The skill explains declarations, unknown vs unsupported, metadata defaults, persistent/per-call selection, whole-block replacement, conflicts, route changes and retention/caching distinctions. Its actual YAML/Python example is tested through NOOA skill loading, registry loading and mocked dispatch; the test failed before the example was added. All 77 selected reasoning/HTTP/skill tests pass, lint and formatting clean. The existing compatibility frontmatter is preserved: NOOA accepts it, although the generic Codex skill validator does not. No inference calls or production-code changes.

furgalep:
## Expressiveness audit: Pi and PydanticAI reasoning controls

**Conclusion: high confidence that #320 can express the major static parameter shapes needed to enable/configure reasoning. This is not a claim that every route already supports them through LiteLLM. No provider mappings or dependencies were added for this audit.**

Reviewed upstream source at Pi `ceea48f5` and PydanticAI `cb417b89`, against this PR's `debe768a`. The important comparison is the request fields their adapters produce, not their public option names or the particular budgets they assign to labels.

`reasoning_levels[label]` is an arbitrary nonempty parameter dictionary (apart from reserved framework/routing fields). Selection assigns its complete top-level values. It can set strings, booleans, numbers, nested objects, arrays, and several related fields together. Labels are not constrained to a shared provider vocabulary.

### Coverage by mechanism

The payloads below describe **shapes**, not ready-to-paste declarations for every NOOA route. Native SDK wrappers and PydanticAI-prefixed settings must be expressed using the configured NOOA transport's parameter names.

| Mechanism found upstream | Parameter shape | Fit in #320 |
|---|---|---|
| OpenAI Chat / compatible effort controls | `reasoning_effort: high` | Yes: scalar field. PydanticAI also uses this for supported Groq/xAI/Cerebras/Mistral paths. |
| OpenAI Responses / Codex | `reasoning: {effort: high, ...}` | Yes: complete nested object; additional mode/summary/context fields are expressible without new NOOA branches. |
| Older Anthropic budget-based thinking | `thinking: {type: enabled, budget_tokens: 4096}` | Yes; the same label can set an appropriate `max_tokens` alongside the budget. |
| Anthropic adaptive thinking | `thinking: {type: adaptive}` plus `output_config: {effort: high}` | Yes: multiple coordinated fields. Required beta headers/settings are also static data where supported by the transport. |
| Gemini level or token budget | thinking config with `thinkingLevel` or `thinkingBudget`, optionally `includeThoughts` | Yes as nested data; native Google SDK names are not automatically NOOA/LiteLLM argument names. The existing Hub example uses its supported `reasoning_effort` interface. |
| DeepSeek / Z.AI / GLM switch plus optional effort | `thinking: {type: enabled}`, optional `clear_thinking` and `reasoning_effort` | Yes. The readable-reasoning retention mechanism is separate from selecting these fields. |
| Qwen-compatible switches | `enable_thinking: true` or `chat_template_kwargs: {enable_thinking: true, ...}` | Yes. Pi's generic template variables can be resolved into literal values per declared label. |
| Baseten-style template arguments | `chat_template_args: {...}`, optionally effort/budget fields | Yes: arbitrary nested fields, not a maintained field allowlist. |
| OpenRouter / Together / Ant-Ling variants | nested `reasoning` containing effort, enabled, or budget fields, with route-appropriate combinations | Yes as data. Do not copy another library's `extra_body` wrapper verbatim. |
| String thinking / Mistral prompt mode | `thinking: high` or `prompt_mode: reasoning` | Yes. `prompt_mode` is a request field, not an instruction inserted into history. |
| Bedrock family-specific wrappers | `additionalModelRequestFields` containing thinking/output_config, reasoning_effort, or reasoning_config | Yes as a nested-object shape; the installed adapter must expose the corresponding kwargs. |
| Reasoning visibility options | `reasoning_format: parsed`, `includeThoughts`, display settings | Expressible, but visibility is not equivalent to enabling/disabling computation. Likewise a provider's lowest level does not necessarily turn reasoning off. |

Source anchors: [Pi Chat variants](https://github.com/badlogic/pi-mono/blob/ceea48f5d5d12fd7915dfefba2835ccd55f23bb9/packages/ai/src/api/openai-completions.ts#L863), [Pi Anthropic](https://github.com/badlogic/pi-mono/blob/ceea48f5d5d12fd7915dfefba2835ccd55f23bb9/packages/ai/src/api/anthropic-messages.ts#L1122), [Pi Google](https://github.com/badlogic/pi-mono/blob/ceea48f5d5d12fd7915dfefba2835ccd55f23bb9/packages/ai/src/api/google-generative-ai.ts#L390), [Pi Mistral](https://github.com/badlogic/pi-mono/blob/ceea48f5d5d12fd7915dfefba2835ccd55f23bb9/packages/ai/src/api/mistral-conversations.ts#L371); [PydanticAI OpenAI](https://github.com/pydantic/pydantic-ai/blob/cb417b896c2d1f49a6d9df91df8cd1d75f20d5c9/pydantic_ai_slim/pydantic_ai/models/openai.py#L2949), [Anthropic](https://github.com/pydantic/pydantic-ai/blob/cb417b896c2d1f49a6d9df91df8cd1d75f20d5c9/pydantic_ai_slim/pydantic_ai/models/anthropic.py#L1115), [Google](https://github.com/pydantic/pydantic-ai/blob/cb417b896c2d1f49a6d9df91df8cd1d75f20d5c9/pydantic_ai_slim/pydantic_ai/models/google.py#L918), [Bedrock](https://github.com/pydantic/pydantic-ai/blob/cb417b896c2d1f49a6d9df91df8cd1d75f20d5c9/pydantic_ai_slim/pydantic_ai/models/bedrock.py#L960), [Groq](https://github.com/pydantic/pydantic-ai/blob/cb417b896c2d1f49a6d9df91df8cd1d75f20d5c9/pydantic_ai_slim/pydantic_ai/models/groq.py#L352), [Z.AI](https://github.com/pydantic/pydantic-ai/blob/cb417b896c2d1f49a6d9df91df8cd1d75f20d5c9/pydantic_ai_slim/pydantic_ai/models/zai.py#L240), [OpenRouter](https://github.com/pydantic/pydantic-ai/blob/cb417b896c2d1f49a6d9df91df8cd1d75f20d5c9/pydantic_ai_slim/pydantic_ai/models/openrouter.py#L639).

### What this deliberately does not reproduce

1. **Transport translation and capability inference.** PydanticAI's `google_thinking_config`, for example, is its own interface, not a universal wire key. Some adapters put fields in `extra_body`; #320 reserves that wrapper, so a declaration must use the equivalent supported top-level transport kwargs. If the current adapter exposes no equivalent, that route has a transport gap, not automatic coverage. The existing Sonnet alias rejection/drop behavior demonstrates why inspecting the outgoing HTTP body remains necessary.
2. **Dynamic algorithms.** Pi adjusts thinking and answer budgets using model limits and remaining context ([budget helper](https://github.com/badlogic/pi-mono/blob/ceea48f5d5d12fd7915dfefba2835ccd55f23bb9/packages/ai/src/api/simple-options.ts#L54), [Anthropic call path](https://github.com/badlogic/pi-mono/blob/ceea48f5d5d12fd7915dfefba2835ccd55f23bb9/packages/ai/src/api/anthropic-messages.ts#L879)). #320 can declare the resulting fixed budget and output cap, but does not recompute them from each request. Provider restrictions involving sampling, forced tools, and output mode are not automatically discovered or repaired either.
3. **Message-based controls.** Pi inserts historical/current effort messages on some Anthropic routes ([implementation](https://github.com/badlogic/pi-mono/blob/ceea48f5d5d12fd7915dfefba2835ccd55f23bb9/packages/ai/src/api/anthropic-messages.ts#L1405)); OpenAI documents cache-preserving `configuration_update` input items ([official guidance](https://developers.openai.com/api/docs/guides/latest-model#whats-new)). Neither is a parameter-block update. Prompt-only switches and switching to a separate reasoning model are also outside #320: messages/input/model are deliberately reserved. Ordinary request-level reasoning can still be selected; cache-preserving mid-thread changes are a separate feature.
4. **Omission/deletion semantics.** Selection sets/replaces fields; it cannot delete an inherited field. `None` is a value, not a generic remove-key operation. A named empty/no-op patch is rejected. Always-on/default reasoning can use the unselected base configuration (`reasoning_level=None`), and explicit off/disabled/zero controls fit where the provider actually supports them. No universal 'off' semantics are promised.

### Verification

Ran an offline, in-memory check of `ReasoningConfig` + `apply_reasoning_level` with **18 representative static shapes**: scalar effort; nested Responses options; budget plus output cap; adaptive thinking plus headers; Google level/budget; GLM switch; boolean switch; both template-object forms; nested OpenRouter/Together options; string thinking; Mistral prompt mode; Groq format; Bedrock wrapper; false/zero; and nullable values. All 18 were preserved exactly, with unrelated defaults intact. Also confirmed rejection of the extra_body wrapper, message/model edits, and empty named patches.

These were mechanism-only checks, **not live calls or proof of provider acceptance**; no mappings were committed. The PR's existing mocked HTTP tests provide transport evidence for its three declared example routes. Recommendation: keep the mechanism unchanged and small; populate future declarations during model onboarding/route validation rather than copying Pi or PydanticAI's runtime decision trees into NOOA.

furgalep:
Rebased onto main 3083ed02 after #319 merged; published head 26658c2480333dd4cbb8f454bc946f94a37e4cca. Resolved overlapping UnifiedLLM setup by preserving cache validation and preparing reasoning settings before cache/provider dispatch in both sync and async clients. Preserved both cache and reasoning sections in the agent-authoring skill. No reasoning-level behavior changes intended.

Validation: 2,298 passed, 6 skipped in UnifiedLLM/context/runtime suites; final combined reasoning/cache tests 161 passed; lint and formatting clean. Wren accepted the earlier rebase and reviewed the skill update; the final post-#319 resolution is sent for recheck. Fresh GitHub checks will run on this head.

src/nooa/unifiedllm/registry.py
```diff
@@ -386,6 +386,9 @@ def get_llm_client(name: str, *, client_type: str | None = None, **overrides) ->
         "max_tokens",
         "reasoning",
         "reasoning_effort",
+        "reasoning_levels",
+        "reasoning_default",
+        "reasoning_level",
```
alessiodevoto, line 391: **[P2] Invalidate inherited reasoning declarations when the registry route is overridden**

These fields are copied from the alias even when the caller changes its endpoint, model, or client type. For example, after loading the example registry, `get_llm_client("gpt-5.6-sol", api_base="https://different-route.test/v1")` retains the Hub alias's reasoning levels. A subsequent `acall(..., reasoning_level="low")` sends the Hub settings to the replacement endpoint; I reproduced this with mocked HTTP interception.

The runtime route guard cannot catch this because the constructor has already made the replacement endpoint the client's baseline. This assumes constructor overrides preserve the original route's capabilities, contrary to the route-specific declaration contract. When routing or client type changes, please require an explicit replacement declaration or reset reasoning support to unknown.

src/nooa/unifiedllm/reasoning.py
```diff
@@ -0,0 +1,113 @@
...
+    if isinstance(extra, Mapping) and (set(extra) & (_DECLARATIONS | {"reasoning_level"})):
+        raise ValueError("Reasoning configuration cannot be passed through extra_body")
+    level = params.pop("reasoning_level", default_selection)
+    if level is None:
+        return params
+    patch = declaration.settings(level)
+    if (
+        any(
+            key in overrides and overrides[key] != defaults.get(key)
+            for key in ("api_base", "base_url", "custom_llm_provider")
```
alessiodevoto, line 100: **[P2] Account for SDK client overrides in the route-change guard**

Routing is not fully determined by these kwargs: LiteLLM can use the endpoint bound to a supplied SDK `client`. With the Gemini example client, `acall(..., reasoning_level="low", client=other_sdk_client)` passes this guard even when `other_sdk_client` is an `AsyncOpenAI` instance pointing at another endpoint. My mocked HTTP probe recorded `reasoning_effort="low"` being sent to `https://different-route.test/v1/chat/completions`.

This silently applies one route's declaration to another route despite the stated validation guarantee. Please reject unverified SDK client overrides during managed selection, or validate their effective route before dispatch.

src/nooa/unifiedllm/reasoning.py
```diff
@@ -0,0 +1,113 @@
...
+        )
+        or overrides.get("model", model) != model
+    ):
+        raise ValueError("Reasoning levels are route-specific; create a client for the new route")
+    if conflict := patch.keys() & (
+        overrides.keys() | (extra.keys() if isinstance(extra, Mapping) else set())
```
alessiodevoto: **[P2] Distinguish constructor extra_body defaults from per-call conflicts**

`extra` comes from the merged `params`, so this check also rejects inherited constructor defaults. For example, construct the GPT example with `extra_body={"reasoning": {"effort": "high"}}`, then call `acall(..., reasoning_level="low")` without any per-call raw settings: it raises `reasoning_level conflicts with explicit request field(s): ['reasoning']`. I reproduced this locally.

The documented precedence says a selected level replaces constructor defaults and conflicts with per-call raw controls. The same default works when supplied as a top-level `reasoning` constructor kwarg, making behavior depend on where the default was stored. Please distinguish inherited `extra_body` values from per-call values and remove superseded inherited keys before transport so they cannot override the selected settings later.

alessiodevoto (commented):
Reviewed this head and ran the UnifiedLLM suite: 951 passed, 5 deselected. The three inline findings were reproduced with additional mocked HTTP probes through the installed SDK; no live inference was used.

docs/reasoning-levels.md
```diff
@@ -0,0 +1,128 @@
...
+  Changing effort can invalidate a provider's cached prefix. This PR does not add
+  cache-preserving mid-turn steering, TUI controls, or selection persistence.
+
+## Tests
+
+Run `uv run pytest tests/unifiedllm/test_reasoning_levels.py tests/unifiedllm/test_reasoning_levels_wire.py`.
+The tests check configuration ownership, invalid selections, route changes,
+unchanged defaults and the serialized HTTP requests for the example routes.
+
+For paid route probes, supply `NVIDIA_INFERENCE_API_KEY` (or
```
alessiodevoto: Should we remove refs to internal nvidia endpoints here ?

src/nooa/unifiedllm/registry.py
furgalep: Fixed in 5a40b67c. Registry endpoint/model/provider/client-type or SDK-client overrides now clear inherited levels, default and persistent selection. Explicit replacement declarations are applied afterward; otherwise support is unknown. Identical route values and a credential-only change preserve the declarations. Six route-change regressions and a same-route control cover this; the original configuration is unchanged.

src/nooa/unifiedllm/reasoning.py
furgalep: Fixed in 5a40b67c. Managed selection rejects a non-null per-call SDK client before dispatch, rather than trying to inspect its private routing state. The client key is also reserved in level declarations. Unmanaged raw calls retain their existing behavior. Both client classes have regressions; registry-level SDK overrides also clear inherited declarations.

src/nooa/unifiedllm/reasoning.py
furgalep: Fixed in 5a40b67c. Conflict detection now uses only the per-call extra_body. Selected keys are removed from inherited extra_body before applying the level, so SDK merging cannot overwrite the selection later. Unrelated defaults and the caller's original dictionaries remain unchanged. Both client paths are covered, and a mocked HTTP test asserts the actual outgoing low-effort block.

docs/reasoning-levels.md
furgalep: Agreed; removed in 5a40b67c. The public example uses placeholder endpoints, generic credential configuration and illustrative request shapes. The optional live test resolves aliases from the caller's registry, with no internal hostname or key-name assumptions. Internal deployment results were removed from this guide and the PR description; Wren is preserving them in private documentation and coordinating the downstream cleanup.

alessiodevoto (approved):
LGTM! Thanks!

furgalep:
Comment audit: all human and CodeRabbit review threads are addressed and resolved at 5a40b67c. I rechecked the reasoning selection, outbound request and registry tests; no additional code change was needed in this pass.

On CodeRabbit’s remaining docstring-coverage warning: the public configuration and selection APIs document their behavior, including replacement rather than nested merging, route isolation and defaults as metadata. I am retaining concise, self-describing test names instead of adding repetitive docstrings solely to meet a percentage. This is an intentional documentation-style exception, not a claim that the automated coverage warning is green. Automated reviews are currently paused; this audit read both the review threads and ordinary PR comments.

PR #321, opened 2026-09-13
Reuse parent prompt cache for asynchronous token-budget summaries

furgalep:
## Summary

Make token-budget summarization a background fork of the parent's completed model request. It reuses the messages, client, tools, settings and cache key, then appends a summary instruction. This is the only token-budget summarization path: no mode switch, separate summarizer model or standalone fallback.

The experiment directory is removed. Permanent regression tests and [the architecture and usage guide](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/fa06a4fc2125da90f21e9c8e2282b6ce95757e02/docs/cache-sharing-summaries.md) remain. This PR targets main, which now includes #318 and #319.

## Why

The old summarizer rebuilt the conversation under a different prompt, losing the parent's cached prefix. The middleware already has the actual request. Forking that request preserves the prefix without another renderer or a copy of the Python agent, while the parent keeps working.

1. **`agents/summarization.py`**: `TokenBudgetSummarizer` subscribes to completed model requests and BeforeTurn. Current provider usage triggers one background fork. The fork copies dictionary/list containers but shares tools and immutable response objects, so asynchronous edits cannot change its input and bound tools do not clone their owning agent.
2. **Results and collapse**: the fork executes no tools. It accepts text or one `return_result` containing a nonempty string, read as data. Truncated, errored, empty, malformed and executable-tool replies leave history unchanged. Two consecutive failures disable the hook with a warning, preventing repeated paid failures. Completed summaries apply before a later turn only if the original event IDs still match; new responses and recent events stay active.
3. **`runtime/actor.py` and `runtime/middleware.py`**: expose the effective client and filtered-history status, and retain the effective cache key at the existing middleware boundary. The fork goes through the same middleware chain. A task-local guard prevents a summary from launching another summary.
4. **`config/summarizer_config.py`**: remove `reuse_parent_prefix`; obsolete settings are rejected instead of silently ignored. A separate token-budget summarizer client is rejected. The separate `MethodSummarizer` feature remains unchanged because it summarizes completed methods rather than completed model requests; it is not a fallback.
5. **Shutdown**: background components register async cleanup with `EventManager.on_close()`. `Agent.aclose()` awaits those callbacks before the owner closes shared resources. The coding agent no longer needs a summarizer-specific shutdown loop.
6. **Tests and documentation**: delete the experiment and tests for removed paths. Keep permanent tests for unchanged parent requests, actual OpenAI/Anthropic HTTP prefixes, shared tool ownership, isolated containers, one pending task, recursion prevention, failure handling, stale results and shutdown. Summary recovery hints explain how to search and read archived events, with tests executing those calls.

## Limits

- Filtered-history agents cannot use this summarizer: the fork cannot summarize events it did not see. They receive a warning; the runtime's context-overflow safety net still applies.
- Structured-output parents fork with text output. Removing the output schema can reduce cache reuse; tools and tool choice remain unchanged.
- The summary is one request behind and sees the parent's rendered history, including truncation. This does not change collapse semantics.
- Cache reuse is not guaranteed. Offline tests verify preserved request prefixes, not provider cache hits or summary quality. Deployment-specific measurements belong with private configuration. The installed-agent live test passed on both providers; route-specific measurements are recorded privately.

## Validation

Wren independently accepted `0bae11f4`: the human-requested generic shutdown hook, rejection of errored summaries, and a developer-friendly guide. New regressions failed before the implementation and pass afterward.

- Broad affected suites on `d0daa2fb`: **2,357 passed, 4 skipped, 67 deselected**.
- Current agents/runtime/coding-agent suite: **1,258 passed, 3 skipped, 62 deselected**; the focused fork/coding-agent suite has **46 passing tests**.
- New behavior was tested failing before implementation; Ruff, formatting and whitespace checks pass.
- The experiment and standalone fallback remain deleted. No new provider or UnifiedLLM mechanism is introduced.

## Live end-to-end validation

Wren reviewed and ran `fa06a4fc`: both OpenAI and Anthropic cases passed in six provider calls, with retries disabled and each request capped at 2,048 output tokens. The parent returned while the summary ran; the next turn applied exactly one summary, retained the three planted handoff identifiers, preserved raw and recent events, and omitted archived notes from the next request. The outgoing fork preserved the parent request prefix, and both providers reported cached input for the summary call. Detailed deployment-specific usage is kept privately.

The permanent opt-in test is `tests/integration/test_summarizer_live.py`; its mocked counterpart and fork suite pass all 43 tests locally. These three new commits add tests and documentation only, and carry DCO sign-offs. Release-gate wiring follows in #324.


furgalep:
Added deterministic archive-recovery instructions to Summary.doc (96a74d88): search including archived events with a bounded result count, read an original event by tag, and expand this summary or nested summaries. The LLM does not generate these instructions. Four new regression cases failed before the change; 88 targeted tests pass afterward, including rendered hints and collapse/resume coverage. Ruff and formatting pass. Sent the delta to Wren for re-review.

furgalep:
Updated in 3376aca9: removed the experiment, reuse_parent_prefix setting, separate-client option and standalone fallback. Token-budget summaries now only fork the parent request asynchronously. Failed summaries preserve history; two consecutive failures stop further attempts with a warning. MethodSummarizer remains a separate feature, not a fallback. Updated the description and architecture guide with the what/why and limits. Wren accepted the cleanup; 2,357 affected tests passed, and all 32 fork contract tests pass after the final validation fix.

packages/nooa-cli/src/nooa_cli/coding/agent.py
```diff
@@ -182,6 +182,8 @@ async def handle(self, notification: dict[str, list[Any]]) -> RespondResult:
     async def close(self) -> None:
         shell = self.shell
         try:
+            for summarizer in getattr(self, "_summarizers", []):
+                await summarizer.aclose()
```
furgalep: This looks like an abstraction leak. Why does the agent have to know about the summarizer? It's part of our design goals that the agent doesn't have to know about the summarizer.

furgalep: Maybe we're missing something in our middleware design so that middleware subscribers get a notification when the agent is shutting down.

docs/cache-sharing-summaries.md
sklinglernv, line 1: remove? or make developer friendly?

sklinglernv:
[Codex Take]

**P2 — Errored provider responses can become active summaries.**

`src/nooa/agents/summarization.py:744` rejects `"length"` and the unreachable `"max_tokens"`, but not the valid normalized value `"error"`. An errored response containing partial text is therefore accepted into `_pending_summary`; at the next turn boundary, `_apply_pending_summary()` collapses the active source events using that incomplete text. This leaves subsequent turns relying on a failed summary until the archived originals are explicitly recovered.

I reproduced this with an `LLMResponse(content="partial provider output", finish_reason="error")`: the pending value was accepted and active tags changed from `["1", "2", "3", "4"]` to `["1..3", "4"]`.

Suggested correction: reject `response.finish_reason in {"length", "error"}` before assigning `_pending_summary`, and add a regression case alongside the existing `"length"` coverage.

sklinglernv:
Did you do any end-to-end tests with the summarization? If yes, with what models ?

sklinglernv (approved):
A few small comments from my side, but otherwise looks great. Thanks a lot!

furgalep:
@sklinglernv Re your end-to-end question: yes—the earlier prototype ran an installed CodeAct agent with the summarizer against GPT-5.6 Sol and Claude Sonnet 5, not just a standalone summary prompt. It checked that the summary retained the decision, budget and owner, then applied it at the normal safe boundary; both providers reported cache reads. Those were bounded synthetic smoke tests, not a general summary-quality evaluation.

Those live runs predate the subsequent fork-only/shutdown cleanup. I have not presented them as a live rerun of today’s head. The current changes are covered by permanent offline tests for the real outgoing HTTP prefix, background execution, result validation, collapse timing and cancellation. Deployment-specific routes and measurements remain in private documentation.

packages/nooa-cli/src/nooa_cli/coding/agent.py
furgalep: Fixed in 0bae11f4. Shutdown now uses a generic async lifecycle hook: components register cleanup with EventManager.on_close(), Agent.aclose() waits for it, and CodingAgent.close() calls that before closing shared resources. The shutdown method no longer iterates or knows about summarizers. The callbacks run in reverse order, are awaited once, support unsubscribe, and log individual failures without skipping other cleanup. Both summarizers register themselves. Regression tests cover cancellation, ordering, unsubscribe, repeated close and a failing callback. Wren reviewed and accepted this change.

src/nooa/agents/summarization.py
```diff
@@ -573,6 +612,170 @@ class TokenBudgetSummarizer(SummarizationAgent):
...
+                raise ValueError("Summary middleware returned no response")
+            text = response.content
+            if response.tool_calls:
+                # Read CodeAct's final value as data, never as an invocation.
+                if len(response.tool_calls) != 1 or response.tool_calls[0].name != "return_result":
+                    raise ValueError("Summary fork requested executable tools")
+                text = json.loads(response.tool_calls[0].arguments).get("result")
+            if not isinstance(text, str) or not text.strip():
+                raise ValueError("Summary fork must return nonempty text")
+            if response.finish_reason in {"length", "max_tokens"}:
```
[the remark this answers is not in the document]
furgalep: Fixed in 0bae11f4: both length and error are rejected, and the unreachable max_tokens value is removed. The error case first failed against the old code; it now proves partial text is never collapsed, history stays unchanged, warnings are logged, and two failures stop further summary calls. The focused tests pass; the broader agents/runtime/coding-agent run has 1,258 passing tests. Wren accepted the fix.

docs/cache-sharing-summaries.md
furgalep: Reworked the guide in 0bae11f4 for developers: a plain-language introduction, installation/shutdown example, conversation behavior, failure recovery and limitations come before the source walkthrough. It now explains the generic agent.aclose() hook and distinguishes failed summaries from the main agent’s context-overflow recovery. The experiment remains deleted. Wren reviewed the revised guide.

furgalep:
@sklinglernv Your errored-summary reproducer is fixed in 0bae11f4. A response with finish_reason=error is rejected even when it contains nonempty text; length remains rejected too. The regression test first failed on the previous head and now verifies unchanged event tags and shutdown of repeated failed attempts. Wren reviewed and accepted the fix. The related CodeRabbit thread has the implementation/test details.

Audit disposition for the remaining CodeRabbit docstring-coverage warning: the new lifecycle API has behavior/ownership docstrings and the usage guide now explains failure recovery in plain language. I am not adding repetitive docstrings to self-describing assertion tests only to meet the blanket percentage; that style warning remains acknowledged, not claimed green.

PR #324, opened 2026-09-14
test: gate release drafts on provider replay and cache checks

furgalep:
## Release provider checks: public logic, private routes

`provider_checks()` runs seven live provider cases before the capability comparison and release-draft creation. They do not run on every pull request. The tests catch provider and SDK changes that offline mocks cannot detect; they are not answer-quality evaluations.

- **`scripts/make_release.py`** invokes the seven checks on both release paths, disables retries, uses a fresh report directory and requires the exact seven test names and modules to pass exactly once, with no skips. A substituted or duplicate case cannot satisfy the gate. The supplied alias wheel is loaded using `uv run --frozen --with <wheel>` so a locked project sync cannot remove it from the gate environment. The runner copies its environment and sets test opt-in flags; it does not know credential names.
- **The two live test modules and `tests/integration/_release_gate.py`** resolve `release-gate-<family>` registry aliases. The alias supplies the model route, API style, endpoint and credential configuration; HTTP capture hooks use that endpoint. Missing aliases skip, which the release runner rejects. Private deployment details no longer live in the public tests.
- **The scenarios** check native or readable reasoning after SQLite close/reopen, exact outgoing replay and cache reads with changing trailing context. Seven cases use at most 17 capped requests with retries disabled. Session databases contain provider state and stay in private artifacts.
- **`src/nooa/util/quickstart.py`** drops the internal-gateway branch. `NOOA_QUICKSTART_MODEL` selects a registry alias or model name through the existing interface.
- **Public documentation** explains the behavior and failure conditions. Alias definitions, credentials guidance, deployment budgets and recorded measurements move to private release-controller documentation. The companion private configuration change must land with this change.

## Validation

Wren implemented and tested the registry-based cleanup; Rowan reviewed it and rebased it onto main as `3ff5bcbf`. The import resolution preserves main's CacheBoundary and image-cache checks. `15cc01c5` also removes the older capability gate's hard-coded private credential prerequisite; credentials belong to the configured aliases and private controller, not the public runner.

- **133 runner and quickstart tests pass** offline; missing/partial/skipped/error, duplicate and substituted evidence prevents draft creation. Six CI tests first failed without the old private credential variable, then passed after its hard-coded prerequisite was removed.
- Wren accepted `ecd3e833`, including ephemeral alias-wheel loading and exact JUnit identity checks. Regression cases failed against the old implementation and now pass.
- The seven expected JUnit identities were checked against an actual pytest report with live opt-ins disabled (seven skips, not live evidence).
- Ruff and formatting checks pass.
- No live inference calls were made for this cleanup. A configured controller must run the live gate before a release draft; offline results do not substitute for it.

[Public guide](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/feat/release-provider-checks/docs/release-provider-validation.md).

## Summary by CodeRabbit

- **New Features**
  - Release preparation now runs seven provider replay and cache checks across supported model families.
  - Validation requires all expected checks to complete successfully, preventing draft creation when results are missing, duplicated, skipped, or unsuccessful.
  - Quickstart model selection supports `NOOA_QUICKSTART_MODEL` with an existing fallback when supported credentials are unavailable.

- **Documentation**
  - Updated provider-validation guidance with alias resolution, exact test-result requirements, and offline validation behavior.
  - Updated replay guidance with capped calls, SQLite resume checks, optional alias availability, and retained results.


scripts/make_release.py
```diff
@@ -328,6 +329,74 @@ def tool_versions(image_digest: str | None) -> dict[str, str]:
...
+            die("Provider validation requires all seven cases to pass; no skips")
+    except (ReleaseError, OSError, ET.ParseError) as exc:
+        if manifest:
+            manifest.update(provider_validation={**evidence, "outcome": "failed"})
+        die(f"Provider validation failed; inspect private evidence in {directory}: {exc}")
+    if manifest:
+        manifest.update(provider_validation={**evidence, "outcome": "passed"})
+    ok("all seven provider cases passed")
+
+
```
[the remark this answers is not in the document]
furgalep: Fixed in ecd3e833. The gate now requires the exact seven (classname, name) identities, exactly once, in addition to rejecting skips, failures and errors. Regression tests first demonstrated that duplicate/substituted cases passed the old gate; they now fail it. I also checked the identities against an actual pytest JUnit report with live opt-ins disabled. The status lookup uses only fixed skipped/failure/error tag names, not user-controlled XPath. 133 runner/quickstart tests pass; Wren accepted.

scripts/make_release.py
```diff
@@ -1667,6 +1737,8 @@ def ci_main(args: argparse.Namespace) -> int:
         distributions = _copy_distributions(artifact_dir)
         manifest.update(distributions=distributions)
 
+        provider_checks(artifact_dir, manifest)
```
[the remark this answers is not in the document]
furgalep: Fixed in ecd3e833. CI passes its resolved model-alias wheel to provider_checks(), which executes uv run --frozen --with <wheel> pytest. The alias package is therefore available in the temporary run environment even after an exact locked sync; the lockfile is unchanged. Local releases can supply --internal-wheel too. Tests cover CI/local forwarding, the exact command, and failure preventing draft creation. No live provider run was made for this change. Wren reproduced the original gap and accepted the fix.

furgalep:
Comment audit completed at ecd3e833: the exact-case report check and model-alias package loading fixes are published, tested and Wren-approved. Both CodeRabbit review threads have detailed replies. Validation: 133 offline runner/quickstart tests; no new paid provider calls. The actual release controller must still obtain seven live passes before creating a draft.

The docstring-coverage warning is acknowledged as a style exception: provider_checks documents the environment, evidence and failure contract, and the guide explains how to run the gate. I am keeping concise assertion-test names instead of duplicating them in boilerplate docstrings. The existing report-status lookup uses fixed tag names only, so the generic XPath warning does not describe a user-controlled XPath input.

furgalep:
The failing check was DCO, not the build/test suites. Fixed at db3f0c34: the imported contribution now has the matching author and Signed-off-by trailer, with its original co-author attribution retained. Wren authorized and verified the correction. Every rewritten commit has exactly the same source tree as before. DCO is now passing; the unchanged-code CI suites are rerunning.

docs/release-provider-validation.md
sklinglernv, line 1: Let's keep docs for humans, no?

sklinglernv:
[AI Take] **P2 — Release aliases can be shadowed by ambient configuration**

`scripts/make_release.py:355` preserves `NEMO_OO_LLM_CONFIG` via `os.environ.copy()`, while `tests/integration/_release_gate.py:30` loads the normal layered registry. User, project, and environment configuration has higher precedence than bundled wheel defaults. Consequently, an existing `release-gate-*` alias can replace the supplied wheel route, allowing all seven checks to pass against unintended models/endpoints and producing invalid release evidence.

Please isolate the gate to aliases originating from the supplied wheel, or reject/clear higher-priority registry layers, and add a collision test.

sklinglernv:
[AI Take]

**Correctness: `provider_checks()` runs even when `--skip-capability` is set**

`local_main()` in [`scripts/make_release.py:1871`](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/feat/release-provider-checks/scripts/make_release.py#L1871) calls `provider_checks()` unconditionally, before the `if args.skip_capability:` branch at line 1874.

Failure scenario: a developer runs `make_release.py vX.Y.Z --skip-capability` intending a cheap docs-only release — per the flag's own help text ("skip the capability diff (docs-only releases; the LLM eval is the slow, costly step)"). Because `provider_checks()` executes before the skip check, it still fires 17 real, billed provider API calls and requires the `release-gate-<family>` registry aliases (or `--internal-wheel`) to be configured locally, even though this workflow never previously needed that setup.

Suggested fix: gate the `provider_checks()` call the same way as the capability diff (move it inside/after the `--skip-capability` branch), or add its own skip flag.

sklinglernv (approved):
A few comments. Otherwise, looks good to me.

scripts/make_release.py
```diff
@@ -328,6 +329,105 @@ def tool_versions(image_digest: str | None) -> dict[str, str]:
     return versions
 
 
+def provider_checks(
```
sklinglernv, line 332: Did you test the changed script?

sklinglernv:
> check native or readable reasoning after SQLite close/reopen, 
What does SQL have to do with the providers?

scripts/make_release.py
```diff
@@ -328,6 +329,105 @@ def tool_versions(image_digest: str | None) -> dict[str, str]:
...
+    Seven cases make at most 17 capped calls without retries. A fresh report
+    directory and exact case identities prevent stale or skipped evidence from
+    satisfying the gate. Session databases and reports stay in private artifacts.
+    """
+    # Credentials are not this runner's concern: each ``release-gate-<family>``
+    # registry alias names its own credential variable, and the private
+    # controller provides it. A missing credential fails the cases, which the
+    # count below rejects before any draft is created.
+    env = os.environ.copy()
+    env.update(NOOA_RUN_OPEN_MODEL_REPLAY="1", NOOA_RUN_CACHE_RESUME_LIVE="1")
```
sklinglernv, line 356: Why are these environment variables and not arguments to the make release script?

src/nooa/util/quickstart.py
```diff
@@ -24,12 +24,15 @@
...
-#   * NVIDIA_INFERENCE_API_KEY -> NVIDIA internal inference gateway
-#                                 (inference-api.nvidia.com; NVIDIA employees)
+#   * NOOA_QUICKSTART_MODEL    -> any registry alias or litellm model name, with
+#                                 its own key; installed bundled-config packages
+#                                 supply extra aliases through the registry
 # To use a specific model, set MODEL to any litellm name and provide its key,
 # e.g. MODEL = "claude-haiku-4-5" with ANTHROPIC_API_KEY.
-_internal_key = os.getenv("NVIDIA_INFERENCE_API_KEY") or os.getenv("NVIDIA_INTERNAL_API_KEY")
-if os.getenv("NVIDIA_API_KEY"):
+if os.getenv("NOOA_QUICKSTART_MODEL"):
```
sklinglernv, line 32: Can we use the new `nooa connect` flow for this instead?

scripts/make_release.py
```diff
@@ -1773,6 +1868,7 @@ def local_main(args: argparse.Namespace) -> int:
     head_sha, prev_tag, _prev_sha, existing = preflight(args.tag, args.allow_dirty)
     fast_checks()
     build_and_smoke(args.tag, head_sha)
+    provider_checks(REPORT_PATH.parent, internal_wheel=args.internal_wheel)
```
sklinglernv, line 1871: **Code review finding 1:** local_main forwards --internal-wheel only to provider_checks, not to the capability_diff() call that runs right after, so the two release-gate phases validate against inconsistent wheel/route sets.

**Failure scenario:** Running `python scripts/make_release.py v1.2.3 --internal-wheel dist/nemo_oo_agents_nvidia-*.whl` locally: provider_checks() correctly resolves release-gate-<family> aliases from the supplied wheel, but the capability_diff() call further down in local_main (unchanged by this diff) never receives internal_wheel, so GATE_MODELS with internal aliases resolve only against whatever is ambiently installed in the synced environment -- silently a different, unpinned route/credential set than what provider_checks just validated, with no error or warning.

tests/integration/test_cache_resume_live.py
```diff
@@ -118,15 +115,15 @@ def _client(family):
             cache_breakpoint="openai",
```
sklinglernv, line 115: **Code review finding 2:** gate_client('openai', ...) never pins client_type='responses'; it relies entirely on the private release-gate-openai alias YAML to select ResponsesClient, unlike the old code that constructed ResponsesClient(...) directly.

**Failure scenario:** If the private bundled-config package's release-gate-openai alias omits or misconfigures client_type, get_llm_client() (src/nooa/unifiedllm/registry.py:419-420) defaults to CompletionClient. CompletionClient.__init__ (src/nooa/unifiedllm/unifiedllm.py:1713-1716) then raises ValueError("CompletionClient cache_breakpoint must be 'auto', 'anthropic', or None") because this call passes cache_breakpoint='openai' -- a confusing crash pointing at cache_breakpoint instead of the real cause (wrong client_type in an out-of-repo config the public gate can't see or assert against).

tests/integration/test_cache_resume_live.py
```diff
@@ -310,11 +311,14 @@ def remove_live_suffix(body):
         assert response.finish_reason in {"stop", "tool_calls"}
         assert bool(response.tool_calls) is (response.finish_reason == "tool_calls")
     assert resumed.usage is not None
+    record_property("model", model_name)
+    for phase, response in (("seed", seed), ("warm", warm), ("resumed", resumed)):
+        record_property(f"{phase}_usage", response.usage.model_dump_json())
```
sklinglernv, line 316: **Code review finding 3:** record_property(f'{phase}_usage', response.usage.model_dump_json()) is called unconditionally for seed/warm/resumed, but LLMResponse.usage is Optional and only resumed.usage is asserted non-None.

**Failure scenario:** If a live provider omits usage on the seed or warm call (already anticipated elsewhere in this same file via `response.usage.model_dump() if response.usage else None` in the final print block), this line raises AttributeError: 'NoneType' object has no attribute 'model_dump_json', crashing the test with a raw traceback after real API budget has already been spent, instead of a clear assertion failure.

tests/integration/test_open_model_tool_reasoning_live.py
```diff
@@ -137,13 +138,16 @@ async def capture(client, request, *args, **kwargs):
...
-    async with CompletionClient(**options) as client:
+    async with gate_client(family, **options) as client:
         result = await client.acall(history, tools=tools)
     assert len(sent) == 2, "probe must not retry"
     replay = next(m for m in sent[1]["messages"] if m.get("role") == "assistant")
     assert replay.get("reasoning_content") == raw, "native reasoning field changed on wire"
     assert result.finish_reason == "stop"
     assert result.content
+    record_property("model", model_name)
+    record_property("seed_usage", seed.usage.model_dump_json())
```
sklinglernv, line 149: **Code review finding 4:** record_property('seed_usage', seed.usage.model_dump_json()) / record_property('resumed_usage', result.usage.model_dump_json()) assume usage is never None, with no guard.

**Failure scenario:** Same pattern as test_cache_resume_live.py: LLMResponse.usage is `LLMUsage | None` (src/nooa/llm_types.py:264); if a provider omits usage on either call, this raises an unguarded AttributeError instead of a clean test failure, right after the release-budget-relevant API calls have already run.

tests/integration/test_cache_resume_live.py
```diff
@@ -310,11 +311,14 @@ def remove_live_suffix(body):
...
     print(
         json.dumps(
             {
                 "family": family,
-                "model": MODELS[family],
+                "model": model_name,
```
sklinglernv, line 321: **Code review finding 5:** The 'model' field printed via print(json.dumps(...)) now carries the real, privately-resolved route (client.model) instead of the old hardcoded public literal, and this print streams unfiltered into the CI job console because provider_checks() runs it with capture=False.

**Failure scenario:** Every real CI run of the openai/anthropic/gemini provider-validation case prints the resolved private model route -- the exact string the registry-alias indirection in tests/integration/_release_gate.py exists to keep out of the public repo -- straight to the pipeline console log. The sibling file (test_open_model_tool_reasoning_live.py) only records the model via record_property (private JUnit report), showing this is an inconsistency rather than a deliberate choice.

tests/integration/_release_gate.py
```diff
@@ -0,0 +1,56 @@
...
+        pytest.skip(
+            f"registry alias {gate_alias(family)!r} is not installed; the private "
+            "bundled-config package provides the release-gate routes"
+        )
+    return config
+
+
+def gate_host(family: str) -> str:
+    """Hostname of the family's endpoint, for request-capture hooks."""
+    host = urlparse(_entry(family)["api_base"]).hostname
```
sklinglernv, line 42: **Code review finding 6:** gate_host() does `_entry(family)['api_base']` as a required dict subscript, but the registry schema and get_llm_client() both treat api_base as optional (`config.get('api_base')`).

**Failure scenario:** If a release-gate-<family> alias legitimately omits api_base (e.g. relying on litellm's built-in provider routing), gate_host() raises an unhandled KeyError instead of the clean pytest.skip()/assert-with-message path _entry() otherwise provides for a missing alias, breaking the test with a confusing traceback instead of a clear skip.

scripts/make_release.py
```diff
@@ -328,6 +329,105 @@ def tool_versions(image_digest: str | None) -> dict[str, str]:
...
+    # registry alias names its own credential variable, and the private
+    # controller provides it. A missing credential fails the cases, which the
+    # count below rejects before any draft is created.
+    env = os.environ.copy()
+    env.update(NOOA_RUN_OPEN_MODEL_REPLAY="1", NOOA_RUN_CACHE_RESUME_LIVE="1")
+    env.pop(
+        "NOOA_TEST_OMITTED_REASONING", None
+    )  # Optional A/B calls are outside the release budget.
+    env.pop("OTLP_ENDPOINT", None)
+    artifact_dir.mkdir(parents=True, exist_ok=True)
```
sklinglernv, line 361: **Code review finding 7:** artifact_dir.mkdir(...) and tempfile.mkdtemp(...) run before provider_checks()'s try block, so an OSError there (e.g. disk full/permissions) propagates unhandled instead of being converted to the ReleaseError-safe-traceback contract the rest of the function promises.

**Failure scenario:** If the CI runner's artifact_dir is unwritable or the disk fills (plausible right after the preceding wheel-build step), provider_checks() raises a raw OSError that bypasses the `except (ReleaseError, OSError, ET.ParseError)` handler a few lines later, so no provider_validation entry is ever written to the manifest for this failure mode and the traceback isn't the 'safe to show' ReleaseError the module's docstring promises.

scripts/make_release.py
```diff
@@ -328,6 +329,105 @@ def tool_versions(image_digest: str | None) -> dict[str, str]:
...
+                str(directory / "sessions"),
+                "tests/integration/test_cache_resume_live.py::test_reasoning_and_prompt_cache_survive_sqlite_resume",
+                "tests/integration/test_open_model_tool_reasoning_live.py::test_open_model_tool_reasoning_after_sqlite_resume",
+            ],
+            env=env,
+            timeout=900,
+            capture=False,
+        )
+        cases = ET.parse(report).findall(".//testcase")
+        expected = {
```
sklinglernv, line 396: **Code review finding 8:** The 'seven expected cases' identity set is hardcoded in provider_checks() and duplicated independently in the FAMILIES tuples of both live test files and again in tests/test_make_release.py's test, with no shared source of truth.

**Failure scenario:** Renaming, adding, or removing a family in either test file's FAMILIES tuple without updating this literal set makes the release gate die() on an otherwise-legitimate pass ('Provider validation requires the seven expected cases...'), or in the rename/removal direction, could leave expected_cases: 7 in the manifest silently describing fewer real checks than actually run.

scripts/make_release.py
```diff
@@ -1592,8 +1693,6 @@ def ci_main(args: argparse.Namespace) -> int:
         die("an unmerged candidate can never create a draft")
     if args.checks_only and args.create_draft:
         die("--checks-only can never be combined with --create-draft")
-    if not os.getenv("NVIDIA_INTERNAL_API_KEY"):
-        die("NVIDIA_INTERNAL_API_KEY is required for the live capability gate")
     if not unmerged_candidate and not os.getenv("GH_TOKEN"):
```
sklinglernv, line 1696: **Code review finding 9:** The early `if not os.getenv('NVIDIA_INTERNAL_API_KEY'): die(...)` credential preflight was removed from ci_main with no equivalent early check, so a CI run with missing provider credentials now fails only after wheels/sdists are built and smoke-tested.

**Failure scenario:** A CI run where the alias wheel file is present (passes the new --internal-wheel existence/name checks) but the actual provider secrets are not injected into the environment (e.g. a broken secret-manager integration) now burns several minutes building 5 wheels/sdists before failing inside provider_checks(), instead of failing in under a second as before.

src/nooa/util/quickstart.py
```diff
@@ -38,12 +41,6 @@
 elif os.getenv("OPENAI_API_KEY"):
     MODEL = "gpt-5-mini"
     llm = get_llm_client(MODEL)
-elif _internal_key:
-    # NVIDIA-internal inference gateway (OpenAI-compatible).
-    MODEL = "openai/azure/openai/gpt-5-mini"
-    llm = get_llm_client(
-        MODEL, api_base="https://inference-api.nvidia.com/v1", api_key=_internal_key
-    )
 else:
```
sklinglernv, line 44: **Code review finding 10:** The automatic NVIDIA-internal-gateway fallback (auto-detecting NVIDIA_INFERENCE_API_KEY/NVIDIA_INTERNAL_API_KEY and configuring a working client with zero extra setup) was removed with no equivalent automatic path; it now requires explicitly setting NOOA_QUICKSTART_MODEL.

**Failure scenario:** An NVIDIA employee who has only the internal inference key set (previously the common case for the quickstart examples to 'just work') now falls through to the final else branch and gets MODEL='gpt-5-mini' with no OPENAI_API_KEY set -- a 'missing OPENAI_API_KEY' error for the wrong provider, confusing anyone who doesn't know they must now also set NOOA_QUICKSTART_MODEL.

scripts/make_release.py
```diff
@@ -1756,13 +1857,7 @@ def ci_main(args: argparse.Namespace) -> int:
...
 
 def local_main(args: argparse.Namespace) -> int:
-    if (
-        args.create_draft
-        or args.candidate_sha
-        or args.candidate_ref
-        or args.internal_wheel
-        or args.artifact_dir
-    ):
+    if args.create_draft or args.candidate_sha or args.candidate_ref or args.artifact_dir:
```
sklinglernv, line 1860: **Code review finding 11:** local_main() now legitimately accepts --internal-wheel (forwarded to provider_checks()) but, unlike ci_main(), never validates that the wheel file exists, has a .whl suffix, or is named like the nemo-oo-agents-nvidia package.

**Failure scenario:** A local operator passes a typo'd or wrong --internal-wheel path; instead of ci_main's clear '--internal-wheel must be the nemo-oo-agents-nvidia wheel' message, they get a generic 'Provider validation failed; inspect private evidence in {directory}: ...' after `uv run --with <bad-path>` fails deep inside the subprocess.

src/nooa/util/quickstart.py
```diff
@@ -38,12 +41,6 @@
 elif os.getenv("OPENAI_API_KEY"):
```
sklinglernv, line 41: **Code review finding 12:** After removing the NVIDIA-internal branch, the `elif os.getenv('OPENAI_API_KEY'):` branch is functionally dead: it and the trailing `else:` branch both now execute the exact same `MODEL = 'gpt-5-mini'; llm = get_llm_client(MODEL)`.

**Failure scenario:** Not a crash, but the explicit OPENAI_API_KEY check no longer changes behavior, silently defeating its own purpose; the `else` comment ('No key set -- default to OpenAI...') is now technically inaccurate since the same code also runs when OPENAI_API_KEY IS set.

tests/integration/_release_gate.py
```diff
@@ -0,0 +1,56 @@
...
+ALIAS_PREFIX = "release-gate-"
+
+
+def gate_alias(family: str) -> str:
+    return f"{ALIAS_PREFIX}{family}"
+
+
+def _entry(family: str) -> dict:
+    ensure_loaded()
+    config = MODELS.get(gate_alias(family))
```
sklinglernv, line 31: **Code review finding 13:** _entry() reads the shared MODELS dict directly (`MODELS.get(...)`) instead of going through the registry's own locked accessor, unlike get_llm_client()'s internal lookup which holds _registry_lock specifically to avoid observing a half-cleared dict during a concurrent reload_registry().

**Failure scenario:** Currently dormant (no concurrent reload_registry() call happens in this test session), but if a future fixture or test-runner mode reloads the registry mid-session, gate_host()'s unlocked read (at test start) could observe a different registry state than gate_client()'s later locked read moments later, causing the request-capture host hooks to silently never match real traffic.

scripts/make_release.py
```diff
@@ -328,6 +329,105 @@ def tool_versions(image_digest: str | None) -> dict[str, str]:
...
+    env.pop("OTLP_ENDPOINT", None)
+    artifact_dir.mkdir(parents=True, exist_ok=True)
+    directory = Path(tempfile.mkdtemp(prefix="provider-validation-", dir=artifact_dir))
+    report = directory / "results.xml"
+    evidence = {"outcome": "running", "report": str(report), "expected_cases": 7}
+    if manifest:
+        manifest.update(provider_validation=evidence)
+    step("Provider replay and cache checks (17 capped provider requests)")
+    try:
+        run(
```
sklinglernv, line 369: **Code review finding 14:** The 7 independent per-family provider cases (hitting different provider endpoints with no cross-family dependency) run serially inside one pytest process rather than in parallel, adding wall-clock time to every release's critical path before the multi-hour capability diff even starts.

**Failure scenario:** With HttpConfig(read_timeout=120) and reasoning-heavy models, cumulative serial latency across up to 17 calls can consume a meaningful fraction of the 900s timeout on every CI release and every local run, purely because the two node IDs are passed to a single non-parallel pytest invocation with no worker parallelism (pytest-xdist is not a project dependency).

sklinglernv (approved):
Some comments, but otherwise looks good to me.

PR #325, opened 2026-09-14
test(llm): verify HTTP prefixes across history round trips

furgalep:
## Why

A successful provider reply does not prove that retained reasoning or the cacheable prefix reached the provider unchanged. These tests compare requests produced by the current code at the HTTP boundary, rather than relying on saved recordings or HTTP success.

Test-only follow-up to merged #318 and #319, targeting main. Independent of #320, #321 and #324. No production changes, new dependencies, credentials or live inference calls.

## Code walkthrough — what and why

- **`tests/unifiedllm/test_history_wire_contract.py`** supplies synthetic Responses, Anthropic, OpenAI-compatible Gemini and readable-reasoning replies to the real SDK and UnifiedLLM paths. Mock transports capture outgoing JSON; unexpected HTTP requests fail. Both separate signature metadata and inline-only signatures are covered.
- The matrix compares complete requests before and after rendering, SQLite close/reopen, relay JSON reconciliation, or both, using fresh clients and sync/async dispatch. Only trailing dynamic text is excluded; stable tool results remain checked even when Anthropic combines them with that suffix.
- Explicit assertions check reasoning, signatures, tool arguments, result associations and cache markers. Edited turns and model/provider switches must omit incompatible state while preserving readable text.
- Negative controls corrupt reasoning, arguments, instructions and tool schemas and require assertion failures. A canned successful reply cannot hide a broken request.
- **`docs/stable-prefix-caching.md`** explains the offline tests and limits. Deployment-specific cache measurements and credentials instructions are removed from the public guide; those belong with private configuration.

This follows Pydantic AI's history-round-trip testing approach without adding a recording dependency. It proves request preservation, not cache hits, and does not replace the optional provider checks in #324.

## Validation

Head `add6b159` rebases the accepted tests onto main and removes deployment-specific documentation. Synthetic fixture labels now describe the protocol rather than an internal service.

- **177 passed** across the history matrix and cache-policy tests; **75** are the history matrix itself.
- Ruff, formatting and whitespace checks pass.
- No live provider calls. Only synthetic opaque values appear in fixtures.

Run: `uv run pytest tests/unifiedllm/test_history_wire_contract.py`

[Walkthrough and limits](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/test/provider-prefix-contracts/docs/stable-prefix-caching.md#offline-request-contracts).

## Summary by CodeRabbit

- **Documentation**
  - Expanded provider guidance for Anthropic multimodal requests and OpenAI Responses media breakpoints.
  - Clarified offline request-contract coverage, provider validation, and test commands.
  - Noted that preserving request history does not guarantee provider cache hits.

- **Tests**
  - Added coverage for conversation history across supported providers and synchronous/asynchronous workflows.
  - Verified reasoning content, tool calls, ordering, signatures, cache boundaries, and request integrity.
  - Confirmed that edits and provider changes retain readable reasoning while removing provider-specific secrets.


docs/stable-prefix-caching.md
```diff
@@ -104,51 +104,48 @@ the framework setting must never become a provider request field.
...
-Including these attempts, the round used 14 requests, 140,307 input tokens and
-5,856 output tokens, with retries disabled.
-
-Offline tests:
+### Offline request contracts
+
+`tests/unifiedllm/test_history_wire_contract.py` runs on ordinary PR CI without
+credentials. Synthetic provider replies pass through the real clients and SDKs;
+an HTTP mock captures the outgoing requests and unexpected httpx requests fail.
+It covers Responses encrypted reasoning, Anthropic signed thinking, the Gemini
```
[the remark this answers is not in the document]
furgalep: Updated in 5fbae800 to “Gemini tool signatures on OpenAI-compatible routes.” This makes the provider and transport distinction explicit without extra jargon. Documentation only; all 75 HTTP-history contract tests pass. Wren accepted.

furgalep:
Comment audit completed at 5fbae800: the Gemini wording is clarified, with a reply on the CodeRabbit thread. Wren accepted the documentation-only change; all 75 HTTP-history contract tests pass.

The blanket docstring-coverage warning is acknowledged as a style exception, not claimed fixed: the module and guide explain the shared wire-prefix contract, while the individual tests name the transformation or failure they assert. Adding a second copy of each test name as a docstring would not explain more of the design.

tests/unifiedllm/test_history_wire_contract.py
```diff
@@ -0,0 +1,376 @@
...
+    if family == "responses":
+        reasoning = [item for item in body["input"] if item.get("type") == "reasoning"]
+        assert len(reasoning) == 1
+        assert reasoning[0]["encrypted_content"] == SECRET
+        assert reasoning[0]["summary"] == [{"type": "summary_text", "text": THOUGHT}]
+        calls = [item for item in body["input"] if item.get("type") == "function_call"]
+        assert calls[0]["arguments"] == ARGUMENTS
+        results = [item for item in body["input"] if item.get("type") == "function_call_output"]
+        assert len(results) == 1 and results[0]["call_id"] == "call_1"
+        assert "prompt_cache_breakpoint" in json.dumps(results[0])
```
sklinglernv, line 259: Responses-family cache marker is only checked via substring (`'prompt_cache_breakpoint' in json.dumps(...)`), unlike the Anthropic branch below which asserts exact equality (`== {"type": "ephemeral"}`). A regression that keeps the key but corrupts its value (e.g. `null`, `{}`, wrong shape) would pass here but be caught on the Anthropic path. Worth tightening to an exact-value assertion for consistency.

tests/unifiedllm/test_history_wire_contract.py
```diff
@@ -0,0 +1,376 @@
...
+            for block in item["content"]
+            if block["type"] == "tool_result"
+        ]
+        assert len(results) == 1 and results[0]["tool_use_id"] == "call_1"
+        assert results[0]["content"][-1]["cache_control"] == {"type": "ephemeral"}
+    else:
+        assistant = next(item for item in body["messages"] if item["role"] == "assistant")
+        assert assistant["reasoning_content"] == THOUGHT
+        call = assistant["tool_calls"][0]
+        assert call["function"]["arguments"] == ARGUMENTS
```
sklinglernv, line 280: For the `gemini-compatible` family, this branch only checks the inline `call_id` suffix (`__thought__` + secret); it never asserts that the native `thought_signature` in `provider_specific_fields` survives the round trip. Given the PR's stated purpose is verifying "Gemini tool signatures on OpenAI-compatible routes," a future change that dropped `provider_specific_fields` preservation while keeping the id-embedded suffix would go undetected here.

docs/stable-prefix-caching.md
```diff
@@ -104,51 +104,48 @@ the framework setting must never become a provider request field.
...
-NOOA_RUN_CACHE_RESUME_LIVE=1 uv run --extra nemo-relay pytest tests/integration/test_cache_resume_live.py -m integration -k reasoning_and_prompt_cache -s
-```
-
-Set `NVIDIA_INFERENCE_API_KEY` separately. Three same-provider cases use nine
-requests with retries disabled, roughly 90k input tokens. Additional cases cover
-cross-provider stripping and portable reasoning from Nemotron, Qwen and DeepSeek.
-The suite prints usage and checks, not opaque contents.
+Live tests are opt-in and spend tokens. Configure their registry aliases for
+your deployment. Keep endpoint details, credentials guidance and measured
+results alongside that private configuration. Exact replay does not control
```
sklinglernv, line 150: This replaces concrete, non-secret info (live suite file path, exact pytest invocation, env var name) with an unnamed "private configuration" pointer that names no file/repo/location. The sibling doc `docs/assistant-turn-replay.md` still keeps the equivalent info public and inline for its own live suite. As written, an engineer who wants to run the opt-in live suite has no path to `tests/integration/test_cache_resume_live.py` (or whatever it's called) without asking a teammate — worth keeping the file path/invocation public even if credentials/endpoints move to a private config.

sklinglernv (approved):
Some findings from Claude. Otherwise, looks good to me.

PR #331, opened 2026-09-14
Backport current benchmark agents and runner lifecycle from dev/tui

furgalep:
## What does this PR do?

Backports the current benchmark agents from `dev/tui`: single `python_cell` execution, automatic summarization, method writing, bounded delegation, and Todo updates merged after successful worker cleanup. `BenchAgent` and the delegation-oriented `RLMBenchAgent` share the same capabilities; both await workers and return `TaskResult` with solution, observed evidence, and **How to Verify**.

Workers share the model client, summarization configuration, and workspace path, with separate shells and histories. Supplied context is an ordinary worker-method argument: NOOA renders it with its existing parameter formatting and makes it available in the Python namespace. **No custom delegation renderer or redaction policy.**

Default delegation depth is four; strategy settings are ten retries and a 1,800-second cell timeout. Awaited delegation shares the parent cell deadline. Timeouts cancel the worker without merging partial Todo state and do not consume the strategy retry counter; the enclosing harness budget limits repeated attempts. These semantics are now documented, not changed.

## Review follow-up

- Promote `CodeActV2` (`nooa.CodeActV2` or `nooa.strategies.CodeActV2`). Remove `CodeActLiteStrategy`, its exports and obsolete formatter/tests; the evaluation CLI offers `codeact_v2`. Breaking removal documented in CHANGELOG. `CodeActStrategy` remains the default, but its model-visible fixes are explicitly documented.
- Keep `CurrentCall` mutable for its live execution namespace, while preserving its correlation ID. The Task event's display tag now has its own field. Event filtering by the current call works for both strategies.
- Preserve validated inline-return values on real execution outputs without fabricating provider tool-call history. Retain original task arguments after compaction; restore concise context usage and compaction guidance for both benchmark variants.
- Resolve output reserves from UnifiedLLM's actual request settings (client defaults, selected reasoning level, per-call cap). Context percentages, automatic summary triggers and overflow recovery use the same usable input window. Unknown caps are labelled planning reserves; explicit summarizer thresholds stay fixed. Context management rejects a configured cap at or above the known window instead of installing a one-token summary budget; valid large caps retain their actual reserve. Normalize Responses reply caps to `max_output_tokens` and replace inherited cap aliases. Tests inspect actual serialized requests for all three interfaces, sync and async.
- Make Todo updates/restores atomic. Migrate legacy `notes` and non-done statuses through the **real stored-session deserializer**, not only `TodoManager.from_dict`. Document downgrade data loss.
[...]

PR #332, opened 2026-09-14
feat: configure models with NOOA Connect library and CLI

furgalep:
## Summary

Add `nooa connect`: an interactive model-configuration wizard, independent JSON stages, and the reusable `nooa.unifiedllm.connect` library. All prompts and terminal dependencies remain in `nooa-cli`; core gains no dependencies. UnifiedLLM is not made separately installable.

Current main (including #331) is merged and conflicts are resolved. Frozen head: `8893f15e53baa50859fd7f83e845c09063443424`. The dedicated `nooa-model-configuration` skill owns model setup and diagnostics; agent authoring only links to it.

### Configuration and checks

- New entries default to `transport: direct`. Earlier runtimes ignore this sidecar and use LiteLLM; #337 will honor it. Edits preserve explicit transport choices. Completed checks record the actual runtime transport separately.
- Configured routing, tool, reasoning and conversation checks send the saved reply cap, including level overrides. Initial interface discovery alone uses a small cap. Insufficient approved budget skips checks, never lowers the cap. Truncation is inconclusive; no automatic length retries.
- Endpoint limits precede catalogue metadata. Unknown limits and untested saved caps stay explicit. Caps at or above a known context window are refused; automatic defaults leave input room.
- Reasoning acceptance, on-wire settings, observed reasoning, puzzle correctness and replay remain separate evidence. Provider cache misses do not alone break an otherwise working connection.
- Responses defaults to stateless encrypted-reasoning requests. Empty `include` is a persisted opt-out. No model tools are executed.

### Review fixes

- Recursive credential-field rejection is shared by library/wizard/stage save. Reports scrub active credential values throughout, including keys/model names.
- Registry read/modify/replace is serialized with a retained sidecar lock. Symlinks are followed, existing permissions/newlines retained, and staging files cleaned on I/O errors. Null `models` and `env` mappings work; layered tombstones are respected.
- Anchored YAML is refused with actionable guidance instead of a traceback or destructive flattening. Other alias values and neighboring comments remain unchanged.
- Level-file editing, discovery `/v1` fallback reuse, missing encrypted-reasoning provenance and 422 reporting are fixed. Unobserved request capture is distinguished from settings proven absent; malformed HTTP-200 replies are not described as provider rejection.
- Wizard steps are separated; both frontends share `verdict(entry)` and a documented `ProbeRecord` shape. CLI prompt-toolkit minimum is corrected to 3.0.52, with no core dependency change.

## Review guide

1. `skills/nooa-model-configuration/SKILL.md` and `docs/model-connect.md`.
2. `src/nooa/unifiedllm/connect/__init__.py`: configuration, plans, evidence, persistence.
3. `_session.py` and `_records.py`: bounded conversation checks and shared evidence/verdict policy.
4. `packages/nooa-cli/src/nooa_cli/commands/_connect_wizard.py` and `_connect_stages.py`: terminal steps and JSON contract.
5. `test_connect_final_review.py`, both `test_connect_severin_review.py` files, and wizard transcript tests.

## Validation

Exact offline selection:

```sh
LITELLM_LOCAL_MODEL_COST_MAP=True \
PYTHONPATH=src:packages/nooa-cli/src:. \
uv run --no-sync --with prompt-toolkit --with msgpack pytest \
  tests/unifiedllm packages/nooa-cli/tests/test_connect* \
  tests/test_secrets.py tests/test_model_configuration_doc.py -q
```

**1,457 passed, 5 deselected** on the frozen head. Ruff lint/format, `git diff --check`, and dedicated skill validation passed. These are offline HTTP-contract tests, not new live-provider evidence. Fresh reviewer approval and CI are still required; previous approvals do not automatically cover this larger revision.


src/nooa/connect.py
```diff
@@ -0,0 +1,776 @@
...
+        replacement = yaml.safe_dump(models, default_flow_style=True, sort_keys=False).strip()
+        text = (
+            source[: models_node.start_mark.index]
+            + replacement
+            + source[models_node.end_mark.index :]
+        )
+    elif alias in data["models"]:
+        key, value = next((key, value) for key, value in models_node.value if key.value == alias)
+        lines = dumped.splitlines()
+        replacement = "\n".join([lines[0], *("  " + line for line in lines[1:])])
```
alessiodevoto, line 774: [P2] Preserve the existing YAML indentation when saving

The replacement hardcodes two spaces, as does insertion via `indented`. For a valid registry whose aliases are indented four spaces beneath `models`, replacing an alias raises `ValueError`, and adding an alias raises a YAML `ParserError`. I reproduced both cases. This prevents the wizard from saving at its final step, potentially after paid checks. Derive the alias and field indentation from the existing mapping instead of assuming two spaces.

src/nooa/connect.py
```diff
@@ -0,0 +1,776 @@
...
+        suffix = source[value.end_mark.index :]
+        if suffix and not suffix.startswith("\n"):
+            replacement += "\n" + " " * value.end_mark.column
+        text = source[: key.start_mark.index] + replacement + suffix
```
alessiodevoto, line 778: [P2] Preserve comments preceding the next alias

PyYAML’s `value.end_mark` can extend past comments preceding the next alias, so this splice silently deletes comments belonging to that neighbor. For example, replacing `local` in the following registry removes the description above `other`:

```yaml
models:
  local:
    model_name: openai/old
  # Important description for other
  other:
    model_name: openai/other
```

I reproduced the loss. The parsed-data equality check cannot catch it because YAML comments are discarded during parsing. Preserve trailing comments when calculating the replacement boundary.

packages/nooa-cli/src/nooa_cli/commands/connect.py
```diff
@@ -0,0 +1,483 @@
...
+        # Validate before using an endpoint or collecting a credential.
+        connect.plan(
+            alias or "candidate", model or "candidate", discovery_style, endpoint, api_key_env
+        )
+        needs_key = (
+            not yes and approval != "none" and api_key_env and not os.environ.get(api_key_env)
+        )
+        api_key = (
+            prompt("API key (used only for this setup)", hide_input=True)
+            if prompt_key or needs_key
```
alessiodevoto, line 191: [P2] Collect discovery credentials even with --no-probe

With `NVIDIA_API_KEY` unset, `nooa connect --provider nvidia --no-probe --no-catalogue` skips the temporary-key prompt because approval is `none`, then sends an unauthenticated GET to `/v1/models` and exits on 401. I reproduced this with mocked HTTP. Disabling generation checks should still allow authenticated model discovery; request credentials when MODEL is absent and discovery needs them, even when paid probes are disabled.

alessiodevoto (commented):
Reviewed this head with 128 focused tests passing. The three findings below were reproduced locally; HTTP was mocked and no paid API calls were made.

furgalep:
Added the cache/reasoning conversation checks at f6f81f08. Connect now asks once before generation calls, with a fixed shared estimated-token budget (65,536 by default). Full setup includes three bounded seed/replay/repeat calls through the candidate UnifiedLLM client and default cached renderer, without forcing cache or replay settings. It reports cache-read fraction and marker presence; a small tool-schema-only hit is not confirmation. Reasoning retention requires the seed reasoning state to reach the follow-up reasoning fields unchanged, with configured controls retained. Ordinary text fallback does not count. No tools are executed and native state/raw captured bodies are not persisted in provenance. --no-probe remains available; --yes approves the chosen checks as well as saving. Validation: 251 offline Connect, CLI/prompt and reasoning-level tests passed; lint clean. No paid live calls made for this increment. Wren has been asked to review the frozen delta; review is pending.

furgalep:
Encrypted reasoning default added at 85e0215d. Responses plans now set store:false and include:[reasoning.encrypted_content], with source connect in provenance; the wizard explains the choice without another question. An explicit 400/422 rejection of the option persists include:[] and reports it. That empty list now suppresses the runtime automatic include and omits the field on the wire, tested for both native and gateway URLs. A single no-include routing check is allowed only within the original approved budget; session rejection stops rather than rerunning. Authentication/rate/server errors and rejected model or input-history fields do not disable the option. No raw error bodies are recorded. Validation: 1171 broad offline tests passed, plus the expanded 13-case feature suite; lint and format clean. Wren has the frozen commit for delta review; the full PR review remains on hold. No paid calls made. Existing caveat: this branch still needs #341 incorporated to obtain the new Responses cache default; the checks report the current runtime rather than forcing that default.

furgalep:
Connection recovery added at 54629fa0. If no interface works, interactive setup offers Change key / Edit server and model / Retry / Cancel instead of a terminal error. Failed attempts remain charged to the initial budget; budget exhaustion or cancellation saves nothing. Existing model listing is described as listing, not credential validation. CLI shows safe authentication/routing/rate/timeout explanations and temporarily suppresses legacy help banners, restoring the flag after progress iteration. Responses explanation now appears only for the chosen Responses interface. Validation: 209 offline Connect/library/CLI/prompt tests passed, lint and format clean. Wren has the frozen delta for review; full PR review remains on hold. No paid calls made. Output-token preset selector is a separate design proposal, not part of this patch.

furgalep:
Added independent agent-facing stages at 16cbc0359f56f37b14888606600bc73a226852e2. `uv run nooa connect --stage ...` now exposes discovery, catalogue, interfaces, planning, routing, tools, reasoning, conversation checks, all checks, and explicit saving. Stage mode is non-interactive and emits JSON with existing library results/provenance, exit status, and a sanitized diagnostic prompt on failure or missing evidence. The wizard uses the same diagnostic helper; neither frontend launches an agent automatically.

The public `nooa.connect.check_stage(plan, stage)` runs selected checks afresh through the existing UnifiedLLM path without mutating the plan. `diagnostic_prompt(...)` is available to NOOA agents and TUI callers. Existing wizard approval behavior is unchanged; no cost-policy changes or live calls in this delta.

Validation: 231 Connect library/CLI tests passed; Ruff check/format clean; actual `uv run nooa connect --stage plan` smoke test returned JSON with no network calls. Wren has the frozen commit for delta review; the full PR review remains on hold.

Usage: [Independent stages and library API](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/16cbc0359f56f37b14888606600bc73a226852e2/docs/model-connect.md#independent-stages-for-agents).

furgalep:
Follow-up published at 60eeb5cbb6a04bc94ab2cccb2836eca3340906fb: Wren reviewed the stage delta and requested preserving actionable local errors. File failures now keep the path and reason; YAML errors name the file, line and column without echoing its contents. Validation errors remain useful, active credentials are redacted, and provider error bodies stay hidden. Both CLI modes include the safe detail in their diagnostic handoff. The documentation distinguishes the full library result from the request-free checks view.

234 Connect library/CLI tests passed, lint/format clean, no live calls. Wren follow-up verification requested; the full PR review remains on hold.

furgalep:
Published UI-only delta 6b9ac255223435761f6bac108fb9561e067c85f1: compact check rows with current-step feedback, elapsed times and passed/attention/skipped counts; numbered conversation steps plus separate cache/reasoning-retention results. A real terminal replaces the active line, while piped output remains plain readable lines. Full final YAML is now an explicit --show-config preview before saving; agent --stage JSON is unchanged.

Model details now call the catalogue number a reported reply ceiling, explain that it is metadata rather than the request reply budget, and flag ceilings near the context window. Boolean thinking/default metadata is shown even without a named effort list. No reasoning mappings or request caps changed. A mocked HTTP regression confirms the reported ceiling is not silently sent as the runtime reply cap.

Validation: 240 Connect library/CLI tests passed, Ruff and whitespace clean; no paid model calls. Wren has the frozen SHA for delta review. Explicit runtime reply-budget selection is a separate follow-up, not included here.

src/nooa/unifiedllm/replay_state.py
```diff
@@ -332,6 +332,9 @@ def prepare_chat_messages(
 def add_encrypted_reasoning_include(api_params: dict[str, Any], scope: str | None) -> None:
     """Request OpenAI encrypted reasoning only on endpoints known to support it."""
     configured = api_params.get("include")
+    if configured == []:
+        api_params.pop("include", None)
+        return  # Explicit opt-out, including on native OpenAI/Azure endpoints.
```
furgalep, line 337: Can you explain this?

furgalep: This makes `include: []` a persisted opt-out. Without the early return, the native OpenAI/Azure default below would add `reasoning.encrypted_content` back even after Connect observed a field rejection or the user disabled it. We remove the empty field from the outgoing request and stop automatic insertion; the saved entry retains `include: []`. Omitted `include` still permits the existing native-endpoint default, and a nonempty explicit include list still passes through. Tests cover the empty-list opt-out on native and compatible endpoints; Connect itself writes the encrypted-content include for stateless Responses unless explicitly opted out.

src/nooa/connect.py
furgalep: Fixed in fcefd249: derive the alias indentation from the existing YAML mapping for both insertion and replacement. Regression coverage exercises 2-, 4-, and 6-space indentation and verifies that every other parsed value is unchanged.

src/nooa/connect.py
furgalep: Fixed in fcefd249: replacement and insertion boundaries now use the last actual YAML value rather than a mapping end mark that consumes neighboring comments. Tests assert that the next alias's description and the following top-level section's comment stay intact.

packages/nooa-cli/src/nooa_cli/commands/connect.py
furgalep: Fixed in fcefd249: interactive discovery collects an unset credential even with --no-probe. A mocked HTTP regression asserts authenticated GET-only discovery, masked input, cancellation without a save, and no key in output.

src/nooa/connect.py
```diff
@@ -0,0 +1,809 @@
...
+            if isinstance(value, int) and not isinstance(value, bool) and value > 0:
+                entry[field] = value
+    reasoning = (catalogue or {}).get("reasoning") or {}
+    if reasoning_levels is None and reasoning.get("supported_efforts"):
+        template = "adaptive" if api_style == "anthropic" else "effort"
+        reasoning_levels = {
+            label: reasoning_settings(template, api_style, label)
+            for label in reasoning["supported_efforts"]
+        }
+        provenance["reasoning_levels"] = {"source": "catalogue", "template": template}
```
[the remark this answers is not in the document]
furgalep: Fixed in fcefd249: catalogue() validates string ids, normalizes provider/reasoning mappings and supported-effort lists, and removes non-string defaults before frontend use. Regression tests cover malformed optional values and invalid ids.

src/nooa/unifiedllm/connect/__init__.py
```diff
@@ -0,0 +1,831 @@
...
+                "outcome": "not_probed",
+                "reason": "not approved or earlier check failed",
+            }
+            yield ProbeUpdate("session", deepcopy(checks["session"]))
+        else:
+            async with aclosing(
+                session_steps(
+                    proposal.alias,
+                    proposal.entry,
+                    api_key=key,
```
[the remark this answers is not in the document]
furgalep: Fixed and rechecked in 8a87ffdc: approved runs resolve the configured key before reusing earlier probe results. A registry-to-HTTP regression verifies that reused routing/tools checks still produce three authenticated session calls, and a missing variable raises before HTTP.

src/nooa/unifiedllm/connect/__init__.py
```diff
@@ -0,0 +1,913 @@
...
+    if isinstance(param, str) and not (
+        param == "include" or param.startswith("include[") or param == "reasoning.encrypted_content"
+    ):
+        return False  # Rejected history/model fields are not include-option rejection.
+    return bool(
+        re.search(r"\b(include|encrypted_content)\b", message)
+        and re.search(
+            r"unsupported|unknown|unrecognized|not supported|not allowed|invalid|reject", message
+        )
+    )
```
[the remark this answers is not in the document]
furgalep: Fixed in 8a87ffdc: a validated structured include parameter on 400/422 is sufficient even with a generic message. Both routing retry and session opt-out tests now use generic error messages; unrelated-field/status negative controls remain.

docs/model-configuration.md
```diff
@@ -0,0 +1,96 @@
+# Configure a model without paid checks
+
+You can write a model configuration yourself, or ask an agent using the
+`nooa-agent-authoring` skill to do it. Neither requires a model API call.
```
sklinglernv: nit: that seems like a weird skill name for this?

docs/model-configuration.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Write the configuration
+
+Add an entry to `llm_config.yaml`, preserving any existing models. Replace the
+example endpoint and model ID with yours:
+
+```yaml
+models:
+  my-model:
+    model_name: openai/your-model
```
sklinglernv, line 17: question: where are we storing all of the intricacies of the models, like what reasoning is stored or any other stuff? Shouldn't this also be in this config?

docs/model-configuration.md
```diff
@@ -0,0 +1,96 @@
...
+    client_type: completion
+    api_base: https://gateway.example/v1
+    api_key_env: MY_MODEL_KEY
+    max_tokens: 8192
+```
+
+`my-model` is the local name your agents use. `your-model` is the exact ID the
```
sklinglernv, line 24: nit: this reads a lot like a schedule instead of human documentation.

docs/model-connect.md
```diff
@@ -0,0 +1,519 @@
...
+When a context bound is unknown, both library provenance and stage JSON warn.
+
+Save discovery JSON and reuse it with `--discovery-file discovery.json` on the
+wizard or a plan/check stage. The file must name the same normalized endpoint;
+limits are selected by exact model ID. This keeps `--stage plan` offline:
+
+```sh
+uv run nooa connect --stage discover --endpoint https://gateway.example/v1 \
+  --api-key-env MODEL_KEY > discovery.json
+uv run nooa connect your-model --stage plan --api-style chat \
```
sklinglernv, line 31: How do you know the API style upfront?

docs/model-connect.md
```diff
@@ -0,0 +1,519 @@
...
+proposal = connect.plan(
+    "work", "your-model", "chat", "https://gateway.example/v1", "MODEL_KEY",
+    budget_tokens=65536,
+    reasoning_levels={"high": {"reasoning_effort": "high"}},
+)
+result = await connect.check_stage(proposal, "reasoning")
+checks = result.entry["provenance"]["probes"]
+handoff = connect.diagnostic_prompt("reasoning", result.entry, checks)
+# Explicit persistence, when wanted:
+# connect.write(result.entry, registry_path, alias=result.alias)
```
sklinglernv, line 143: Does connect verify against persisted settings?

sklinglernv:
**[AI Take]** Review of `9950645d` (full diff, ~10k lines). The connect suite passes locally (359 tests); the three issues below are in paths the suite doesn't currently cover, and each was confirmed against the source.

---

### 1. Reasoning-level probes that carry their own reply cap are always skipped

`src/nooa/unifiedllm/connect/__init__.py:770`

The level probe is budgeted at `reasoning_output_tokens + 512`:

```python
probes.append(Probe(f"level:{label}", level_body, reasoning_output_tokens + 512, 120))
```

but lines 790-792 then overwrite the body with the level's own reply cap:

```python
for probe in probes:
    if probe.name.startswith("level:"):
        probe.body.update(entry["reasoning_levels"][probe.name.removeprefix("level:")])
```

`token_estimate` is not updated to match. For any level whose cap exceeds the probe budget, the guard at `__init__.py:1115` rejects it:

```python
or cap > probe.token_estimate - 512
...
record["reason"] = "declared output cap exceeds the approved probe cap"
```

Concretely with the built-in `budget` template: `configure_entry:510` bumps the level cap to `thinking.budget_tokens + 1024` = 5120, while `token_estimate` stays `4096 + 512` = 4608, so the check is `5120 > 4096` and the probe is skipped. The reasoning-level check never runs for these levels — and it fails as a budget rejection rather than surfacing as a capability result.

Suggested fix: derive `token_estimate` after the level patch is applied (or set it from the level's effective cap `+ 512`).

### 2. `KeyError: 'encrypted_reasoning'` when re-running against an existing Responses entry

`src/nooa/unifiedllm/connect/__init__.py:1080` and `:1198`

Both sites index the provenance record unconditionally:

```python
if probe.name == "routing" and entry.get("include"):
    provenance["encrypted_reasoning"]["outcome"] = "accepted"
```

but `configure_entry` only creates `provenance["encrypted_reasoning"]` when it adds the `include` itself. An existing Responses entry that already carries `include: [reasoning.encrypted_content]` passes the `entry.get("include")` truth test with no such record, so the routing probe raises `KeyError` and aborts — on the reused-result path (1080) and the fresh-probe path (1198).

Suggested fix: `provenance.setdefault("encrypted_reasoning", {})["outcome"] = "accepted"` at both sites.

### 3. `entries()` hard-fails on a null `models:` placeholder the runtime tolerates

`src/nooa/unifiedllm/connect/_registry.py:21`

```python
if not isinstance(data, dict) or not isinstance(data.get("models", {}), dict):
    raise ValueError(f"Registry {path} must contain a models mapping")
```

A file containing just `models:` parses to `{"models": None}`, so `data.get("models", {})` is `None` and this raises. The runtime loader deliberately accepts exactly that shape (`src/nooa/unifiedllm/registry.py:163-167`):

```python
models = data.get("models")
if models is None:
    return {}
```

Since `entries()` scans the whole `llm_config_chain()`, a single such placeholder layer makes every `nooa connect` invocation fail — including with `--output` pointing at an unrelated file, and including `--stage` runs. Worth matching the loader's tolerance here (skip the layer rather than raise), since connect is the tool a user reaches for when their registry is in a half-configured state.

---

Checked and found sound, for the record: the atomic/TOCTOU logic in `secrets.write_secret_env` and `connect.write`; the YAML splicing and `shadowing_source` precedence math; the `auto` cache-policy branch and its `explicit` return flag; `ResponsesClient._prepare_call_config` reply-limit alias resolution against `apply_reasoning_level`'s new cap-conflict check; the `include: []` opt-out in `replay_state`; the `tool_choice="auto"` drop (deliberate and tested); budget accounting across `check_interfaces` / `run_steps` / `session_steps`; and nooa-cli exit-code passthrough. The reasoning puzzle's expected answer `BGDACEFH` is the unique solution to the stated constraints.

packages/nooa-cli/src/nooa_cli/commands/_connect_stages.py
```diff
@@ -0,0 +1,302 @@
...
+                or not isinstance(data.get("alias"), str)
+                or not data["alias"]
+                or not isinstance(data.get("entry"), dict)
+            ):
+                raise click.UsageError("Save input requires an alias string and an entry mapping")
+            alias, entry = (
+                data["alias"],
+                connect.configure_entry(data["entry"], reply_tokens=reply_tokens),
+            )
+            if "api_key" in entry:
```
sklinglernv: [AI Take] **P1 — Stage save allows nested plaintext API keys and echoes them.** This check only rejects a top-level `api_key`. An entry containing `extra_body: {api_key: secret}` passes `configure_entry()`, is persisted, and is returned verbatim on JSON stdout; at runtime, `extra_body` is forwarded into the provider request body. I reproduced this locally: stage save exited 0 and the secret appeared in both stdout and the saved YAML. Please recursively reject credential-bearing fields in stage-save input—especially under `extra_body` and `reasoning_levels`—and scrub serialized output.

src/nooa/unifiedllm/connect/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+    path.parent.mkdir(parents=True, exist_ok=True)
+    with tempfile.NamedTemporaryFile(
+        mode="w", encoding="utf-8", dir=path.parent, prefix=path.name + ".", delete=False
+    ) as temporary:
+        temporary.write(text)
+        temporary.flush()
+        os.fsync(temporary.fileno())
+        temporary_path = Path(temporary.name)
+    try:
+        if (path.read_text() if path.exists() else None) != original:
```
sklinglernv: [AI Take] **P2 — Concurrent registry writes can silently lose updates.** The changed-file comparison and `os.replace()` are separate operations. Two Connect processes can read the same original registry, both pass this check, then replace sequentially; the second write silently discards the first process’s alias while both report success. Please serialize the complete read-modify-write operation with an inter-process lock or equivalent concurrency mechanism.

src/nooa/unifiedllm/reasoning.py
```diff
@@ -117,4 +117,12 @@ def apply_reasoning_level(
     if isinstance(extra, Mapping) and patch.keys() & extra.keys():
         params["extra_body"] = {key: value for key, value in extra.items() if key not in patch}
     params.update(patch)
+    # Reply-limit aliases are one setting, even when the destination API names
```
sklinglernv: question: I don't understand this comment. Maybe simplify the language.

src/nooa/unifiedllm/registry.py
```diff
@@ -350,6 +348,21 @@ def get_llm_client(name: str, *, client_type: str | None = None, **overrides) ->
     with _registry_lock:
         config = dict(MODELS.get(name, {}))
 
+    return client_from_config(name, config, client_type=client_type, **overrides)
+
+
+def client_from_config(
```
sklinglernv, line 354: How is this different from the `get_llm` that we had?

src/nooa/secrets.py
```diff
@@ -68,4 +68,52 @@ def load_secrets_into_env() -> list[str]:
     return applied
 
 
-__all__ = ["load_secrets_into_env"]
+def write_secret_env(path, name: str, value: str) -> None:
```
sklinglernv, line 71: Why do we need this?

src/nooa/unifiedllm/connect/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+    )
+    settings_sent = []
+
+    async def capture(request):
+        from nooa.unifiedllm.connect._session import settings_on_wire
+
+        if probe.name.startswith("level:"):
+            settings_sent.append(settings_on_wire(settings, json.loads(request.content)))
+
+    hooks = client._http.httpx_async.event_hooks["request"]
```
sklinglernv, line 1030: **Capture hook never fires when there is no API key — every reasoning level and the whole conversation check are falsely reported as failures on unauthenticated endpoints.**

The CLI explicitly offers `-` for "no authentication" (`connect.py:363`) — i.e. local vLLM / Ollama / llama.cpp. On that path `client_from_config(alias, entry, api_key=None, …)` reaches `unifiedllm.py:262-277`, `AsyncOpenAI(...)` raises *"The api_key client option must be set"*, and the documented fallback sets `async_client = None` so **litellm builds its own httpx client**. Verified:

```
>>> c = client_from_config('local', {'model_name':'openai/local-model','client_type':'completion',
...     'api_style':'chat','api_base':'http://localhost:8000/v1','api_key_env':''}, api_key=None)
>>> c._http.async_client, c._http._openai_clients
(None, [])
```

So the hook installed here (and the identical one at `_session.py:240`) never runs. Consequences:

* `_run_probe` returns `settings_sent=False` → `run_steps:1191-1195` rewrites every *accepted* `level:*` probe to `outcome: not_confirmed`, reason *"The client did not send the requested settings; check parameter filtering"* — pointing the user at parameter filtering for a bug that is purely a missing hook.
* `_session.py:328` sees `len(bodies) != before + 1` → the cache and reasoning-retention checks abort with *"request capture missing or server exceeded the reply cap"*.

No test exercises an empty `api_key_env` through the probe path. Either detect that the hook is not installed (`client._http.async_client is None`) and record *"could not observe the request"* instead of *"settings not sent"*, or capture at a layer litellm cannot bypass.

packages/nooa-cli/pyproject.toml
```diff
@@ -14,6 +14,7 @@ dependencies = [
     "click>=8.1.0",
     # CLI config files
     "pyyaml>=6.0",
+    "prompt-toolkit>=3.0.50",
```
sklinglernv: **`prompt-toolkit>=3.0.50` is two releases too low — `show_frame` first exists in 3.0.52.**

`_connect_prompts.py:152` passes `show_frame=True` unconditionally. Verified by downloading the sdists:

| version | `show_frame` in `shortcuts/prompt.py` |
|---|---|
| 3.0.50 | absent |
| 3.0.51 | absent |
| 3.0.52 | present |

`PromptSession.prompt()` in 3.0.51 is keyword-only with no `**kwargs`, so the call raises `TypeError: prompt() got an unexpected keyword argument 'show_frame'`. Any environment that already satisfies `>=3.0.50` (e.g. a venv with ipython/jupyter, which commonly pin 3.0.51) will not upgrade, and `nooa connect` dies on its **first** question. `TypeError` is not in the caught tuple at `connect.py:1051`, so it surfaces as a raw traceback. The committed `uv.lock` pins 3.0.53, which is why CI never sees it.

```suggestion
    "prompt-toolkit>=3.0.52",
```

src/nooa/secrets.py
```diff
@@ -68,4 +68,52 @@ def load_secrets_into_env() -> list[str]:
...
+    path = Path(path)
+    original = path.read_text() if path.exists() else None
+    try:
+        with path.open() as source:
+            data = yaml.safe_load(source) or {}
+    except FileNotFoundError:
+        data = {}
+    except yaml.YAMLError:
+        raise ValueError(f"Secrets file {path} contains invalid YAML; no changes made") from None
+    if not isinstance(data, dict) or not isinstance(data.get("env", {}), dict):
```
sklinglernv, line 99: **A `secrets.yaml` with a present-but-null `env:` key aborts the whole onboarding run — after the paid checks.**

Verified by execution:

```python
>>> p.write_text("env:\n")            # scaffold file, or all keys commented out
>>> write_secret_env(p, "MY_KEY", "sk-123")
ValueError: Secrets file …/secrets.yaml must contain an env mapping
```

`yaml.safe_load("env:\n")` is `{'env': None}`, so `data.get("env", {})` returns `None`, not `{}`. `load_secrets_into_env` **in this same module** (line 51) explicitly tolerates that shape and returns `[]`.

The damage is amplified by call ordering: in the wizard `write_secret_env` runs at `connect.py:1031` (inside `if save_key:`) while `connect.write(result.entry, path, alias=alias)` is only at `connect.py:1040`. The `ValueError` propagates to the handler at `connect.py:1051`, so **the model alias is never written** — the user answered every prompt, paid for every probe, and ends up with neither the key nor the alias.

```suggestion
    if not isinstance(data, dict) or not isinstance(data.get("env") or {}, dict):
```

Worth also writing the registry entry *before* the secret, so a failure in either write can't silently discard the other.

src/nooa/unifiedllm/connect/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+    Splice the selected YAML entry instead of reformatting the whole file, so
+    unrelated entries and comments remain intact. The final replace is atomic.
+    """
+    entry = configure_entry(entry)
+    original = path.read_text() if path.exists() else None
+    source = original or ""
+    data = yaml.safe_load(source)
+    if data is None:
+        data = {}
+    if not isinstance(data, dict) or not isinstance(data.get("models", {}), dict):
```
sklinglernv, line 1513: **A registry file whose `models:` key is present but empty (`models:\n`) is rejected everywhere, and crashes `--stage save` with a TypeError.**

`yaml.safe_load("models:\n")` → `{'models': None}`, so `data.get("models", {})` returns `None`, not `{}`. The runtime loader explicitly tolerates this shape (`registry._load_models_from_yaml`: `if models is None: return {}`), so `reload_registry()` happily returns `{}` for the same file. Five sites in this PR disagree:

| site | behaviour (verified) |
|---|---|
| `connect/__init__.py:1310` (`write`) | `ValueError: Registry must be a mapping with a models mapping` — cannot populate an empty scaffold |
| `_connect_registry.py:21` (`entries`) | `ValueError: Registry <path> must contain a models mapping` |
| `connect.py:270`, `:982` | `ClickException("Registry must contain a models mapping.")` |
| `_connect_stages.py:117` | `alias in None` → **`TypeError: argument of type 'NoneType' is not iterable`** |

`entries()` is called at `connect.py:221` for **every** file in `llm_config_chain()`, so one such file anywhere in the chain makes `nooa connect` unusable end-to-end — even with `--no-probe`, even when saving to an unrelated `--output`. Reproduced `--stage save` returning `{"error": {"type": "TypeError", …}}`.

That `write()` *intends* to support this state is visible three lines below, at 1321: `alias_indent = models_node.start_mark.column if models_node and models_node.value else 2` explicitly handles an empty `models` node.

Fix is `data.get("models") or {}` at all five sites (and the matching `env` case in `secrets.py`).

```suggestion
    if not isinstance(data, dict) or not isinstance(data.get("models") or {}, dict):
```

packages/nooa-cli/src/nooa_cli/commands/_connect_registry.py
```diff
@@ -0,0 +1,208 @@
...
+    if extra_path is not None and extra_path.exists():
+        paths = [p for p in paths if p.resolve() != extra_path.resolve()] + [extra_path]
+    resolved = {}
+    for path in paths:
+        with Path(path).open() as source:
+            data = yaml.safe_load(source) or {}
+        if not isinstance(data, dict) or not isinstance(data.get("models", {}), dict):
+            raise ValueError(f"Registry {path} must contain a models mapping")
+        for alias, entry in data.get("models", {}).items():
+            if isinstance(alias, str) and isinstance(entry, dict):
```
sklinglernv, line 26: **`entries()` resurrects aliases that a higher layer has tombstoned with `null`.**

`reload_registry` treats `models: {alias: null}` as a deletion (`registry.py:233-238` — `if cfg is None: fresh.pop(name, None)`, and the same for any non-mapping). This loop only *skips* those values, so the lower layer's entry survives. Verified:

```
low.yaml:  models: {ghost: {model_name: openai/ghost}}
high.yaml: models: {ghost: null}
NEMO_OO_LLM_CONFIG=low.yaml,high.yaml

runtime registry sees ghost: False
connect entries() sees ghost: True
```

Concretely: a project that nulls out a bundled alias still sees it offered by `nooa connect --edit-model`, edits it, and writes the **resurrected** entry back into `llm_config.yaml`; `shadowing_source` warns that a deleted definition "currently defines this alias and takes precedence"; and `credential_names` suggests an `api_key_env` from an alias the runtime cannot resolve.

Rather than re-implementing the merge, consider reusing `registry._load_models_from_yaml` + the same pop semantics, so the two copies cannot drift on the next layering rule.

src/nooa/unifiedllm/connect/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+        models = {**data["models"], alias: entry}
+        # An inline mapping must be replaced as a unit, not patched by lines.
+        replacement = yaml.safe_dump(models, default_flow_style=True, sort_keys=False).strip()
+        text = (
+            source[: models_node.start_mark.index]
+            + replacement
+            + source[models_node.end_mark.index :]
+        )
+    elif alias in data["models"]:
+        key, value = next((key, value) for key, value in models_node.value if key.value == alias)
```
sklinglernv, line 1559: **`next(...)` with no default → bare `StopIteration`, and YAML anchors make `write()` fail outright — both after the paid checks.**

Reproduced with a registry that uses a merge key inside `models:`:

```yaml
defaults: &base
  shared: {model_name: openai/base, api_style: chat}
models:
  <<: *base
  other: {model_name: openai/other}
```

`yaml.safe_load` resolves `data['models']` to `{'shared', 'other'}`, so line 1343's `alias in data["models"]` is True — but `models_node.value` contains only the `<<` key, so no `key.value == 'shared'` match exists and this `next()` raises a **bare `StopIteration` with an empty message**. `StopIteration` is not in the caught tuple at `connect.py:1051`, so the user gets a raw traceback and loses a run they already paid for.

Separately, an anchored entry *value*:

```yaml
defaults: &d
  model_name: openai/a
models:
  a: *d
```

makes `content_end()` descend into the anchored node, whose `end_mark` points **earlier** in the file, so the splice lands in the wrong place. The `yaml.safe_load(text) != expected` net catches it (good), but the result is that even adding a *brand-new, unrelated* alias fails with "Could not construct the registry update without changing other entries" — i.e. `nooa connect` cannot save into any registry that uses anchors.

At minimum use `next(…, None)` and raise the same descriptive `ValueError`; ideally detect a merge key / anchored value and fall back to a whole-file rewrite.

src/nooa/unifiedllm/reasoning.py
```diff
@@ -117,4 +117,12 @@ def apply_reasoning_level(
...
+    # it differently. A level's cap replaces inherited defaults as a unit.
+    caps = {"max_tokens", "max_completion_tokens", "max_output_tokens"}
+    if caps & patch.keys():
```
sklinglernv: **The new cap-alias unification never looks inside `extra_body`, so a level's reply cap is silently defeated.**

The block three lines above deliberately strips replaced defaults out of `extra_body` ("*SDKs otherwise merge those back over the selected top-level values*"). The new cap-alias rule doesn't do the same for *sibling* aliases. Verified by running `apply_reasoning_level` directly with `levels={'high': {'reasoning': {'effort':'high'}, 'max_tokens': 65536}}`:

| input | result |
|---|---|
| `defaults={'max_output_tokens': 4096}` | ✅ scrubbed → `{'reasoning': …, 'max_tokens': 65536}` |
| `overrides={'extra_body': {'max_tokens': 500}}` | ✅ `ValueError: … conflicts with explicit request field(s)` |
| `defaults={'extra_body': {'max_output_tokens': 4096}}` | ❌ `{'extra_body': {'max_output_tokens': 4096}, 'max_tokens': 65536}` |
| `overrides={'extra_body': {'max_output_tokens': 500}}` | ❌ same, no error |

litellm merges `extra_body` into the HTTP body, so the effective Responses cap is 4 096 while the level asked for 65 536 — high-effort replies burn the reasoning budget and return `finish_reason='length'`. Connect-written entries are shielded by `configure_entry`'s `managed & extra.keys()` check, so this only bites hand-written aliases, i.e. every non-Connect user of this shared module.

Extend the `caps` handling to the `extra_body` mapping the same way the `patch.keys() & extra.keys()` branch already does.

src/nooa/unifiedllm/connect/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+    )
+    probes.append(Probe("tools", tool_body, output_tokens + 512))
+    for label, params in levels.items():
+        level_body = deepcopy(body)
+        level_body[token_key] = reasoning_output_tokens
+        level_body["input" if api_style == "responses" else "messages"][0]["content"] = (
+            REASONING_CHECK_PROMPT
+        )
+        level_body.update(params)
+        probes.append(Probe(f"level:{label}", level_body, reasoning_output_tokens + 512, 120))
```
sklinglernv, line 914: **`--reasoning-template budget` can never actually check a level at the documented defaults.**

`level_body.update(params)` on the line above overwrites the body's cap with the level's own cap, but `Probe.token_estimate` here is still sized from `reasoning_output_tokens`. Verified:

```
>>> connect.reasoning_settings('budget', 'chat', 'high')
{'thinking': {'type': 'enabled', 'budget_tokens': 4096}, 'max_tokens': 5120}
>>> [p for p in plan(..., reasoning_levels=…).probes if p.name.startswith('level:')]
probe level:high  caps [5120]  token_estimate 4608  ->  guard `cap > est-512` : True
```

`run_steps:1115-1122` therefore records `outcome: not_probed`, reason *"declared output cap exceeds the approved probe cap"* for **every** level. The CLI renders that as a grey skip, exits 0, and writes the alias with an entirely unverified reasoning level.

The thinking budget is hard-coded — `reasoning_settings`' `budget=4096` kwarg is never passed from the CLI (`connect.py:741`) — so there is no flag to lower it; the user would have to know to pass `--reasoning-output-tokens 5632`. The same path fires for any level whose thinking budget triggers `configure_entry`'s documented auto-bump at line 510.

Only the `effort` template is covered by tests, and the reason string *"declared output cap exceeds the approved probe cap"* appears nowhere in the test suite. Size `token_estimate` from the **effective (post-merge) cap** instead of `reasoning_output_tokens`.

src/nooa/unifiedllm/connect/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+            record["settings_sent"] = settings_sent
+            if not settings_sent:
+                record["outcome"] = "not_confirmed"
+                record["reason"] = (
+                    "The client did not send the requested settings; check parameter filtering"
+                )
+        spent += max(0, tokens - probe.token_estimate)
+        yield ProbeUpdate(probe.name, deepcopy(record))
+        if probe.name == "routing" and entry.get("include"):
+            provenance["encrypted_reasoning"]["outcome"] = "accepted"
```
sklinglernv: **`provenance["encrypted_reasoning"]` is written unguarded, but `configure_entry` only creates it when it *adds* the include.**

`run_steps` calls `configure_entry(proposal.entry)` itself, and `configure_entry` (lines 520-538) writes `provenance["encrypted_reasoning"]` only in the two branches where it *inserts* `reasoning.encrypted_content`. If `include` is already `['reasoning.encrypted_content']` — which is exactly what a previously-saved entry looks like — none of the branches fire, and this line raises `KeyError: 'encrypted_reasoning'` **after the routing request has already been sent and billed**, so the caller loses the result. The same unguarded write is at line 1080 in the probe-reuse branch.

Repro shape (the module docstring advertises this library as "shared by the CLI and TUI", so any second frontend that round-trips an entry hits it):

```python
e = connect.configure_entry(connect.plan(..., 'responses', ...).entry)
e['provenance'].pop('encrypted_reasoning')          # or: entry loaded from YAML
async for _ in connect.run_steps(replace(p, entry=e), approved='minimal', api_key=k): ...
# KeyError: 'encrypted_reasoning'
```

A hand-written *chat*-style entry carrying an `include` key reaches the same line too, since `configure_entry` only manages `include` for Responses entries. Use `provenance.setdefault("encrypted_reasoning", {})["outcome"] = "accepted"`, or have `configure_entry` always establish the key.

src/nooa/unifiedllm/replay_state.py
```diff
@@ -332,6 +332,9 @@ def prepare_chat_messages(
 def add_encrypted_reasoning_include(api_params: dict[str, Any], scope: str | None) -> None:
     """Request OpenAI encrypted reasoning only on endpoints known to support it."""
     configured = api_params.get("include")
+    if configured == []:
```
sklinglernv: **The opt-out matches only a literal `list`; an empty tuple or set falls through and gets the include added back.**

The very next line already treats list/tuple/set as equivalent, so the two checks disagree on adjacent lines. Verified against scope `responses:openai:https://api.openai.com/v1`:

```
include=[]      -> key removed        ✅
include=()      -> ['reasoning.encrypted_content']   ❌
include=set()   -> ['reasoning.encrypted_content']   ❌
```

`get_llm_client('alias', include=())`, or any config layer that normalises sequences to tuples, therefore sends the exact field the opt-out exists to suppress — and on the endpoint Connect already recorded as rejecting it (`provenance.encrypted_reasoning.outcome == 'rejected'`, written by `_disable_encrypted_reasoning`) the call fails with the very 400 this opt-out was added to avoid.

```suggestion
    if isinstance(configured, (list, tuple, set)) and not configured:
```

packages/nooa-cli/src/nooa_cli/commands/connect.py
```diff
@@ -0,0 +1,1071 @@
...
+            )
+        patches = None
+        if editing is not None:
+            labels = candidate.get("reasoning", {}).get("supported_efforts", [])
+            original_levels = editing.get("reasoning_levels", {})
+            if not levels_file and any(label not in original_levels for label in labels):
+                raise click.UsageError(
+                    "New reasoning levels need request settings; supply --levels-file"
+                )
+            patches = {label: deepcopy(original_levels[label]) for label in labels}
```
sklinglernv: **`KeyError` on exactly the recovery path the error message four lines above recommends.**

The guard on line 727 is skipped when `levels_file` is truthy, but this comprehension still indexes `original_levels[label]` for every label. Sequence:

1. `nooa connect --edit-model saved --levels-file levels.yaml`
2. At the interactive *"Reasoning levels"* prompt the user adds `ultra`, which the saved entry doesn't have — precisely what lines 728-730 instruct: *"New reasoning levels need request settings; supply --levels-file"*.
3. `original_levels['ultra']` → `KeyError: 'ultra'`.

`KeyError` is not in the `except (ValueError, OSError, yaml.YAMLError, httpx.HTTPError)` handler at line 1051, so the command dies with a raw traceback and exit 1.

The comprehension's result is immediately overwritten by lines 732-734 on this branch anyway, so it is pure dead work:

```suggestion
            if not levels_file:
                patches = {label: deepcopy(original_levels[label]) for label in labels}
```

src/nooa/unifiedllm/connect/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+            for effort in efforts
+        )
+        enabled |= isinstance(thinking, dict) and thinking.get("type") in ("enabled", "adaptive")
+        enabled |= isinstance(template, dict) and template.get("enable_thinking") is True
+        if enabled:
+            missing.append(label)
+    return missing
+
+
+def configure_entry(entry: dict, *, reply_tokens: int | None = None) -> dict:
```
sklinglernv, line 481: **A literal `api_key:` survives `configure_entry` and is written straight into `llm_config.yaml` — only `--stage save` rejects it.**

Verified by execution:

```python
entry = {..., 'api_key': 'sk-LITERAL-SECRET'}
configure_entry(entry)   # -> 'api_key' still present
connect.write(entry, p, alias='mine')
'sk-LITERAL-SECRET' in p.read_text()   # True
```

`_connect_stages.py:108` is the *only* guard in the PR:

```python
if "api_key" in entry:
    raise click.UsageError("Use api_key_env, never a literal key")
```

The interactive path never checks — `--edit-model` deep-copies the existing entry into `merged` at `connect.py:763`, preserving every custom field. So `nooa connect --edit-model foo` cheerfully rewrites a plaintext credential into `llm_config.yaml` (a file that is frequently committed), while `nooa connect --stage save` refuses the byte-identical entry.

Move the check into `configure_entry`, which both frontends already call, so every future save-time rule protects both.

_(Distinct from the existing thread on `_connect_stages.py:108`, which is about **nested** keys inside that one frontend's check: this is about the check not existing at all on the interactive path, for a plain top-level `api_key`.)_

packages/nooa-cli/src/nooa_cli/commands/_connect_stages.py
```diff
@@ -0,0 +1,302 @@
...
+def read_discovery(path, endpoint):
+    """Read a previous discovery result only for the selected endpoint."""
+    import json
+    from pathlib import Path
+
+    from nooa.unifiedllm import connect
+
+    document = json.loads(Path(path).read_text())
+    data = document.get("data", document)
+    if not isinstance(data, dict) or connect.normalize_endpoint(
```
sklinglernv: **A discovery file produced by `--stage discover` is rejected when replayed with the same `--endpoint` the user originally typed.**

`discover()` rewrites its base on a 404 (`connect/__init__.py:264-267`: `base += "/v1"`) and returns `Discovery(base, …)`, so the saved JSON records the rewritten form. This comparison then fails on exact normalized equality:

```
$ nooa connect --stage discover --endpoint https://api.test     # GET /models -> 404, falls back to /v1
   ... d.json: {"data": {"api_base": "https://api.test/v1", ...}}
$ nooa connect MODEL --discovery-file d.json --endpoint https://api.test
   ValueError: Discovery metadata belongs to a different endpoint
```

The endpoint is in fact identical; only the adapter's normalization of `discover()`'s own `/v1` fallback differs. Accept a match after `removesuffix("/v1")` on both sides (the way `_connect_registry.credential_names` already does).

Minor, same line: `data.get("api_base", "")` on a document without that key calls `normalize_endpoint("")`, which raises the generic *"Endpoint must be an HTTP(S) URL without credentials, query or fragment"* instead of the intended message.

src/nooa/unifiedllm/connect/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+        insertion = sum(
+            len(line) for line in source.splitlines(keepends=True)[: end.line + bool(end.column)]
+        )
+        prefix = "" if insertion == 0 or source[insertion - 1] == "\n" else "\n"
+        text = source[:insertion] + prefix + indented + source[insertion:]
+    expected = {**data, "models": {**data.get("models", {}), alias: entry}}
+    if yaml.safe_load(text) != expected:
+        raise ValueError("Could not construct the registry update without changing other entries")
+    path.parent.mkdir(parents=True, exist_ok=True)
+    with tempfile.NamedTemporaryFile(
```
sklinglernv: **`write()` silently narrows the registry file to 0600, replaces a symlink with a regular file, and leaks the temp file on I/O errors.**

1. **Mode.** `tempfile.NamedTemporaryFile` creates 0600 and `os.replace` carries the temp inode's mode onto the destination. Verified: an `llm_config.yaml` at 0644 comes back as **0600** after `connect.write()`, with no warning and no mention in the *"Saved &lt;alias&gt; to &lt;path&gt;"* message. Contrast `write_secret_env`, which chmods 0600 *deliberately* because it holds a credential — the registry holds only `api_key_env` **names**. A project- or system-level registry readable by a service account or another user silently loses access. Capture `path.stat().st_mode` first and restore it (or use `0o666 & ~umask` for new files).

2. **Symlink.** `~/.config/nooa/llm_config.yaml` symlinked into a dotfiles repo is replaced by a regular file; the tracked target keeps the old content and silently stops receiving updates. `path.resolve()` before writing.

3. **Temp leak.** `temporary_path = Path(temporary.name)` is the **last** statement inside the `with`, so an `OSError` during `write`/`flush`/`fsync` (ENOSPC, EDQUOT, EIO) escapes before the `try/finally` that unlinks it — a `llm_config.yaml.<random>` droppings file is left next to the registry on every failed attempt. `nooa.secrets.write_secret_env` gets this right by binding `temporary = None` before the `try` and assigning it as the first statement inside the `with`.

_(Distinct from the existing thread on line 1368 about the concurrent-write race — these three are independent of the optimistic-lock question.)_

src/nooa/unifiedllm/connect/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+            tokens = usage.input_tokens + usage.output_tokens if usage else 0
+        except Exception as exc:
+            from nooa.unifiedllm.connect._diagnostics import timeout_details
+
+            status = getattr(exc, "status_code", None)
+            record["error"] = type(exc).__name__
+            record["elapsed_seconds"] = round(time.monotonic() - started, 3)
+            if isinstance(status, int):
+                record["status_code"] = status
+                record["outcome"] = "rejected" if status == 400 else "not_probed"
```
sklinglernv: **A 422 rejection is persisted and displayed as "Not checked".**

Only `400` maps to `rejected`; every other status falls to `not_probed`, whose documented meaning is *we never checked*. But the request **was** sent and refused. 422 is the common shape for an unsupported tools schema or reasoning field on vLLM/FastAPI-based gateways — and `_include_rejected` at line 58 explicitly treats `{400, 422}` as a field rejection, so the two halves of the same feature disagree about what a 422 means.

The display compounds it: `_connect_view.check_failure` has arms for 401/403, 404, 429, `>=500`, `Timeout`, `ReasoningReplayError` and `APIConnectionError` — none for 422 — and the record carries no `reason`, so `detail` is `None` and the row renders from the fallback table:

```
! Tool use: Not checked
```

The same mislabel goes into `--stage`'s JSON report and into the diagnostic prompt handed to an agent. Suggest treating any 4xx that isn't 401/403/429 as `rejected` (or at least adding 422), and giving `check_failure` a generic 4xx arm.

sklinglernv (approved):
Overall looks good to me. A few comments from me and AI.

sklinglernv:
**[AI Take]** Follow-up sweep at head `34f3a919`, six defects not covered by the inline comments already on this PR or by my earlier comment. (That earlier comment cites `9950645d`; it was one commit behind. All three findings in it still apply here.)

Each mechanism below was confirmed by reading the code at this head.

---

### 1. You cannot answer `Y` to a `[Y/n]` prompt

[`_connect_prompts.py:177-187`](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/34f3a91955d111ccfb1df34e0db24aadbf4c2609/packages/nooa-cli/src/nooa_cli/commands/_connect_prompts.py#L176-L188)

`confirm()` renders `" [Y/n]"` but passes `choices=("yes", "no", "y", "n", "")`. The validator is an exact membership test with no case folding ([line 75-79](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/34f3a91955d111ccfb1df34e0db24aadbf4c2609/packages/nooa-cli/src/nooa_cli/commands/_connect_prompts.py#L74-L80)):

```python
validator = Validator.from_callable(
    lambda value: (
        effective(value) in choices if choices else ...
    ),
    error_message="Choose a matching value." if choices else "Enter a value.",
```

and `effective()` only substitutes the default, it does not normalize case. So in a real terminal, typing the capital letter the prompt itself displays fails with "Choose a matching value." and refuses to submit — likewise `N`, `Yes`, `No`. The user is stuck until they guess lowercase, or Ctrl-C, which raises `click.Abort` and ends setup.

This is invisible to the suite because `CliRunner` supplies a non-tty stdin, which takes the `click.confirm` branch at [line 179](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/34f3a91955d111ccfb1df34e0db24aadbf4c2609/packages/nooa-cli/src/nooa_cli/commands/_connect_prompts.py#L178-L180) — that one *is* case-insensitive. The only tty-mode confirm test uses lowercase `y`.

Worth prioritizing: it's the first interaction a human has with the wizard, and it affects the budget-approval prompt.

### 2. `--stage save` reports failure after the entry is already written

[`_connect_stages.py:121-127`](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/34f3a91955d111ccfb1df34e0db24aadbf4c2609/packages/nooa-cli/src/nooa_cli/commands/_connect_stages.py#L120-L128)

```python
connect.write(entry, path, alias=alias)
data = {"alias": alias, "path": str(path), "entry": entry}
from ._connect_registry import shadowing_source

if shadow := shadowing_source(alias, path):
```

`shadowing_source()` runs *after* the write and inside the same `try`. It calls `entries()`, which raises on any malformed file anywhere in `llm_config_chain()` — not just the `--output` target. The handler then discards `data` and returns `ok: false`, `data: null`, exit 1, while the alias is on disk.

An agent driving the documented stage interface sees a clean failure and will retry or roll back against a registry that was in fact modified. The interactive path calls `shadowing_source` *before* the write, so only stage mode has this ordering. Swapping the two lines fixes it.

### 3. The `encrypted_reasoning` KeyError also hits non-Responses entries

[`__init__.py:1199`](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/34f3a91955d111ccfb1df34e0db24aadbf4c2609/src/nooa/unifiedllm/connect/__init__.py#L1198-L1200) and [`:1080`](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/34f3a91955d111ccfb1df34e0db24aadbf4c2609/src/nooa/unifiedllm/connect/__init__.py#L1079-L1081)

Broader than reported earlier. The write is guarded only by `entry.get("include")`, but `plan()`/`configure_entry` create that provenance key only for `api_style == "responses"`. `registry.client_from_config` copies `include` into client params for *every* client type, so a chat- or anthropic-style entry carrying `include` reaches the same crash via `--edit-model` — after the routing probe has been paid for. `setdefault` at both sites still covers it.

### 4. Interface checks relabel every inner event, producing contradictory rows

[`__init__.py:1294`](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/34f3a91955d111ccfb1df34e0db24aadbf4c2609/src/nooa/unifiedllm/connect/__init__.py#L1293-L1295)

```python
yield ProbeUpdate(style, event.outcome)
```

`check_interfaces` rewrites the name of every event from the inner `run_steps`, so an `encrypted_reasoning` rejection and the re-queued `routing` retry both surface under the interface name. Against a Responses endpoint that rejects the include with 400, the user sees the interface reported as rejected twice *and* connected, and never learns the rejected thing was the encrypted-reasoning setting rather than the interface itself. `CheckProgress.update` also pops the timer on the first mislabelled row, so the genuine routing failure loses its elapsed time.

### 5. Edited limits are persisted as endpoint-reported facts

[`_connect_prompts.py:207-212`](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/34f3a91955d111ccfb1df34e0db24aadbf4c2609/packages/nooa-cli/src/nooa_cli/commands/_connect_prompts.py#L206-L213)

`edit_model_details` overwrites `context_length` and `top_provider.max_completion_tokens` on a deep copy but never touches `limit_sources`. A user who edits the context window to 999 gets a saved entry claiming `context_window: 999` alongside `limit_sources: {context_length: endpoint}` and `endpoint_limits: {context_window: 128000}` — the same file asserting the endpoint reported both numbers, with the typed reply ceiling landing in `provenance.catalogue_limits`. The confirmation screen reads the same stale `limit_sources` and prints "999 tokens · endpoint" directly under "Your edited settings."

Given this PR's stated care about limit provenance ("per-field sources", "never invented total context"), this seems worth closing.

### 6. Vacuous assertion hides the encrypted-reasoning invariant

[`test_connect_interfaces.py:194`](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/34f3a91955d111ccfb1df34e0db24aadbf4c2609/tests/unifiedllm/connect/test_connect_interfaces.py#L193-L195)

```python
assert "encrypted_reasoning" not in entry
```

Nothing ever writes that key to the entry top level — it goes to `entry["provenance"]["encrypted_reasoning"]`. The assertion cannot fail, so it would still pass if `_disable_encrypted_reasoning` recorded nothing at all or recorded the wrong outcome. The real invariant is `provenance.encrypted_reasoning.outcome == "rejected"`, which `plan(existing_entry=...)` later relies on to keep `include: []` on reconnect, and it is untested here.

---

A pattern across #1, #2 and #6: the suite is green, and several of these sit in paths it structurally cannot reach — tty-mode input, stage-mode ordering, and an assertion aimed at the wrong object.

docs/model-configuration.md
furgalep: Addressed in 63dbd4c1: added skills/nooa-model-configuration/SKILL.md for endpoint/registry setup, credentials, reply budgets, reasoning, caching and Connect diagnostics. Removed the setup instructions and configuration trigger from agent authoring, leaving only a link and agent-side client selection/inheritance. Connect help and diagnostic handoffs now name the dedicated skill, including absolute installation paths and pinned-source fallback. The executable reasoning example moved with its test. Skill validation and 424 Connect/CLI/reasoning/doc tests pass; no new dependencies or paid calls.

docs/model-connect.md
furgalep: We do not infer it from the model name. The wizard selects the model first, tries Chat/Responses/Anthropic interfaces within the approved allowance, and offers the successful formats. Staged callers can run --stage interfaces first; --api-style is explicit only when choosing a known format. HTTP success with an unreadable response is inconclusive, not support.

docs/model-configuration.md
furgalep: The saved entry owns the runtime settings: endpoint/interface, reply cap, exact reasoning-level request blocks, Responses store/include policy and any explicit cache override. Observations and catalogue identity live under provenance, not request settings. Raw reasoning/signatures/encrypted payloads are never registry configuration. The dedicated model-configuration skill and docs now make that distinction explicit.

docs/model-configuration.md
furgalep: The configuration guide is now separate from the agent procedure: the operator page starts with a concrete YAML example and explains loading/using it; skills/nooa-model-configuration/SKILL.md owns the agent workflow. I also replaced the PR description with a shorter review guide and removed obsolete cap/retry instructions from model-connect.md.

docs/model-connect.md
furgalep: Yes after 8893f15e: configured routing/tools/levels/session checks use the same registry factory and saved max_tokens (or the selected level override), and inspect the serialized request. The initial interface discovery cap is the only smaller cap. A budget shortage now skips the check instead of lowering the limit. Tests compare a saved-and-reloaded entry with Connect's actual HTTP request on all three styles. A saved cap with no accepted wire-verified check is explicitly unverified.

packages/nooa-cli/src/nooa_cli/commands/_connect_stages.py
furgalep: Fixed in 8893f15e. configure_entry now recursively rejects credential-bearing fields, including nested api_key and Authorization, so library, wizard and stage save enforce the same rule. Stage rejection returns safe JSON without echoing the input or writing a file; active keys are also scrubbed throughout reports and prompts. Library and CLI regressions cover top-level and nested credentials.

src/nooa/unifiedllm/connect/__init__.py
furgalep: Fixed in 8893f15e. write() now holds an inter-process flock across the complete read/modify/replace operation, using a stable adjacent .FILENAME.lock inode that is retained (unlinking it would split waiters across different locks). A two-process regression holds the first writer inside its update and verifies the second cannot read until release, then checks both aliases survived. The existing changed-file check remains for non-cooperating edits.

src/nooa/secrets.py
furgalep: This is only for the wizard's optional, separately confirmed Save this key action. It updates the named variable in the existing secrets.yaml, preserves other variables, and atomically writes owner-only permissions; transient keys and --yes never save implicitly. In 8893f15e it also follows symlinks and accepts a scaffold with env: null. The registry still stores only the environment-variable name.

src/nooa/unifiedllm/reasoning.py
furgalep: Reconciled this with main's shared reply-limit resolver in merge 59f8567e rather than retaining the duplicate Connect cap-alias implementation. The rule is simply: a selected level replaces inherited reply limits; combining it with an explicit per-call reply limit is an error. Nested settings are replaced whole, not recursively merged.

src/nooa/unifiedllm/registry.py
furgalep: get_llm_client remains the public alias lookup. client_from_config is the extracted construction half: it accepts a detached, unsaved entry and builds the identical client without registering the candidate globally or writing a file before the user approves. get_llm_client delegates to that same factory after lookup, so Connect does not maintain a second client-construction path.

packages/nooa-cli/pyproject.toml
furgalep: Fixed in 8893f15e: nooa-cli now requires prompt-toolkit>=3.0.52 and uv.lock metadata matches. This only raises the CLI's existing minimum; NOOA core gains no dependency.

src/nooa/secrets.py
furgalep: Fixed the reported abort in 8893f15e: write_secret_env treats env: null as an empty mapping, consistently with the loader. The regression writes a key through that real path and verifies the file contents. Secret persistence still needs separate consent; the two-file save is not claimed to be a cross-file atomic transaction.

src/nooa/unifiedllm/connect/__init__.py
furgalep: Addressed conservatively in 8893f15e: when no request is captured, settings_sent is unknown and the result says the actual request could not be observed because the runtime bypassed the instrumented client. It no longer asserts parameter filtering removed settings. The regression constructs the real keyless client, verifies its owned SDK client is absent, then exercises an accepted fallback response without capture. Full wire verification on that legacy fallback remains unconfirmed; no dummy credential is invented.

src/nooa/unifiedllm/connect/__init__.py
furgalep: Fixed in 8893f15e across the writer, registry reader, both wizard reads and stage-save collision check. Null models mappings are treated as empty while non-mapping values are still rejected. Library and stage-save regressions populate models: successfully.

packages/nooa-cli/src/nooa_cli/commands/_connect_registry.py
furgalep: Fixed in 8893f15e: a higher-layer non-mapping alias removes the lower-layer entry, matching runtime tombstone semantics. A regression verifies the alias is absent both from editor entries and credential suggestions.

src/nooa/unifiedllm/connect/__init__.py
furgalep: Handled safely in 8893f15e, with a deliberate limitation: Connect detects YAML anchors/aliases and refuses the update with instructions to expand them first. It does not flatten the whole file and discard comments. Both reported shapes now produce an actionable ValueError and leave the original bytes intact, rather than StopIteration or a broken splice. Full anchor-preserving editing is not implemented.

src/nooa/unifiedllm/connect/__init__.py
furgalep: Fixed in 8893f15e: accepted routing records establish encrypted_reasoning provenance with setdefault, including the reuse path. A real mocked-HTTP regression removes that provenance before running the saved Responses entry and verifies the accepted result is retained.

src/nooa/unifiedllm/connect/__init__.py
furgalep: Fixed in 8893f15e. refresh_plan builds each request and its reservation from the effective saved/level cap after configuration normalization, so a thinking-budget bump is included in token_estimate. Reasoning-output-tokens no longer silently substitutes a different cap. Wire regressions exercise level-specific caps across all three styles, and stale requests are rejected rather than sent above their reservation.

src/nooa/unifiedllm/reasoning.py
furgalep: Resolved by merging main/#331 in 59f8567e and retaining its shared limits.py resolver. Inherited extra_body cap aliases are promoted/replaced before level application; an explicit per-call cap alias, including inside extra_body, conflicts with a level cap. The old duplicate Connect-only normalization was removed at the conflict. The UnifiedLLM limit/reasoning suites pass in the final 1457-test selection.

src/nooa/unifiedllm/connect/__init__.py
furgalep: Fixed in 8893f15e at the shared configure_entry boundary, including recursive nested fields. Both interactive/library writes and stage save now reject literal credentials before persistence. Error messages name the rule without echoing credential values; stage-save regressions verify no file or secret output.

packages/nooa-cli/src/nooa_cli/commands/connect.py
furgalep: Fixed in 8893f15e: editing only copies original level blocks when no levels file is supplied. The regression adds ultra via the file while the original entry lacks it and verifies the resulting proposal uses the supplied settings.

src/nooa/unifiedllm/replay_state.py
furgalep: Fixed in 8893f15e: empty list, tuple and set values are all explicit opt-outs before native endpoint defaults are added. A parametrized regression verifies include is omitted in all three cases.

packages/nooa-cli/src/nooa_cli/commands/_connect_stages.py
furgalep: Fixed in 8893f15e: discovery reuse compares normalized endpoints with the terminal /v1 removed, matching discovery's own fallback. Missing api_base gets the discovery-specific validation error. Tests accept the original root endpoint and still reject a different host.

src/nooa/unifiedllm/connect/__init__.py
furgalep: Fixed in 8893f15e: resolve symlinks before the update, preserve the existing registry mode and CRLF newlines, and bind the temp path before any write/flush/fsync inside the cleanup try/finally. An injected fsync failure leaves the registry unchanged and no staging file. New registries remain conservatively owner-only; secrets remain deliberately 0600.

src/nooa/unifiedllm/connect/__init__.py
furgalep: Fixed in 8893f15e: actual 422 responses are rejected and the CLI gives a request-settings explanation. A response hook also records the observed HTTP status: malformed HTTP-200 replies that LiteLLM labels 422 are instead not_confirmed (reply not understood), so local parse failure is not confused with provider rejection. Timeout wording retains priority over generic 4xx text.

PR #337, opened 2026-09-14
feat(llm): add opt-in direct provider SDK transports

furgalep:
## Summary

Add opt-in direct OpenAI and Anthropic SDK transports behind UnifiedLLM. LiteLLM remains the default; a direct failure never falls back to it. SDK retries are disabled and NOOA owns the retry policy.

Includes the release-gate work from #324. NOOA Connect, now on main, constructs the configured direct transport and forwards per-call output-cap overrides. Conflicting cap aliases within one call layer remain invalid. The playground preserves main's historical `python_cell` support using canonical Tool objects.

## Stack

`main` → **#337** → #330 → #346 → #347 → #350 (`dev/tui`). Merge from the bottom up. Base: `main`. This PR remains draft.

The insertion also fixes two tracing integration gaps: the trace explorer recognizes provider-neutral `llm.call` spans, and those spans carry the existing LLM viewer plugin hint. Regression tests use actual SDK calls with mocked HTTP and exported spans for both transports.

## Review follow-up

- Cross-library comparison fixes: translate Anthropic `stop` to `stop_sequences`; map context exhaustion to truncation and raise a specific unsupported-stop error for paused/unknown native outcomes. Preserve the native reason in normalized raw response metadata. Direct native OpenAI Chat maps existing `max_tokens` to `max_completion_tokens`; other endpoints keep `max_tokens`. No new model-entry setting or model-name capability guesses: the proposed `chat_max_tokens_field` override has been removed, and LiteLLM's existing translation is unchanged. Conflicting cap fields are rejected. Six targeted paid checks passed on `02dadb17` through compatible endpoints; native OpenAI evidence remains offline-only because native credentials were unavailable.
- Latest test-only delta `dd98637a`: the GLM replay seed now requires computing a scheduling-dependent lookup key before the tool call. Missing readable reasoning produces an explicit assertion instead of AttributeError. Gate fixtures explicitly disable exporters, including local-viewer auto-discovery; application tracing behavior is unchanged. Offline contracts verify the task has a unique optimum and retain the other families' prompts.
- Runtime delta `fde6e141`: preserve readable history as portable text before legacy adapters such as Mistral strip the extension field. OpenAI-compatible routes retain their native field; signed-state adapters are unchanged. Twenty two-turn wire cases cover both transports and sync/async calls, with constructor/request guard regressions and dependency/metric migration notes.
- Restore actual request capture and journal inputs for legacy per-call model/URL/key overrides and handler-backed vLLM, DeepSeek and Gemini routes. Overrides get temporary bound clients; Gemini contents are translated into journal messages. Repeated HTTP attempts do not consume the journal's one-shot input metadata twice.
- Preserve direct HTTP redirect and environment certificate settings without importing LiteLLM. SDK-specific headers are not claimed to be identical.
- Stop converting raw Chat reasoning_content into spoken assistant text in native Anthropic translation. Signed thinking remains native; the shared portable-history projection is unchanged.
- Validate replay_vendor before either transport is constructed. Keep api_style and existing routing-prefix behavior for compatibility with saved entries and replay scopes; document all ten stripped prefixes and literal-name handling explicitly.
- Add real SDK/HTTP tests for multimodal success paths, unknown-field passthrough, structured output, named tool choice, and sync/async 401/429/503/timeout retry behavior. Default-cache registry-to-wire and summarizer negative-control tests now run both transports.

## Release gate

Ten scenarios run on each transport, with separate JUnit identities: three closed-provider resume cases, four readable tool-replay cases, two installed summarizer cases, and one Anthropic-to-OpenAI switch reusing the resume seed. The runner requires exactly 20 cases and rejects skipped, missing, substituted, duplicate, or failed evidence. It rejects NOOA_LLM_TRANSPORT before dispatch, so the environment cannot silently relabel the results.

All routes come from release-gate aliases supplied by the private configuration package. Those aliases cannot set test-specific cache policy; the gate exercises shipped defaults. The bound is 48 capped calls, no retries, 40 minutes; the controller operator must approve the increased budget.

The approved paired run on `fde6e141` passed 18 of 20 cases. Both GLM cases returned a tool call without observable readable reasoning on the trivial seed prompt, so replay could not be validated. This was not a transport-specific failure. The targeted GLM rerun on `dd98637a` passed both cases, with unchanged readable reasoning after resume and no viewer/journal exports. The original and targeted JUnit reports remain separate and private; they are not presented as one complete gate run. Six additional paid checks on `02dadb17` verified Anthropic string/list stop sequences and compatible-Chat cap spelling on both transports, capturing the actual HTTP bodies through UnifiedLLM. Removing the unused cap-field override does not change those default request paths.

## Intentional shared changes / migrations

Opt-in transport does not mean all default-path observables are unchanged. The changelog and [migration notes](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/feat/direct-provider-sdks/docs/direct-provider-sdks.md) explicitly cover:

- Removal of UnifiedLLM.count_tokens and TokenCalibration, with no replacement tokenizer. Actual usage comes from responses; summarization retains its character heuristic.
- Readable reasoning_content replay on compatible Chat routes outside the former allow-list changes second-turn bodies.
- Lazy initialization of LiteLLM globals/cache patch at legacy-client construction; raw LiteLLM calls outside UnifiedLLM are no longer instrumented.
- One llm.call span per logical request, scrubbed wire input and public-response JSON output, changed message/model/usage semantics, and cost-attribute limitations.

Provider formatter removal is deliberately not part of this PR: it is a separate public-API cleanup.

## Current stack validation

Rebased on main `7880c771` (2026-09-17). Remaining merge order: main → #337 → #330 → #346 → #347 → #350 (`dev/tui`). The current #337 transport-review fixes are included throughout the upper stack, as are the previously merged cache-glyph and daemon-job changes.

Independent frozen-tree review verified patch preservation across every layer. The only manual conflict combines two unchanged test fixtures: offline transport-environment isolation and writable settings-directory isolation. Main's UTF-8 trajectory export, monitor decoding, reserved-budget documentation, and SQLite cleanup fixes are retained; shared session-ownership cleanup remains intact.

- Provider, tracing, viewer, ATIF, agent/summarizer, release, native CLI and benchmark suites: **4,797 passed, 4 skipped, 7 deselected**. These ran before the final six-file main update; subsequent storage, session, producer, truncation and configuration suites on the final tree: **352 passed**.
- Whole-tree Ruff lint and formatting pass (1,062 Python files); frozen lockfile check passes.
- Final scripted real-agent ACP stdio smoke: four completed turns across three processes, with restart/replay and cancellation/reuse. No paid provider, Pool, browser, or frontend checks were rerun.


furgalep:
Added a full Anthropic summarizer-fork wire regression in `1547c68e`.

The test invokes the installed budget summarizer through the real middleware chain with both transports. It covers adaptive thinking and high effort, a captured assistant turn containing signed thinking and a tool call, its tool result, the stable cache boundary and changing context, and the actual appended compaction instruction. Eight cases cover implicit/explicit automatic tool choice, system/history cache markers, and plain-text/`return_result` summary replies. Tool replies are read as data and never executed.

Every parsed request field and value matches across transports. Raw HTTP bytes are not identical: the SDKs serialize JSON object keys in different orders. The comparison does not reorder arrays, discard fields, or normalize text/signatures.

Validation: 307 tests passed, 2 skipped across agent, direct-transport, history-contract and tracing-boundary suites; ruff clean. No production code changed and no live calls were made for this change. This closes the tested request-body parity gap; it does not by itself establish live summary quality or explain a prior missing-facts outcome.

furgalep:
Addressed the integrated review at `91df8f98`, keeping the existing client/transport boundaries rather than adding another routing framework. This remains a draft pending fresh review and live evidence.

| Findings | Change / qualification |
| --- | --- |
| 1, 7 | Shared Anthropic route policy for cache markers and dummy tools. `api_style: chat` does not suppress markers for Anthropic behind a gateway; bare native Messages models get the same tool-history rule. |
| 2 | Replacing a registry route clears inherited format/replay metadata and context-window attachment as well as reasoning choices. Transport remains explicit. Cross-format per-call model switches reject before dispatch and request a new client. |
| 3, 4 | Unrecognized slash routes require an explicit endpoint rather than defaulting to OpenAI. Native Azure is explicitly unsupported and rejected, not emulated with the wrong SDK. Compatible gateways retain the explicit `openai/<wire-model>` route. |
| 5 | The existing emergency local path can run without the private alias wheel, with a warning and a missing-evidence note in its draft. Supplying the wheel runs the provider gate. CI still mandates the private wheel and provider gate. No new skip flag. |
| 6 | Playground converts captured Anthropic thinking/tool blocks to portable `LLMResponse` parts. It retains readable text, not journal/redacted signatures. The core raw-native-history guard stays intact. |
| 8 | Explicit registry `context_window` already worked; the actual defect was the fallback metadata shape. Both transports now share the existing model-info fallback. |
| 9, 10 | Unknown legacy resolution no longer fails in the readable-history helper. Unsupported native Bedrock/Vertex opaque scopes are disabled; readable reasoning uses portable projection rather than being silently stripped. |
| 11 | Temporary legacy overrides allocate only the sync or async resources used. The shielded provider operation owns async cleanup, so caller cancellation cannot close an in-flight pool. |
| 12–14 | Reject empty Anthropic conversation turns explicitly. Treat `num_retries=None` and an empty transport environment variable as unset. |
| 15–18 | Omit unknown cost from spans (preserve explicitly reported zero); exclude Gemini history from invocation parameters; expose thinking blocks to the empty-reply check; visibly label and warn when an adapter bypasses wire capture. Arbitrary LiteLLM adapters still do not have guaranteed wire capture—we do not manufacture request evidence. |
| 19, 20 | Replace the exact OpenAI requirement with `>=2.44.0,<3` and retain the tested lock version; correct SDK license notices and remove the obsolete instrumentor notice. No optional-dependency framework added. |
| 21–23 | Isolate offline matrices from ambient soak overrides while retaining live fail-fast guards. Test gate startup behavior rather than a fixture's stubbed value. Exercise request guards through actual dispatch, not just the body builder. |

Regression evidence: the initial focused runs reproduced the routing/cache/default failures, then the cancellation, reasoning-only, telemetry, and Playground failures. The fixed focused runs pass.

Final broad offline selection: **2,552 passed, 2 skipped, 7 deselected**. The two skips are optional Slack-dependent agent imports; live checks are not claimed.

```
LITELLM_LOCAL_MODEL_COST_MAP=True PYTHONPATH=src:packages/nooa-cli/src:. \
UV_PROJECT_ENVIRONMENT=/tmp/nooa-release-provider-venv \
uv run --no-sync --with anthropic --with openai==2.44.0 --with prompt-toolkit --with msgpack pytest \
  tests/unifiedllm tests/tracing tests/viewer tests/trace_explorer \
  tests/test_make_release.py tests/agents tests/integration/test_gate_probe_contract.py -q --tb=short
```

An additional run with ambient `NOOA_LLM_TRANSPORT=direct` passed **139** direct/legacy, review-regression, trace-export and release-gate contract tests. Ruff and `git diff --check` pass.

Connect's requested `transport: direct` default is unchanged. No claim that prior live approvals cover this revision. Fresh offline review and bounded live verification are next.

furgalep:
Simplification follow-up: `30d17b95` (on top of `91df8f98`).

This is a small ownership refactor, not another transport abstraction: net **51 fewer production lines**, no new public settings or dependencies.

- One shielded task owns legacy async dispatch, response collection, and temporary-pool cleanup. Removed the nested cancellation helper and the separate normal-call collection path.
- HTTP pools are the sole resource owners. Removed the SDK-wrapper cleanup list and duplicate closes; the wrappers borrow the same pools.
- Removed unused pool-constructor arguments, duplicate replay-vendor validation, and direct transport's dummy legacy-client attributes. Responses explicitly attaches legacy wrappers only on the legacy path.
- Kept route resolution stateless and all existing route/cache guards intact; no route cache or new routing layer.
- Removed the remaining obsolete `openinference-instrumentation-litellm` notice.

Test-first evidence: new tests reproduced double-closing of OpenAI pools and interrupted response collection on normal-call cancellation. Both now pass. The existing provider-handoff cancellation regression is exercised through dispatch rather than the removed helper. Removing dummy attributes also exposed a Connect test that claimed to exercise legacy fallback while selecting direct; it now explicitly selects and asserts LiteLLM.

Verification (offline, from this checkout):

```sh
LITELLM_LOCAL_MODEL_COST_MAP=True \
PYTHONPATH=src:packages/nooa-cli/src:. \
UV_PROJECT_ENVIRONMENT=/tmp/nooa-release-provider-venv \
uv run --no-sync --with anthropic --with openai==2.44.0 --with prompt-toolkit --with msgpack pytest \
  tests/unifiedllm tests/tracing tests/viewer tests/trace_explorer \
  tests/test_make_release.py tests/agents tests/integration/test_gate_probe_contract.py -q --tb=short
# 2555 passed, 2 skipped, 7 deselected

NOOA_LLM_TRANSPORT=direct LITELLM_LOCAL_MODEL_COST_MAP=True \
PYTHONPATH=src:packages/nooa-cli/src:. \
UV_PROJECT_ENVIRONMENT=/tmp/nooa-release-provider-venv \
uv run --no-sync --with anthropic --with openai==2.44.0 pytest \
  tests/unifiedllm/test_direct_transport.py tests/unifiedllm/test_direct_review_contracts.py \
  tests/unifiedllm/test_review_regressions.py tests/trace_explorer/test_unifiedllm_spans.py \
  tests/test_release_gate_transports.py -q --tb=short
# 139 passed
```

Ruff lint/format checks and `git diff --check` pass. No paid calls in this pass; requesting focused offline re-verification before the held live checks. PR remains draft.

furgalep:
Implemented the three approved simplifications in `7e8968f87716c6688c129899196288a1ff520f6f`, on the rebased PR head `55ad6155`. Native Gemini is explicitly out of scope; no dependencies added.

1. **One context-to-message conversion.** Removed all four provider-formatter classes/exports, `RenderConfig.provider_formatter`, `render_context(provider_formatter=...)`, and the actor's provider selector. `context_blocks.formatter.to_messages` preserves the old canonical conversion, including unchanged `LLMResponse` and `CacheBoundary` carriers. UnifiedLLM alone projects onto provider APIs. Updated callers and retained wire/replay/cache tests; removed tests of the deleted, unused Anthropic renderer.
2. **Cached model identity, not cached request validation.** DirectTransport resolves its constructor model/vendor tuple once and reuses it for the common path. A different per-call model is resolved independently without mutating the cached default. Endpoint/client guards still run on every route lookup, so cached identity cannot bypass native-Azure rejection or the explicit-endpoint requirement. Real SDK wire tests cover sync/async model, endpoint and key overrides followed by a return to the original configuration.
3. **Derived wire format.** Removed the public constructor and registry `api_style` option. ResponsesClient selects Responses; otherwise a leading `anthropic/` selects native Messages and other names select Chat. `openai/anthropic/...` remains a Chat gateway route. Runtime and Connect use the same derivation. Connect retains its interface-selection CLI flag but saves only the model prefix and client type, removing matching old metadata. Conflicting old metadata raises an actionable migration error instead of silently changing protocols. No new routing knob replaces it.

Net production change: **151 added / 280 removed = 129 fewer lines** across src/packages. CHANGELOG and model/direct configuration docs describe the public API removals and migration.

Test-first evidence: the new canonical-conversion/API-surface tests, route-cache tests, Connect saved-shape tests, removed-constructor-option tests and conflicting-metadata tests failed before their implementations and pass afterward.

Final offline verification on the rebased candidate:

```sh
LITELLM_LOCAL_MODEL_COST_MAP=True \
PYTHONPATH=src:packages/nooa-cli/src:packages/nooa-bench/src:. \
UV_PROJECT_ENVIRONMENT=/tmp/nooa-release-provider-venv \
uv run --no-sync --with anthropic --with openai==2.44.0 --with prompt-toolkit --with msgpack pytest \
  tests/unifiedllm tests/context_blocks tests/runtime src/nooa/runtime/tests \
  tests/tracing tests/viewer tests/trace_explorer tests/test_make_release.py tests/agents \
  tests/test_provider_image_rendering.py tests/test_cross_session_tool_call_visibility.py \
  tests/test_event_backend_roundtrip.py tests/integration/test_gate_probe_contract.py \
  tests/integration/test_truncation_e2e.py tests/integration/test_l4_eviction_e2e.py \
  packages/nooa-cli/tests/test_connect_entrypoint.py tests/unit/test_util_and_sqlite.py -q --tb=short
# 4410 passed, 4 skipped, 77 deselected

NOOA_LLM_TRANSPORT=direct LITELLM_LOCAL_MODEL_COST_MAP=True \
PYTHONPATH=src:packages/nooa-cli/src:. \
UV_PROJECT_ENVIRONMENT=/tmp/nooa-release-provider-venv \
uv run --no-sync --with anthropic --with openai==2.44.0 pytest \
  tests/unifiedllm/test_derived_routing.py tests/unifiedllm/test_direct_transport.py \
  tests/unifiedllm/test_direct_review_contracts.py tests/unifiedllm/test_review_regressions.py \
  tests/unifiedllm/test_history_wire_contract.py tests/trace_explorer/test_unifiedllm_spans.py \
  tests/test_release_gate_transports.py tests/context_blocks/test_canonical_messages.py -q --tb=short
# 312 passed
```

Changed Python files pass Ruff lint and format checks; `git diff --check` is clean. Live provider checks remain held pending the combined candidate's review. No new live-pass claim; PR remains draft.

PR #340, opened 2026-09-15
fix(unifiedllm): omit redundant auto tool choice on Chat requests

furgalep:
## Summary

Omit the redundant `tool_choice="auto"` from Chat completion requests while keeping `parallel_tool_calls=False`.

With tools present, omission already means automatic tool selection. Some OpenAI-compatible servers reject the combination of an explicit `"auto"` and `parallel_tool_calls=False`, even though each setting works separately. Removing the redundant default avoids that conflict without permitting parallel tool calls.

The change is in both `CompletionClient.call` and `acall`, with no provider-name special cases. It covers constructor settings and per-call arguments. Explicit `"required"`, `"none"`, and named-tool choices remain unchanged, as do the tools, reply limit, and caller-owned configuration. Responses requests are unchanged. Raw fields explicitly supplied through `extra_body` are not rewritten.

## Validation

- Test-driven regression: before the fix, four explicit-auto cases fail with the simulated HTTP 400; the other sixteen pass. After the fix, all twenty pass.
- Tests capture requests after LiteLLM/OpenAI SDK serialization through a mock HTTP transport, with real network access forbidden. Both sync and async paths and constructor/per-call configuration are covered.
- `tests/unifiedllm tests/strategies tests/agents`: 2,034 passed, 2 skipped, 5 deselected. Ruff check and format check are clean.
- Bounded live comparisons established the interaction; route details and live measurements remain in private evidence, not this PR.

API reference: [Chat tool choice](https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved tool-use requests by omitting the automatic tool-choice setting when tools are provided.
  - Explicit choices such as required, none, or a named tool continue to be sent unchanged.
  - Requests without tools retain their existing behavior, and parallel tool calls remain disabled by default.

- **Tests**
  - Added coverage for synchronous and asynchronous requests, including client-level and per-call configuration.


sklinglernv (approved):
LGTM

PR #341, opened 2026-09-15
fix(unifiedllm): enable stable-prefix caching by default

furgalep:
## Summary

Honor the default cached renderer's stable-prefix boundary without requiring a
cache setting in each model registry entry.

- `ResponsesClient` now defaults to `cache_breakpoint="auto"`. A rendered boundary
  plus eligible stable input enables explicit cache mode and places its marker.
  Without a boundary or eligible prefix, it adds no cache fields.
- `CompletionClient` retains its automatic Anthropic mapping and provider-default
  caching for other Chat routes.
- Explicit `"openai"`/`"anthropic"` mappings retain their behavior. `None` disables
  NOOA-generated markers, not automatic caching performed by a provider. It is
  also the opt-out for Responses endpoints that reject explicit-cache fields.
- Live resume and summarizer tests no longer pass cache overrides, so they test
  the shipped defaults. Update the configuration docs and agent-authoring skill.

The regression was that a normal Responses registry entry silently consumed the
renderer boundary without sending cache mode or a marker. Existing cache tests
explicitly enabled the policy and therefore missed the default-path failure.

## Validation

- New registry -> CachedBlockFormatter -> UnifiedLLM -> actual SDK HTTP body tests,
  sync and async, for Chat, Responses and Anthropic. They cover default behavior,
  no boundary, opt-out, and no eligible prefix, plus input non-mutation.
- Before the fix: 2 failing default Responses cases, 18 passing controls.
- After the fix: 20 new tests pass; affected suites: 1,112 passed, 2 skipped,
  5 deselected. Ruff and `git diff --check` pass.
- No paid provider calls made for this change. Live default-path validation is
  still required; a configured policy does not guarantee a cache hit.

Targets `main`. Separate from the synthetic-prefill reasoning fix. The pending
release-alias and direct-SDK PRs must carry these defaults forward when rebased;
they are not prerequisites for this default Responses-client fix. No redundant
cache fields were present in public main's shipped YAML entries.

Wren review requested on frozen head `8bccb113e3ca8d684b4a1c3f852473680e514308`.

## Summary by CodeRabbit

- **New Features**
  - Automatic cache-boundary handling is now the default for client caching.
  - Responses clients support automatic, explicit OpenAI, or disabled cache-marker modes.
  - Explicit caching is applied only when eligible stable input is available; otherwise provider-default caching is preserved.
  - Cache boundaries can improve reuse for stable instructions and prompt prefixes.

- **Documentation**
  - Updated caching guidance to explain automatic behavior, provider-specific handling, opt-out configuration, and fallback behavior.


furgalep:
Second review of head 8bccb113 (Wren).

**Verdict: accept.** The change does what Paul decided on 2026-09-15: caching is on by default in both clients, a registry entry needs no cache field, and `None` is the only opt-out.

What I verified independently:

- Read the full diff. The new `auto` branch in `apply_cache_policy` is small and explicit: no boundary means no added fields; a boundary with an eligible stable block means explicit mode plus one marker; a boundary with nothing eligible falls back to provider-default caching without a warning. Forced `"openai"` keeps its previous behaviour, including the leading-instructions fallback and the warning. The Chat client's automatic Anthropic mapping is unchanged.
- Ran the four touched test files in this worktree: 174 passed. Ran `tests/unifiedllm`, `tests/context_blocks`, `tests/integration` (offline set) and `tests/tracing`: 1504 passed. Ruff clean.
- The new wire test is the test that was missing. It builds the client through `get_llm_client` from an entry with no cache field, renders through the default `CachedBlockFormatter`, and asserts on the real HTTP body, sync and async, across Chat, Responses and Anthropic, with the no-boundary and opt-out controls. It fails on main before the fix (2 default Responses cases), as described.
- Reproduced the original failure offline against the real `ExperimentalTUIAgent` and confirmed the marker lands on the last stable user item before the dynamic `<context>` block once explicit mode is enabled, so this default is the right fix for the TUI case.
- Description carries no gateway hosts, routes or key names. All CI checks green on this head.

Notes, none blocking:

1. The live tests now exercise the shipped default. The private release-gate aliases contain no cache field, so a gate run after merge validates the default path end to end. Recommend running it before the release rather than as a separate approval.
2. #324 and #337 must carry the new `ResponsesClient` default and the `"auto"` literal through their rebases; #337's `DirectTransport` should need no change since it forwards unknown fields.
3. The design page (ordered-turn design, docscn) is updated to describe the new default.

sklinglernv:
> No paid provider calls made for this change. Live default-path validation is still required; a configured policy does not guarantee a cache hit.

I think to validate this fix we need some end2end tests with actual provider calls, otherwise it is too risky to decide if caching truly works now, no?

sklinglernv:
[AI Take] Automated review findings (Claude Code)

**1. `util/eval_pipeline/src/eval_pipeline/config.py:481` — eval_pipeline's `no_cache` toggle doesn't cover the new default (CONFIRMED)**

`no_cache` (meant to preserve pass@k sampling diversity, per the code comment "Disable NVIDIA inference API caching (important for pass@k diversity)") only sets `extra_body["cache"] = {"no-cache": True}`. It never sets `cache_breakpoint` on the `ResponsesClient` it builds (`config.py:481`, and `agents.py:71` in the arbitrary-agent-loading path), so those clients now pick up this PR's new `"auto"` default instead of the old `None`.

*Failure scenario:* an agent loaded via `agent_from_spec` that renders context through NOOA's standard `CacheBoundary`/`CachedBlockFormatter` path gets OpenAI explicit prompt caching silently enabled even with `no_cache=True`/`--no-cache` set, since that's a separate caching mechanism the pipeline's diversity-preserving knob was never designed to cover.

**2. `src/nooa/unifiedllm/unifiedllm.py:1260` vs `cache_policy.py` — two independent `"auto"` resolution strategies (PLAUSIBLE)**

`_prepare_cache_boundary` eagerly resolves `"auto"` to `"anthropic"`/`None` by model name before calling `apply_cache_policy` (only when `responses=False`), while `apply_cache_policy` separately re-derives auto behavior for the Responses path via its own `automatic = mapping == "auto"` flag keyed on boundary presence. Two different mental models for the same concept, split across two files with no shared resolution function.

*Risk:* a future change to "auto" semantics (e.g. the no-eligible-boundary fallback) is likely to be applied to only one of the two paths, leaving CompletionClient and ResponsesClient with inconsistent auto behavior.

**3. `src/nooa/unifiedllm/cache_policy.py:174` — `automatic` flag threaded through existing explicit-mode logic (PLAUSIBLE)**

The new `automatic` flag adds duplicate no-op early returns (lines 139-140 / 142-143), a `not automatic` check evaluated twice for unrelated purposes (lines 166 and 174), and an opaque final `return clean, instructions, marked or not automatic`, instead of splitting auto vs. explicit into clear branches.

*Risk:* a future contributor editing one `not automatic` check (e.g. the warning condition) without realizing the return-value expression encodes a related but distinct policy could silently change whether explicit caching is reported as enabled on a cache-miss.

**4. `tests/integration/test_summarizer_live.py:198` — cache assertion now depends on external registry config (PLAUSIBLE, low confidence)**

Removing the test's explicit `cache_breakpoint=family` kwarg means the effective cache-breakpoint for the openai-family live test now depends entirely on an external, non-repo registry YAML entry (`release-gate-openai`) rather than being pinned by the test itself.

*Risk:* if that external alias sets `cache_breakpoint` to something other than `"openai"`/`"auto"`, the live test's `assert report["calls"][1]["cached_input_tokens"] > 0` could start failing (or pass for the wrong reason) with no diff-visible cause in this repo. Worth a quick check against the registry config before merge.

---
No other correctness issues found across 8 parallel review passes (correctness, reuse, efficiency, simplification, altitude, conventions); the core cache-default logic (boundary detection, warning suppression, provider-mapping validation) checks out against the new test suite.

sklinglernv (approved):
It would be great to test it on live endpoints to make sure it works. Other than a few AI comments. But looks good to me otherwise!

furgalep:
Replies to the four automated findings (Wren):

1. **eval_pipeline `no_cache`**: not a defect. That flag sets `cache: {"no-cache": true}`, which is the gateway's *response* cache bypass (LiteLLM proxy), so repeated identical prompts are not answered from a stored reply. Prompt caching is a different mechanism: it reuses the computed prefix and does not change sampling, so pass@k diversity is unaffected. Explicit prompt caching engaging under `--no-cache` is the desired outcome. A one-line comment next to the flag would remove the ambiguity; no code change needed.
2. and 3. **Two `auto` resolutions / `automatic` flag**: fair style points, not correctness. The Chat path resolves `auto` by route because its mapping is provider-specific; the Responses path resolves it by boundary presence because its mapping is not. The new wire test pins both. A follow-up that splits `apply_cache_policy` into an auto branch and a forced branch would read better; it should not hold this fix.
4. **Live test depends on the external alias**: checked. The private `release-gate-*` entries carry no cache field, so the live tests now run the shipped default, which is the point of removing the override.

On live validation: agreed, and it is the remaining step. The offline suite proves the wire body; the TUI road test earlier today showed the same explicit-mode path reading about 98 percent of a 320k-token prompt from cache, but that run had the field set by hand. A run of `test_cache_resume_live` and `test_summarizer_live` for the openai and anthropic families through the release-gate aliases on this head validates the automatic default itself. Pending Paul's budget approval; I can run it.

PR #342, opened 2026-09-15
fix(tracing): bound JSON serialization work

sklinglernv:
### Motivation

We observed a single deterministic agent method taking more than **50 seconds** to execute solely because tracing synchronously serialized its input parameters before the method could run. The production-shaped reproduction measured **57.2 seconds inside `before_agent_call`** for 5,000 repeated Pydantic candidate objects, even though the retained trace attribute was only about 6.5 K characters.

The expensive part was not ordinary `json.dumps()` by itself. The old fallback rendered every custom/Pydantic object while walking the complete graph, including values that could never fit in the trace. Gaia's Experimentalist-specific compaction avoided this one incident shape, but encoded domain knowledge in an application plugin and left the general tracing problem unresolved.

### Final design

- Build a bounded JSON-compatible preview tree first, then let standard-library `json.dumps()` encode that small tree.
- Bound all three relevant dimensions: encoded characters (50,000), inspected nodes (2,000), and nesting depth (16).
- Preserve exact built-in JSON values and a narrow, stored-field Pydantic view for normal small objects.
- Honor Pydantic `exclude=True`, `repr=False`, and NeMo visibility metadata without running serializers, computed fields, `repr`, `str`, iteration hooks, attribute hooks, or metaclass descriptors.
- Treat unsupported/custom objects as cheap opaque values.
- Keep the payload ordinary JSON. Truncation is reported out of band with `nooa.{input|output}.preview.*` span attributes, so user data can never collide with a magic envelope.
- Apply the same codec to every framework trace input and output; tool-call arguments remain byte-identical to `input.value`.
- Decode I/O once according to its MIME type in both trace consumers and show an incomplete-preview notice when applicable.

The previous `LimitedWriter`, custom JSON fragment encoder, and `$nooa` envelope are removed.

### Results

Final measurements on the same production-shaped cases:

| Case | Bounded codec | Full tracing hook | Trace value |
|---|---:|---:|---:|
| 1.03 MiB expanded native JSON | 0.0008 s | 0.0009 s | 50,000 chars |
| 5,000 repeated Pydantic candidates | 0.0003 s | 0.0004 s | 50,000 chars |

Isolated 1, 16, 64, and 256 MiB string arguments each completed in about **0.3 ms** after source allocation, with no measurable serialization RSS growth. The former 57.2-second Pydantic case now completes through the full hook in about **0.4 ms**.

### Verification

- 52 focused tracing, explorer, privacy, viewer-plugin, and wire-format tests pass.
- Pre-commit checks pass, including Ruff, Ruff format, Pyright, SPDX, whitespace, and merge-conflict checks.
- Frontend TypeScript and production bundle build successfully.
- Independent Astra implementation review approved after adversarial tests for hostile descriptors, equality/hash hooks, malformed Pydantic metadata, cycles, escaping, and exact size accounting.

The broader async explorer/conformance batches were also exercised, but their processes stalled during existing fixture teardown in this environment; completed assertions before teardown passed.

## Related issues

No public issue.

## Summary by CodeRabbit

- **New Features**
  - Added bounded trace previews that remain valid JSON and clearly identify incomplete content.
  - Added support for MIME-aware decoding of structured inputs and outputs.
  - Added visible warnings when trace previews are incomplete.
  - Improved handling of large values, nested data, cycles, and unsupported objects while protecting sensitive fields.

- **Bug Fixes**
  - Preserved structured method arguments, code, and results when previews are truncated.
  - Prevented incorrect parsing of non-JSON and already-decoded values.

- **Documentation**
  - Added design documentation and benchmarks covering preview behavior and serialization performance.


furgalep (changes requested):
The production-shaped improvement is compelling, and exception-driven early abort is a good way to stop traversal between container elements. I verified the focused Python suite (30 passed), Ruff checks, a reduced reproduction benchmark, and the frontend production build.

I am requesting changes for two architectural reasons:

1. **The limit is below the wrong boundary for scalar values.** `LimitedWriter` sees a string only after CPython's stdlib JSON encoder has fully escaped/materialized that scalar. Retained output is bounded, but CPU and memory are not bounded for one huge string. With a 50,000-character limit, I measured a 10M-character quote-only string still taking 0.0386s and 19.4 MiB peak traced allocation before producing a 49,999-character envelope. This fixes repeated-object/container traversal but does not establish the general work/memory bound claimed by the PR. Please add scalar-aware bounded encoding and a regression test with one large escape-heavy native string.

2. **Overflow changes the semantic application-JSON shape.** `{args, kwargs}` / `{code}` / tool arguments become a proprietary `{$nooa, preview}` object. The updated method/code viewers understand it, but generic OpenInference consumers and internal paths such as the `return_result` preview (`src/nooa/trace_explorer/explorer.py:3062-3067`) still expect the original fields. The envelope is also copied into `tool_call.function.arguments` (`src/nooa/tracing/_hooks_impl.py:1122-1133`) as though it were the original argument object.

Please either preserve known outer semantic shapes with explicit truncation markers, or version and document the representation and centralize encoding/decoding for every consumer. Prefer out-of-band span attributes such as `nooa.serialization.truncated`, `nooa.serialization.version`, and `nooa.serialization.limit_chars` for reliable feature detection.

Suggested architecture: one shared bounded trace-JSON codec owning container traversal, scalar encoding, truncation metadata, and compatibility fixtures. Tracing hooks call it; consumers should not independently infer transport policy from magic payload keys.

Separately, the same bounded-output-versus-bounded-work problem exists in generation-method input display (`InspectInputsPrefill` / `pprint`), especially for pandas, NumPy, and arbitrary custom reprs. I consider that follow-up scope rather than a requirement to expand this tracing PR.

furgalep:
TL;DR, let's fix so that
- it works for all values we can think of.
- the output shape is always the same
- [stretch] it raises the possibility of reusing what we build here for pprint() or prefill

PR #343, opened 2026-09-15
fix(validator): pass cell filename to ast.parse so syntax errors name the cell (fix #267)

### LINKED ISSUE #267: Syntax errors reach the agent without cell location: validator parses with the default filename
### Summary

When a generated cell fails to parse, the agent receives:

    RestrictedCodeError: Syntax error: '(' was never closed (<unknown>, line 1)

A runtime failure in the same cell receives full source context:

      Cell In[1], line 4, in <module>
        return None.strip()
               ^^^^^^^^^^

`<unknown>` tells the model nothing about which cell or where to look, which weakens
the error-recovery loop precisely when the model most needs to locate its own mistake.

### Environment

- Commit `e137e1b`, version `0.0.10.dev137`
- Python 3.12.3, Ubuntu (WSL2)

### Cause

`UnifiedCodeValidator.validate` parses with the default filename
(`code_validator.py:1452`):

    tree = ast.parse(code)

The execution path does it correctly a few lines later (`actor.py:1565-1570`):

    cell_filename = f"Cell In[{execution_count}]"
    tree = ast.parse(code, filename=cell_filename)

But validation runs first (`actor.py:1464`) and raises on failure, so a syntax error
never reaches the well-formed parse. `ValidationContext` already carries
`execution_count`, so the filename is derivable at the validator's parse site.

### Reproduction

    import ast
    code = "a = 1\nb = 2\nc = (3 + 4\nd = 5\n"

    for fn in (None, "Cell In[7]"):
        try:
            ast.parse(code) if fn is None else ast.parse(code, filename=fn)
        except SyntaxError as e:
            print(f"filename={e.filename!r:14} lineno={e.lineno} offset={e.offset} text={e.text!r}")

Output:

    filename='<unknown>'    lineno=3 offset=5 text='c = (3 + 4'
    filename='Cell In[7]'   lineno=3 offset=5 text='c = (3 + 4'

`lineno`, `offset` and `text` are correct in both cases. Only the filename differs, and
the caret column is available but not rendered.

### Two directions, and I'd rather follow your preference than guess

1. Pass `filename=f"Cell In[{context.execution_count}]"` at the validator's parse site.
   One argument; gives the model the cell and line. No source line beneath the header,
   because `linecache` is not populated until `actor.py:1667`.
2. Also register the cell in `linecache` before validation, so the syntax error renders
   with source and caret like a runtime error. Wider change, but the `offset` and `text`
   needed for a caret are already on the exception.

Happy to take this either way.

### Context

Found while writing the "SyntaxError source/caret fidelity" conformance row for #187
(#226). Both execution backends produce identical output here, so this is not a parity
gap — the contract that row names does not currently exist at the CodeAct level. I will
note that separately on #187.

Related to #190, but distinct: that issue covers the formatter's own boundaries,
including not mutating caller-owned `SyntaxError` instances. This is upstream of the
formatter — validation raises first, so a syntax error never reaches it. If you would
rather fold the fix into the #190 cleanup, that seems reasonable; filing separately
because the cause and the file are different.

inaor:
## Problem

Fixes #267.

`UnifiedCodeValidator.validate()` called `ast.parse(code)` without a `filename` argument. Any `SyntaxError` raised during validation therefore carried `filename='<unknown>'`, and the agent received:

```
RestrictedCodeError: Syntax error: '(' was never closed (<unknown>, line 1)
```

Runtime errors in the same cell already show `Cell In[N]` because `actor.py` passes `filename=cell_filename` to its own `ast.parse` call. The mismatch means syntax errors — which arrive at validation time, *before* the well-formed parse — give the model no usable location hint at exactly the moment it most needs one to self-correct.

## Fix

`ValidationContext` already carries `execution_count`. Derive `cell_filename` from it and pass it to `ast.parse`:

```python
cell_filename = f"Cell In[{context.execution_count}]"
tree = ast.parse(code, filename=cell_filename)
```

The `SyntaxError` now carries the correct filename, so the message the agent receives becomes:

```
RestrictedCodeError: Syntax error: '(' was never closed (Cell In[7], line 3)
```

One-line change; no interface impact.

## Tests

Two regression tests added to `TestValidatorIntegration`:

- `test_syntax_error_includes_cell_filename` — verifies `Cell In[7]` appears and `<unknown>` does not, using the unclosed-paren example from the issue report.
- `test_syntax_error_cell_filename_uses_execution_count` — checks the cell number tracks `execution_count` across three values (1, 5, 42).

```bash
uv run pytest tests/runtime/test_code_validator.py -q   # 228 passed
uv run ruff check src/nooa/runtime/code_validator.py tests/runtime/test_code_validator.py
uv run pyright src/nooa/runtime/code_validator.py       # 0 errors
```

---

*Security Joes — contributing to the Open Secure AI Alliance via NOOA.*

## Summary by CodeRabbit

* **Bug Fixes**
  * Syntax validation errors now identify the relevant execution cell, including its execution count, instead of displaying an unknown source location.

* **Tests**
  * Added coverage to verify accurate cell references in syntax-validation errors.


furgalep:
The functional change looks correct and appropriately narrow. I verified the exact head with 228 validator tests passing, Ruff check passing, Pyright reporting 0 errors, and a manual reproduction confirming that `SyntaxError.filename`, line number, and source text are preserved with `Cell In[7]`.

One cleanup before merge: the committed test file is not Ruff-formatted at this head. Running:

```bash
uv run ruff format --check src/nooa/runtime/code_validator.py tests/runtime/test_code_validator.py
```

reports `Would reformat: tests/runtime/test_code_validator.py`. The only change is collapsing the new `test_syntax_error_cell_filename_uses_execution_count` signature to one line.

Please run:

```bash
uv run ruff format tests/runtime/test_code_validator.py
```

After that formatting update, this looks ready to approve, subject to the normal CI matrix completing.

Issue #110, opened 2026-08-06
ShellTools.run() cannot start on Windows: /bin/bash is hardcoded as the executable 🤖🤖🤖


IgorKasianenko:
**Describe the bug**

`BashSession.start()` passes the literal string `"/bin/bash"` to `asyncio.create_subprocess_exec` as the executable ([`src/nooa/tools/_bash_session.py:164`](https://github.com/NVIDIA-NeMo/labs-OO-Agents/blob/main/src/nooa/tools/_bash_session.py#L164)). That path does not exist on native Windows, so `ShellTools.run()` cannot start a session there.

This is **not** a regression and not a new break — it has never worked on Windows. It is filed because it is currently unreported: #95 mentions the `pass_fds` failure in passing, and its own body notes that its STEP 6/7 findings are not covered by any open issue. This is STEP 6's *next* failure.

**Relationship to #98**

#98 fixes the failure that currently masks this one. On `main`, `start()` fails first at:

```
AssertionError: pass_fds not supported on Windows
```

because `start_new_session=True` and `pass_fds=(ctrl_w,)` are passed unconditionally (lines 172-173). #98 correctly moves both behind a `sys.platform != "win32"` guard:

```python
kwargs: dict[str, Any] = {}
if sys.platform != "win32":
    kwargs["start_new_session"] = True
    kwargs["pass_fds"] = (ctrl_w,)
```

That guard works — verified on a native Windows 11 host against #98's branch. But the diff leaves the `"/bin/bash"` argument unchanged, so the call now proceeds one step further and fails with:

```
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

So #98 changes *which* error Windows users get, not whether `ShellTools.run()` works. Worth stating explicitly so #98 is not read as making the shell surface Windows-ready.

**A second obstacle behind this one**

Even with a `bash` on PATH (Git Bash, MSYS2), dropping `pass_fds` on Windows removes the control-channel descriptor the protocol depends on — `ctrl_w` is fd 3 inside bash. Making `ShellTools.run()` genuinely work on Windows is therefore a larger design question than substituting an executable path, which is why this issue proposes the smaller thing below.

**To reproduce**

On a native Windows host (no WSL), with #98's branch checked out:

```python
import asyncio
from nooa.tools import ShellTools

asyncio.run(ShellTools().run("echo hi"))
```

On `main` the same snippet fails earlier, with the `pass_fds` AssertionError.

[...]

PR #351, opened 2026-09-16
fix(tools): raise a clear error from BashSession.start() on native Windows 🤖🤖🤖

Ashfaqbs:
## Summary

`ShellTools.run()` cannot start a session on native Windows — `/bin/bash` is hardcoded, and the fd-3 control channel is POSIX-only. This has never worked on Windows (same family of platform assumptions as #84, #85, #92, #95), but the failure that reaches the caller is a confusing low-level one (`AssertionError` on `pass_fds` today; `FileNotFoundError` once #98 lands, per #110's own investigation).

This PR implements option 1 proposed in #110: fail with a clear, explicit, actionable error instead.

## Changes

- `BashSession.start()` now checks `sys.platform == "win32"` at the very top, before any subprocess/pipe setup, and raises `NemoOOAgentsRuntimeError` with a message that names the actual constraint (POSIX shell + fd-3 control channel) and points to WSL/Linux/macOS as the fix.
- Placed before `os.pipe()` is opened, so the guard fires identically whether or not #98 (which changes `pass_fds`/`start_new_session` handling) has merged — no dependency between the two PRs.
- Added `TestBashSessionWindowsGuard` in `tests/tools/test_bash_session.py`, covering: the error is raised with the expected message, and session state stays clean (`_started`/`_process` untouched) so a caller isn't left with a half-initialized session.

## Test plan

- `uv run pytest tests/tools/test_bash_session.py -v` — new `TestBashSessionWindowsGuard` tests pass.
- `uv run ruff check` / `uv run ruff format --check` on both changed files — clean.
- Verified the guard fires deterministically by reading current `main`'s `_bash_session.py`, confirming `pass_fds`/`start_new_session` are still unconditional there (i.e. #98 has not merged), so this fix does not assume any particular merge order relative to #98.

Out of scope (per #110's own framing): making `ShellTools.run()` actually work on Windows is a larger design question (the fd-3 control-channel protocol has no Windows equivalent) — this PR only replaces the current confusing crash with an actionable one.

Implementation assisted by my AI coding agent, changes reviewed by me.

🤖🤖🤖

## Summary by CodeRabbit

* **Bug Fixes**
  * Bash sessions now clearly report that POSIX shell support requires WSL, Linux, or macOS when started on native Windows.
  * Failed startup attempts leave the session uninitialized and avoid resource leaks.
  * Repeated startup attempts on unsupported Windows environments consistently return the same clear error without creating session resources.


Ashfaqbs:
Added the missing docstring on `test_start_raises_clear_error_on_win32` to satisfy the docstring-coverage gate (was 75%, needed 80%).

PR #353, opened 2026-09-16
refactor(context): introduce composable context views

sklinglernv:
## Summary

This PR makes context assembly a public, composable framework seam. Agents and skills expose a `ContextView` that yields an ordered stream of materialized blocks, typed events, and explicit cache boundaries.

- Add one `ContextView[Owner]` contract for agents, skills, and custom context systems.
- Resolve agent views by call, method, instance, class, then `DefaultAgentView`; resolve skill views by instance, class, then `DefaultSkillView`.
- Move the complete default assembly policy into a readable `DefaultAgentView`, including source materialization, overrides, ordering, event selection, cache placement, and budgeting.
- Let custom views use `self.context`, another state API, or no context manager at all.
- Keep rendering and provider formatting order-preserving and content-neutral; unsupported provider layouts fail explicitly.
- Replace implicit cache injection with typed `CacheBoundary` markers owned by the selected view.

The concise design and ownership model are documented in [`docs/context.md`](docs/context.md).

## Compatibility

The default view preserves the existing public context APIs and prompt behavior: `agent.context`, context and event managers, strategy/scoped overrides, skill activation, and `Skill.context_block` remain supported. Existing state is translated into the new ordered representation by explicit helper functions.

Custom views can replace that policy completely while the runtime remains responsible only for call facts, collection, rendering, transport, and observability.

## Validation

- Full test suite: 7,948 passed, 7 skipped, 3 expected failures.
- Deterministic prompt-parity comparison: 6 scenarios and 7 requests passed.
- Live CodeAct and Predict flows passed against the NVIDIA inference API with GPT 5.6 Terra; rendered prompts and traces were inspected.
- Focused context, replay, caching, provider-layout, and custom-view regression suites passed.
- Current GitHub CI passes lint, tests, build, Python 3.13 compatibility, macOS sandbox IPC, and secret scanning.

## Summary by CodeRabbit

* **New Features**
  * Added customizable context views for agents and skills, including per-call overrides.
  * Added structured context blocks, cache boundaries, event selection, ordering, and budget controls.
  * Added a custom research-context quickstart example with source citations.

* **Improvements**
  * Context rendering now preserves assembled ordering and tool-call groups.
  * Provider-specific caching now requires an explicit cache boundary.
  * Added validation for unsupported context layouts and invalid message ordering.

* **Documentation**
  * Updated caching and context-management guidance.


examples/quickstart/16_custom_context.py
```diff
@@ -0,0 +1,102 @@
...
+class ResearchAnswer(BaseModel):
+    answer: str
+    sources: list[str]
+
+
+class ResearchAgent(
+    Agent,
+    llm=llm,
+    context_view=ResearchContextView(),
+    event_query=EventQuery.current_call(),
```
furgalep, line 78: Shouldn't we remove event query? Wouldn't you do the event query as part of the context view?

furgalep (approved):
I like it a lot. 

I think my only substantial comment is that we should consider removing the event query configuration as well and just fold that into this new context view. 

Once it lands and it soaks for a little bit, we should consider a follow-on PR where we greatly simplify the default.

alessiodevoto (commented):
Three compatibility regressions reproduced on this head and checked against base commit bb1f948f, where the same cases work. Validation also covered 2,527 passing existing context and runtime/strategy tests (two skipped).

src/nooa/default_context_view.py
```diff
@@ -0,0 +1,494 @@
...
+
+def _source_error(key: str, exc: Exception) -> str:
+    logger.warning("Context source %r failed to resolve: %s: %s", key, type(exc).__name__, exc)
+    return f"{type(exc).__name__}: {exc}"
+
+
+async def system_prompt_block(agent: "Agent", call: "CurrentCall") -> Block | None:
+    """Build the default system prompt directly from the agent."""
+    try:
+        content = resolve_agent_system_prompt(agent)
```
alessiodevoto, line 149: **[P2] Preserve subclass system-prompt overrides**

Calling the module helper directly bypasses an agent subclass's `_resolve_system_prompt()` override. I reproduced this with a subclass whose docstring is `Class prompt.` and whose `_resolve_system_prompt()` returns `CUSTOM PROMPT`: the base branch renders `CUSTOM PROMPT`, while this head renders `Class prompt.`. Existing agents using this override silently lose their application instructions, despite the compatibility claim.

Please preserve dispatch through `agent._resolve_system_prompt()` (including the existing async expression behavior), or explicitly treat this as a breaking change with a migration path. The emitted metadata still advertises `self._resolve_system_prompt()` even though that method is no longer called.

src/nooa/skill.py
```diff
@@ -13,6 +14,19 @@
 
 from nooa.agentdoc import hidden
 
+_MISSING = object()
+
+
+class SkillMeta(ABCMeta):
```
alessiodevoto, line 20: **[P2] Keep the Skill metaclass compatible with AgentMeta**

Introducing a separate `SkillMeta(ABCMeta)` breaks previously valid classes that combine agents and skills. This minimal declaration succeeds on the base commit but fails at class creation on this head:

```python
class AgentSkill(Agent, Skill, llm=object()):
    pass
```

It raises `TypeError: metaclass conflict` because `AgentMeta` and `SkillMeta` are unrelated subclasses of `ABCMeta`. Such skill modules can no longer import. Please use compatible metaclasses or consume the instance `context_view` without introducing this conflict.

src/nooa/runtime/context_manager.py
```diff
@@ -457,6 +469,7 @@ def remove_protected(self, key: str) -> None:
             raise KeyError(key)
         del self._blocks[key]
         self.protected_keys.discard(key)
+        self._protected_expressions.pop(key, None)
```
alessiodevoto, line 472: **[P2] Allow removing protected reservations without a stored declaration**

The new `protect()` registration stores default framework keys in `protected_keys`, but not in `_blocks`. This method still raises at the `_blocks` membership check above, so `agent.context_manager.remove_protected("state")` now raises `KeyError('state')` on a freshly constructed agent; the same call succeeds on the base commit. The new cleanup here is therefore unreachable for untouched framework defaults.

Please recognize protected reservations as existing keys, remove the declaration only if present, and clear the protection, expression, placement, disabled state, and resolved cache. Add a regression test using a fresh agent rather than a key installed through `set_*_protected()`.

docs/context.md
```diff
@@ -0,0 +1,178 @@
+# Context Management in NOOA
+
+## Goal
```
alessiodevoto, line 3: This is a doc, maybe it should not start with "goal" ?

alessiodevoto:
Thanks @sklinglernv this is very cool! 👏  Left some inline comments all P2. The only things I am not convinced of is the docs and quickstart. Maybe the docs could be a bit more organic and introduce the whole idea of "context is a view/function that takes any strcuture as input and projects the output into the llm's current turn" or something along this line. For the quickstart, a few inline comments would help follow the story more easily (e.g., what is this agent about and what we want to do with it, else one has to figure out while reading the code).

PR #359, opened 2026-09-17
feat(acp): populate PromptResponse.usage with real per-turn token accounting

furgalep:
## Summary
- ACP's `PromptResponse.usage` was always unset, so clients had no visibility into per-turn token usage.
- `ACPEventBridge` now accumulates real per-call usage (input, output, reasoning/thought, cached-read, cached-write, total tokens) from `LLMResponse` events across a turn.
- A new `take_turn_usage()` method reads and resets that accumulator into an `acp.schema.Usage` object; `server.py::prompt()` attaches it to every `PromptResponse` it constructs. Turns with no LLM call (e.g. a rejected slash command) correctly report `usage=None`.

## Test plan
- [x] `packages/nooa-acp/tests/test_event_bridge.py` — new tests for accumulation, field mapping, reset-between-turns, and the no-call case.
- [x] `packages/nooa-acp/tests/test_server.py` — usage populated after a normal completion turn; `usage is None` on no-LLM-call slash paths.
- [x] Full `nooa-acp` test suite passes (two pre-existing, unrelated console-script-install failures reproduce identically on an unmodified checkout).
- [x] `ruff check` clean on all changed files.

## Summary by CodeRabbit

* **New Features**
  * Added token-usage reporting to prompt responses.
  * Reports input, output, reasoning, cached, and total token counts across all model calls in a turn.
  * Usage resets between turns and is omitted when no model call occurs.
  * Background summarization now reports usage to observers without adding responses to the conversation history or model context.
* **Bug Fixes**
  * Ensured usage information is consistently included across prompt outcomes, including errors, cancellations, and generation limits.
  * Prevented failed turns from carrying token usage into subsequent prompts.


furgalep:
Ran `/code-review high` on this PR and it found two real bugs, both now fixed in the latest commit:

1. **Token leak across turns.** `usage=session.bridge.take_turn_usage()` was only called at each of the 7 explicit `PromptResponse` return sites in `prompt()`. A turn that ended via an uncaught exception (e.g. a `GenerationError` whose message matches neither hard-coded prefix) skipped all of them, leaving that turn's tokens sitting in the bridge's accumulator to get summed into whichever turn read it next. Confirmed empirically: a failed turn followed by a real 15-token turn reported 1009+ tokens instead of 15. Fixed by extracting the turn body into `_run_turn()` (returns just a stop reason) and resolving usage exactly once, in a `try/finally`, in `prompt()` — the reset is now unconditional and there's a single call site instead of seven. Added a regression test (`test_uncaught_exception_mid_turn_does_not_leak_tokens_into_next_turn`) that fails against the old code and passes now.

2. **Missing usage from background summarization.** `TokenBudgetSummarizer`'s context-compaction call goes straight through `request.client.acall()` and never emitted an `LLMResponse` event, so those real, billed tokens were invisible to `ACPEventBridge` — silently undercounting a turn's reported usage. Fixed by notifying event subscribers with `record=False`, which reaches `ACPEventBridge` without writing the fork's response into the durable transcript or LLM context (it's a background compaction call, not a conversation turn — writing it with `record=True` would have corrupted context).

Also collapsed the duplicated `LLMUsage`/`Usage` field-name lists in `event_bridge.py` into one mapping, per the review's third finding, so the two can't silently drift apart.

Full `nooa-acp` suite plus the affected summarizer tests pass (106 total, excluding two pre-existing unrelated console-script-install failures that reproduce on an unmodified checkout).

PR #371, opened 2026-09-19
fix: report summarizer state and trigger async compaction early

furgalep:
## Summary

- report the installed summarizer's real policy, token budget, configured percentage, and pending/ready state
- start automatic compaction earlier, at 75% of the usable input window by default
- make the percentage user-configurable with `threshold_fraction`
- keep compaction fully asynchronous: model generation never waits for an in-flight summary
- retain `max_tokens` as an explicit absolute-threshold override

## Root cause

`get_summarization_status()` read attributes that do not exist on `TokenBudgetSummarizer`; the values live under `summarizer.config`. It therefore reported `policy=none` and zero limits despite an installed summarizer.

Separately, automatic compaction started at 80% of usable input. On a large session, the background summary could still be running while the active prompt consumed the remaining headroom, allowing the TUI to report context above 100% of the usable window.

## Behavior

When `max_tokens` is unset, the trigger is:

```text
(model context window - reply-token reserve) * threshold_fraction
```

`threshold_fraction` defaults to `0.75`, must be strictly between `0` and `1`, and is recalculated when the model, reasoning level, or reply-token limit changes. Setting `max_tokens` pins an absolute threshold instead.

Compaction remains non-blocking. A pending summary is applied at the normal turn boundary once it is ready.

## Verification

- regression tests were observed failing before the production changes
- 100 affected summarization/runtime/coding-agent tests pass
- full nooa-cli suite: 289 passed, 2 skipped
- Ruff and `git diff --check` pass
- live GLM 5.3 E2E deliberately blocked the summary request and proved the parent returned `READY` without waiting; after release, the summary applied, input dropped from 6,663 to 2,155 tokens, and all exact facts survived the continuation


alessiodevoto (approved):
LGTM thanks!

PR #376, opened 2026-09-21
fix(tracing): journal failed LLM call inputs

sklinglernv:
## Summary

- emit a journal call record when a LiteLLM request fails
- retain the failed call's input skeleton and LLM span correlation
- keep output messages empty because a failed request has no successful assistant response
- add portable journal-file regression coverage

## Motivation

`JournalFileExporter` strips ordinary LLM message attributes from every exported
LLM span. Previously `log_failure_event()` discarded the stored input skeleton
without writing a journal call record, so failed LLM spans permanently lost their
prompt content and trace viewers could not reconstruct the diagnostic context.

## Validation

- `pytest tests/tracing/test_journal_file_exporter.py` — 5 passed
- `ruff check src/nooa/tracing/_litellm_journal.py tests/tracing/test_journal_file_exporter.py`
- `ruff format --check src/nooa/tracing/_litellm_journal.py tests/tracing/test_journal_file_exporter.py`

## Summary by CodeRabbit

- **Bug Fixes**
  - Failed LLM calls are now recorded in the journal with available session, model, timing, and span information.
  - Failed calls preserve their input details while recording empty output and token information when unavailable.
  - Missing prior call data is handled gracefully with an empty input record and a warning.


furgalep (approved):
LGTM

